Issue #3, opened 2021-10-09
use of Mapping in __init__.py incompatible with py3.10


2bndy5:
I just updated my local version of python to v3.10, rebuilt a new venv, and then tried to build some docs with this theme. The build failed saying
```
Running Sphinx v4.2.0
loading translations [en]... done

Exception occurred:
  File "C:\Users\ytreh\Documents\GitHub\venv\lib\site-packages\sphinx_immaterial\__init__.py", line 187, in <module>
    def dict_merge(*dicts: List[collections.Mapping]):
AttributeError: module 'collections' has no attribute 'Mapping'
The full traceback has been saved in C:\Users\ytreh\AppData\Local\Temp\sphinx-err-u29qe6xq.log, if you want to report the issue to the developers.
```

So, I took a look at where `Mapping` was really defined, and found it in [CPython's Lib/typing.py](https://github.com/python/cpython/blob/6846d6712a0894f8e1a91716c11dd79f42864216/Lib/typing.py#L2094)
```py
Mapping = _alias(collections.abc.Mapping, 2)
```

The code in this theme's _\_\_init\_\_.py_ is 
https://github.com/jbms/sphinx-immaterial/blob/c69c4cc9caeafe4594eb8b18c086100180b5790c/sphinx_immaterial/__init__.py#L3-L6
https://github.com/jbms/sphinx-immaterial/blob/c69c4cc9caeafe4594eb8b18c086100180b5790c/sphinx_immaterial/__init__.py#L187-L192

## Solution 
Change where `Mapping` is imported from, then remove the `collections` namespaces
```py
from typing import List, Optional, Union, Type, Dict, Sequence, Tuple, Mapping
# ...
def dict_merge(*dicts: List[Mapping]): 
     """Recursively merges the members of one or more dicts.""" 
     result = dict() 
     for d in dicts: 
         for k, v in d.items(): 
             if (isinstance(v, Mapping) and k in result 
```
I haven't gone back to an older version of python to make sure this is backward compatible, but I imagine that this repo's build CI workflow (which uses py v3.9) would error out if there's any problems.

PR #4, opened 2021-10-09
Manually glob assets during install; fix compatibilty with python v3.10

### LINKED ISSUE #1: material.css not found
Hello! It's great that you're taking this project and trying to generalize it!

I have an issue that I get this error:
GET http://localhost:63342/project_name/_build/html/_static/material.css
[HTTP/1.1 404 Not Found 16ms]

The resource from “http://localhost:63342/project_name/_build/html/_static/material.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).

Thanks :)

2bndy5:
This PR has both solution that resolves #1 and #3.

I ran the black fomatter on the setup.py, so its more PEP8 compliant. I suppose that could be undone, if I overstepped.

Another thing I changed while trying to figure things out: I switched the `subprocess.call()` calls to `subprocess.run()` calls. The [subprocess docs say](https://docs.python.org/3/library/subprocess.html#older-high-level-api),
>Prior to Python 3.5, these three functions comprised the high level API to subprocess. You can now use run() in many cases, but lots of existing code calls these functions.

Again I can undo this change as well.

Build CI has no problems with these changes and I have visually verified that the changes/solutions work as expected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ with open("requirements-dev.txt") as dev_reqs:
     REQUIREMENTS_DEV = [dev_reqs.readlines()]
 
 root_dir = os.path.dirname(os.path.abspath(__file__))
-package_root = os.path.join(root_dir, 'sphinx_immaterial')
+package_root = os.path.join(root_dir, "sphinx_immaterial")
 
 
 def _setup_temp_egg_info(cmd):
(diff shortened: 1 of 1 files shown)


jbms:
Thanks!

Issue #8, opened 2021-10-13
Contributing guidelines


2bndy5:
I'm ready to start pushing commits in prep for submitting a PR, but I'd like to verify 2 things first (so I don't have to walk some changes back).
1. Given the use of linting in JS, I'd like to add a _.pylintrc_ for linting selective python sources. Doing so would also instigate the use of pylint in this repo's _build.yml_.
2. I tend to use black to format python sources in accordance with PEP8 standards. This would be applied to everything in the _sphinx_immaterial_ folder.
   - I did find a requirement for `black==19.10b0` in the _requirements-dev.txt_
3. I'm assuming that edits to the package-lock.json & package.json should not be committed. I didn't personally make any changes, but it looks like changes were made in the build process (or setting up nodejs for this project locally).

I raised this issue because I don't see a contributing guidelines in the repo root.

Issue #6, opened 2021-10-11
Updating the docs


2bndy5:
I have been looking into what the docs need and don't need. Much of what is there seems like it was copied from the sphinx-material theme's docs.
1. I don't think using numpy as an example API is a "simple" demonstration of the theme's capabilities. Furthermore, `numpy-doc` is not PEP517 compliant (as reported by pip), which complicates building the docs on different platforms.
2. Using to a Jupyter notebook also seems to further complicate what should be a "simple" demonstration (& even simpler docs build process). It requires pandoc (a 3rd party utility not available via pypi only) to convert the notebook into a sample doc.

My concern here is that if someone looks at the sphinx-immaterial docs, then finds API docs for an entirely separate library, they might get overwhelmed/confused.

ps - I'm under the impression that completed/revised docs are 1 of the last few things preventing a release (or pre-release) to pypi. I have been using a locally copied wheel to try this theme with readthedocs.org... Found another issue with compatibility there (it isn't a deal-breaking issue).

PR #10, opened 2021-10-15
tune-up

### LINKED ISSUE #7: Better readthedocs integration
I've made some mock up edits to better integrate into readthedocs hosted docs. All details can be observed at [the issue I raised upstream](https://github.com/squidfunk/mkdocs-material/issues/3104). Unfortunately, the author has no intention to support RTD. I have 1 more use case to consider before I offer a PR here: when the user disables the nav side menu.

### LINKED ISSUE #9: mkdocs edit_url not compatible with sphinx
I've been looking for what features work as expected to determine what config stuff should go into the docs. I happened across this code
https://github.com/jbms/sphinx-immaterial/blob/6267f13b77dacf371918aa69ea1f3caf0bec5d85/src/base.html#L297-L306

but it seems to be unaltered from the original mkdocs-material theme's _src/base.html_.

### Trying to minimize changes to base.html template
Now, I would like to pass the proper `edit_url` based on config context which is updated from [`html_page_context()`](https://github.com/jbms/sphinx-immaterial/blob/d59f75c815f7018e2a651c8b0c3741db42304c9c/sphinx_immaterial/__init__.py#L200), but I think I need to make the **assumption that the document's source file is located in the docs folder**. The source file's extension (.rst or any other supported suffix) can be fetched with the template's global variable `page_source_suffix`. Maybe there's a better way to do this, but this is what I have working so far.

<details><summary>__init__.py changes</summary>

```diff
--- sphinx_immaterial/__init__.py
+++ sphinx_immaterial/__init__.py
@@ -218,235 +218,241 @@
+    repo_url = theme_options.get("repo_url", None)
     context.update(
         config={
             "theme": theme_options,
             "site_url": theme_options.get("site_url"),
             "site_name": context["docstitle"],
-            "repo_name": theme_options.get("repo_url", None),
+            "repo_url": repo_url,
             "repo_name": theme_options.get("repo_name", None),
+            "edit_url": ""
+            if repo_url is None
+            else "/".join(
+                [repo_url.rstrip("/"), os.path.split(app.srcdir)[1], pagename]
+            ),
             "extra": {
                 "version": version_config,
                 "social": theme_options.get("social"),
                 "disqus": theme_options.get("disqus"),
                 "manifest": theme_options.get("pwa_manifest"),
             },
             "plugins": theme_options.get("plugins"),
             "google_analytics": theme_options.get("google_analytics"),
         },
         base_url=base_url,
    )
```
</details>
<details><summary>template changes</summary>

```diff
--- src/base.html
+++ src/base.html
@@ -298,306 +298,306 @@
-                {% if page.edit_url %}
+                {% if config.edit_url %}
                   <a
-                    href="{{ page.edit_url }}"
+                    href="{{ config.edit_url ~ page_source_suffix }}"
                     title="{{ lang.t('edit.link.title') }}"
                     class="md-content__button md-icon"
                   >
                     {% include ".icons/material/pencil.svg" %}
                   </a>
                 {% endif %}
```
</details>

You may have noticed that RTD has its own (probably more precise) mechanism that provides links to the document's source (see pictures in #7), but I figured I'd try to tackle this for completion (& people coming from mkdocs who've only ever used gh-pages). Here's a screenshot of the above changes' result:

image

2bndy5:
The PR I've been teasing about. This resolves #7 addresses #8 closes #9 and probably solves #6 too.
Huge thanks to @jbms for handy advice (despite my pestering)!

I've been posting the solutions in the corresponding issues' threads, so I won't go over that unless needed.

### _build.yml_ workflow
The CI workflow now accounts for linting python (with `pylint` latest) and checking python code format (with `black` latest). It now also tests Sphinx's latex builder and uploads artifacts of both html & latex built docs.

### python linting
I did run `black` on all python sources in the _sphinx_immaterial_ folder, and I added a _.pylintrc_ file to tell pylint to ignore some common exceptions and all linting violations that are inherent with extending Sphinx builders.

### Theme's docs 
I urge you to visually check the docs currently hosted on [my fork's gh-pages](https://2bndy5.github.io/sphinx-immaterial/index.html#). I may have made some risque style choices, but my intention was to make these docs stick out from the other material themes' docs. The _customization.rst_ page is where I heavily focused on because it describes all the great features this theme has to offer.
   - I don't think the ["new blocks" section](https://2bndy5.github.io/sphinx-immaterial/customization.html#new-blocks) is very relevant, but I didn't really test that info for accuracy.

### Contributing guidelines
This should be improved... I just made a generic procedure for contributing (& possibly getting help). There are notes about what files to avoid changing, but I didn't specify the _package*.json_ files in that list (I defer to @jbms expertise on that topic). 

There is no stipulations about user behavior, but I think that's covered by the Github generic Code of Conduct. I hail from python's mentality in which "we're all adults here."

### gh-pages
If completed successfully the CI workflow will upload the html docs to a gh-pages branch. This is currently done on every commit, but I left a comment in the yml to limit the gh-pages updates to release events only.


.github/workflows/build.yml
```diff
@@ -44,14 +46,26 @@ jobs:
...
+    - name: Build latex docs
+      working-directory: docs
+      run: sphinx-build -b latex . _build/latex
+    - name: Upload doc builds as artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: doc-builds
+        path: docs/_build/
+    - name: upload docs to github pages
+      # if: ${{ github.event_name == 'release'}}
```
2bndy5, line 64:
```suggestion
      # if: ${{ github.event_name == 'release'}}
      if: ${{ github.event_name != 'pull_request'}}
```

.github/workflows/build.yml
```diff
@@ -44,14 +46,27 @@ jobs:
...
+      if: ${{ github.event_name != 'pull_request'}}
```
jbms, line 65: Should only upload for main branch.

Also should use explicit hash for all third-party github actions.

docs/customization.rst
```diff
@@ -1,5 +1,185 @@
 :hero: Configuration options to personalize your site.
 
+.. embedded material icons used for inline demonstration
+.. raw:: html
+
+    <style>
```
jbms, line 6: It would be good to figure out a better way to handle icons, so that we don't have to rely on raw html and hard coding the svg icons into the html.

docs/customization.rst
```diff
@@ -18,99 +198,186 @@ options are:
...
 
-The complete list of options with detailed explanations appears in
-``theme.conf``.
+Heroes
+==========
+
+To set the hero's text for an individual page, use the ``:hero:`` directive for the desired page.
+If not specified, then the page will not have a hero section.
 
 Configuration Options
```
jbms, line 210: Would be nice to create a sphinx object type for config options.  But that could be done later.

sphinx_immaterial/theme.conf
```diff
@@ -11,32 +10,42 @@ direction =
...
+# See https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#configuration
 # Set the repo url for the link to appear
 repo_url =
 # The name of the repo. If must be set if repo_url is set
 repo_name =
 # Must be one of github, gitlab or bitbucket
 repo_type = github
+# url segment that is concatenated with repo_url to point readers to the document's
+# source file. This is typically in the form of 'blob/<branch name>/<docs source folder>'.
+edit_uri = blob/main/docs
```
jbms, line 48: Weren't you going to leave this blank by default?

.github/workflows/build.yml
2bndy5: > Also should use explicit hash for all third-party github actions.

I was wondering why you do that. Using `@v3` helps the author keep patches being distributed without users needing to change their tag.

docs/customization.rst
2bndy5: I could add a extra css for the docs, but I didn't want it used for all the pages. Also, I wanted the icons to change color with the different schemes. This was all I could find that would let me do that.

It is pretty annoying that the icons aren't copied with the theme's assets. (We're talking like less than 5 MB).

sphinx_immaterial/theme.conf
2bndy5: Whoops. Testing artifact.
```suggestion
edit_uri =
```

docs/customization.rst
2bndy5: I don't fully understand. Do you mean to have a dict like bashtage did in conf.py? Personally, this implementation makes more sense to me than how sphinx-material does it.

jbms: Sphinx has the concept of an "object type" --- when you document a python function or class, or an environment variable, you create an "object" to refer to the entity that is being documented.  Sphinx defines some built-in object types but you can define your own as well.

The simplest way to define a new object type is with this:
https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_object_type

.github/workflows/build.yml
2bndy5: Kinda like this? I think the condition below (in the python-publish-package job) will only work if the branch is named master (not main).
```suggestion
      if: ${{ (github.event_name != 'pull_request') && (github.ref == 'refs/heads/main') }}
```

jbms: Yeah, should be changed to be `github.event_name == 'push' && github.ref == 'refs/heads/main'`

The condition in `python-publish-package` should also be updated.

.github/workflows/build.yml
```diff
@@ -44,14 +46,27 @@ jobs:
...
+      uses: peaceiris/actions-gh-pages@v3
```
2bndy5, line 66:
```suggestion
      uses: peaceiris/actions-gh-pages@068dc23d9710f1ba62e86896f84735d869951305 # pined to v3.8.0
```

.github/workflows/build.yml
2bndy5: I changed the condition in `python-publish-package` job and my fork _tried_ to publish to pypi (the `||` bypassed the check for a tagged commit). I have enabled actions on my fork for linting reasons.

Is it ok if we add `release` event triggers and then only publish to pypi on release events? Alternatively I think this might be solved by comparing `refs/heads/jbms/main` instead(?) or making sure that you triggered the workflow with the env var `GITHUB_ACTOR`.

2bndy5: >being pushed to the Test PyPI server --- users can install from there if they want the latest development version

Ah. That explains why you don't use `twine check`. I should've been trying this for my tests on my RTD project.

> Not sure what the tradeoffs are

I've often seen a sorely lacking release description (if any) when people manually push a tag. I often look to the release description for a list of changes, but maybe that could be supplemented with a automated change-log action (there's a few out there) - probably better addressed at a later time.

If the intention is to publish dev versions, then why limit publishing (for the test pypi server) to just the main branch? I can understand this intention for the regular pypi server.

docs/customization.rst
jbms: Sorry, my comment wasn't about `hero` in particular --- I was just suggesting that instead of documenting the supported theme options as an rst definition list, you could instead document the theme options via a new sphinx object type.

For example sphinx-rtd-theme define a `confval` object type here:

https://github.com/readthedocs/sphinx_rtd_theme/blob/c9b1bde560d8ee31400e4e4f92f2e8d7a42265ce/docs/conf.py#L90

And then uses it here to document each config option:

https://github.com/readthedocs/sphinx_rtd_theme/blob/c9b1bde560d8ee31400e4e4f92f2e8d7a42265ce/docs/configuring.rst

.github/workflows/build.yml
jbms: Branches other than main may just be work in progress, may be force pushed, etc.  Additionally, the versions published to the test pypi server will have a version number derived similarly to `git describe`, based on:
- the latest tag that contains the commit
- the number of commits since then

For a single branch without force pushes that ensures a unique, monotonically increasing version number.  But if you have multiple branches then the version numbers won't be unique.

docs/customization.rst
2bndy5: 👍🏼 you should raise this idea as an issue (if only a reminder)

.github/workflows/build.yml
2bndy5: That actually better explains the comments in 
https://github.com/jbms/sphinx-immaterial/blob/d59f75c815f7018e2a651c8b0c3741db42304c9c/setup.py#L195-L202

ok, so I'll change the condition to something like
```yml
  python-publish-package:
    # Only publish package on push to tag or default branch of jbms/sphinx-immaterial.
    if: ${{ github.event_name == 'push' && github.repository == 'jbms/sphinx-immaterial' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main') }}
```

jbms:
Not sure if you are continuously updating your gh-pages site, but I noticed that in dark mode the primary color does not have good contrast against the background color.

Also the icons in the customization page don't display for me --- they just display as white squares.

2bndy5:
I am keeping the gh-pages updated per commit. I didn't really like the accent primary color (I think its `purple` or `deep-purple`). Is that what you're referring to? Or are you talking about the header color? (which is set to `indigo`)?

I did test changing the slate scheme's hue value in an overridden css rule... We could try that as a demonstration, but it might be misleading to users about the default value (`232` I think). I really like how the mkdocs-material site allows temporarily changing the primary and accent colors dynamically, but I didn't look into that much.

2bndy5:
I'll play around with it. Yes, I'd like to avoid looking identical to the mkdocs-material site. If possible I'd like to avoid looking like sphinx-material site (though that doesn't seem difficult 😆).

I also chose green for my project's docs. So it may become the preferred color. As you pointed out: it is in the "eye of the beholder" at the end of it all.

docs/customization.rst
2bndy5: I've found this extension called sphinx-panels which adds custom roles for [inline icons](https://sphinx-panels.readthedocs.io/en/latest/#inline-icons). Unfortunately, the material design icons aren't available. However, They are moving their code base to a newer extension called [sphinx-design](https://github.com/executablebooks/sphinx-design). I've raised an issue to see if inline material icon are on the horizon.

2bndy5:
ok I abstracted the custom css into its own file and added some `-webkit-*` attributes to account for chrome (and hopefully other browsers too). Also I went with `deep-orange` in the dark scheme's primary color with `lime` as its accent color. While I was at it I changed the light scheme's accent color to `light-blue` (`teal` looked kinda sickly).

docs/customization.rst
jbms: What is the use case for referencing icons directly in the rst?  I see that in customization.rst they are used as examples for customizing this theme.  It seems like they would more be useful for customizing admonitions, or defining new directives, rather than being used directly.  Unfortunately pre-building the css makes it a bit tricky to change the icons properly, since they are inlined into the css.  On the other hand, if the css isn't pre-built then some nodejs dependencies are needed when building documentation, which would also be unfortunate.

2bndy5: Instead of just showing the names of the icons, I wanted to show the icon itself (next to the icon's name). A picture speaks a thousand words...

Yes, I would like an easier way to customize the admonitions' icon. Sphinx also has a generic admonition that can be used to add new user-defined admonitions, but for some reason this theme treats generic admonitions like a `.. note::`.

I can imagine that some users might want to put them in the section headers (hopefully resulting in a icon in the nav menu items also).

jbms: Yeah, though unfortunately libsass is deprecated --- I don't think it supports all of the sass syntax used by mkdocs-material, though I could be mistaken.  The recommended replacement is dart sass, but I don't think there are Python bindings for that.

PR #13, opened 2021-10-17
fix: Fix issue with search index that prevented incremental rebuilds

### LINKED ISSUE #12: allow consecutive builds
Due to the altered search system in this theme, any consecutive attempt to build docs results in the following error:
```
Exception occurred:
  File "/path/to/env/lib/python3.10/site-packages/sphinx/search/__init__.py", line 283, in load
    index2fn = frozen['docnames']
KeyError: 'docnames'
```
The current workaround is delete any previously built docs (specific to the project being worked on), then re-build the docs.

This affects IDE extensions designed to generate previews of documentation. I don't see a way to workaround this error in the vscode restructuredtext extension, so it would be nice to have this theme cooperate better with sphinx in this scenario.

jbms:
Fixes #12

### DIFF
--- a/sphinx_immaterial/search_adapt.py
+++ b/sphinx_immaterial/search_adapt.py
@@ -9,7 +9,8 @@
 - the unused list of filenames is removed, since it just bloated the index.
 """
 
-from typing import Dict, Tuple
+import io
+from typing import Any, Dict, IO, Tuple
 
 import sphinx.search
 import sphinx.application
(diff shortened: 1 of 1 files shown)


2bndy5:
Also, IDK if this is a high priority for you, but it seems the pypi token is invalid. I would greatly appreciate using a wheel from the test pypi server. Currently, I've been using a locally copied wheel for my RTD project, and I'd like to avoid that.

jbms:
I just pushed a PyPI package and added the token (you have to push an initial version manually before you can generate a token).

2bndy5:
https://test.pypi.org/project/sphinx-immaterial/
```
pip install -i https://test.pypi.org/simple/ sphinx-immaterial
```
there's a space in the url which prevents using it in a requirements.txt file (I have tried using single and double quotes). Without the space, pip gets a 404. Without quotes, `git clone` errors out.

PR #17, opened 2021-11-02
native dropdown and docs page about admonitions

### LINKED ISSUE #5: [question] How to change the icon of an admonition?
My intention is to have 
image
look like
image

But it looks like the built theme only allows a limited set of options. So, how do I change the icon for admonitions? Usually I'd override the CSS styling to do this, but I don't see the compiled svg I want.

### LINKED ISSUE #11: Use custom sphinx object type to document theme config options
See discussion here:

https://github.com/jbms/sphinx-immaterial/pull/10#discussion_r730147026

### LINKED ISSUE #15: CSS for details/summary html tags is tied to admonitions' CSS
### closed
image
### open
image

### code to reproduce
```py
extensions = [
    "spinx_immaterial",
    "sphinx_panels",
]
```
```rst

    .. literalinclude:: ./conf.py
```

2bndy5:
This resolves #15 and closes #5 and resolves #11 

- augmented the build CI to avoid being triggered twice on the same commit (when PR is open). It also now installs the docs' dependencies, namely sphinx-design (used for tabbed blocks but offers a lot more).
- reverted the `subprocess` usage in _setup.py_ from changes in #13 . Seems that `subprocess.run` was passing an absolute path to nodejs build task (causing it to fail).
- abstracted demonstration about admonitions from _additional_samples.rst_ as there were some easy gotchas for people migrating from mkdocs to sphinx.
- added a new directive called `material-dropdown` to make use of mkdocs-material theme's support for extra styling about undefined admonitions (not specified in rST nor Sphinx admonition directives). This also makes use of mkdocs-material theme's styling for `<details>` & `<summary>` tags.
- added a demonstration about how to customize admonitions that closely mirrors the mkdocs-material theme's docs in the same regard.
- added copied code from sphinx-rtd-theme to docs' _conf.py_. This means the customization page uses the monkeypatched `confval` directive to document theme config options
- filled out list of API icons (`OBJECT_ICON_INFO`) for secondary toc entries based on all builtin Sphinx domains

Unlike mkdocs-material theme, the admonition title cannot be disabled with an empty title string. This is because the generic admonition defined in rST specs require 1 argument (a title). It may be possible to achieve this by augmenting the material-dropdown directive (using the `open` option), but I gave up trying to make the title optional - any string (excluding directive options) immediately following the material-dropdown directive is used as the title.


2bndy5:
rendered [docs new page on my fork for preview](https://2bndy5.github.io/sphinx-immaterial/admonitions.html)

.github/workflows/build.yml
```diff
@@ -1,6 +1,8 @@
 name: Build
 
-on: [push, pull_request]
+on: [push]
```
jbms, line 3: I think we still want this to run even on pull requests.  In the special case that the pull request is coming from this same repository, then that does unfortunately result in the workflow running twice.   But I don't think that is a big deal.

sphinx_immaterial/__init__.py
```diff
@@ -269,6 +280,15 @@ def setup(app):
         "html_use_directory_uris_for_index_pages", False, rebuild="html", types=bool
     )
 
+    app.add_node(
```
jbms, line 283: Perhaps make `dropdown.py` usable as a sphinx extension on its own, by defining a `setup` function there?

2bndy5: Its a workaround. Ideally, we could use other sphinx extensions (much better designed and maintained elsewhere) if we remove the CSS rule for `details` and `summary` tags.

sphinx_immaterial/dropdown.py
```diff
@@ -0,0 +1,78 @@
+"""Taken from sphinxcontrib-details
+"""
+from docutils import nodes
```
jbms, line 3: It seems like this is effectively just a different way of styling an admonition.  I wonder if it would make sense to just augment the built-in admonition directives with an additional option to indicate if they should behave as `details` elements --- and possibly that could be upstreamed in sphinx itself.

sphinx_immaterial/dropdown.py
```diff
@@ -0,0 +1,78 @@
+"""Taken from sphinxcontrib-details
```
jbms, line 1: If this stays as is, please retain the original copyright notice of the source code from which it was derived, and modify the LICENSE file to include the relevant license.

.github/workflows/build.yml
2bndy5: sorry, I'll revert that. I was trying to figure out why the PR wasn't triggering the CI, but I think it has something to do with originating from my fork. The Ci runs (and passes) on my fork...

The double triggered workflow could be avoided by 
```suggestion
on:
  push:
  pull_request:
    types: [opened, reopened]
```
because this leaves out any push event that also triggers a PR synchronize event.

jbms:
Thanks for your work on this!

sphinx_immaterial/dropdown.py
2bndy5: I did look into that. It needs to be a `container` node to customize the HTML tag name. It only behaves like a admonition in this theme because of the CSS rule inherited from mkdocs-material.

docs/admonitions.rst
```diff
@@ -0,0 +1,230 @@
...
+
+To use the following admonitions' styles from the mkdocs-material theme, the rST syntax is
+shown inside the demonstrated admonition.
+
+.. important::
+   The ``:class:`` options below (in the rST code blocks) must use lower case letters for the
+   styling to work. Otherwise, the admonition will look like a ``note`` (as that is the
+   default fallback style).
+
+``abstract``, ``summary``, ``tldr``
```
jbms, line 48: We could define custom sphinx directives for all of these admonitions, though I'm not sure how useful that would be compared to just specifying `:class`.  Some of the names may conflict with existing sphinx directives.

2bndy5: After trying to understand the docutils module, I have to err with using the `class `option. Additionally, using the `class` option also helps mimic the mkdocs-material feature to customize a user-defined admonition (as shown in the bottom of this new docs page).

docs/customization.rst
```diff
@@ -119,174 +117,156 @@ options are:
         'globaltoc_depth': 2
     }
 
-Heroes
-==========
+.. confval:: hero
 
-To set the hero's text for an individual page, use the ``:hero:`` directive for the desired page.
-If not specified, then the page will not have a hero section.
+    To set the hero's text for an individual page, use the ``:hero:`` directive for the desired page.
```
jbms, line 122: I think this is called a document metadata field:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#file-wide-metadata

2bndy5:
```suggestion
    To set the hero's text for an individual page, use the ``:hero:`` metadata field for the desired page.
```

sphinx_immaterial/dropdown.py
2bndy5: It uses [Apache v2](https://github.com/tk0miya/sphinxcontrib-details-directive/blob/master/LICENSE). Again, this module could be removed if I can remove the mkdocs-material CSS that relies on a md ext (something we can't use anyway).

2bndy5:
BTW, I got a reply from the sphinx-design ext's author to include material icons as the extension's inline text rules. It looks like it might be an added option in the future, so we don't need to home-brew an SVG-to-CSS solution here 🤞🏼

sphinx_immaterial/dropdown.py
2bndy5: I added the complete copy of the license in dropdown.py. Additionally, I appended the copyright notice to the LICENSE file. Per the Apache v2 stipulations, I have documented (in source code comments) what was modified in dropdown.py.

sphinx_immaterial/dropdown.py
2bndy5: If we get rid of the CSS rule, the dropdown would not be styled like an admonition because there is nothing in dropdown.py that is meant to trigger the CSS admonition styles.

UPDATE: I made the dropdown title optional, but had to add the `admonition` class manually when not using a `details` tag. This could be done better if/when we exclude the inherited CSS from mkdocs-material about `details`/`summary` tags.
https://github.com/jbms/sphinx-immaterial/blob/ec14e7cda42f4bd314edfe29294ade3ab0ccc47d/src/assets/stylesheets/main/extensions/pymdownx/_details.scss#L38-L39

https://github.com/jbms/sphinx-immaterial/blob/ec14e7cda42f4bd314edfe29294ade3ab0ccc47d/src/assets/stylesheets/main/extensions/pymdownx/_details.scss#L72-L73

jbms: I'm happy to remove the include of `main/extensions/pymdownx/_details` if you think that is the best solution.

sphinx_immaterial/__init__.py
2bndy5: I have defined the setup function in dropdown.py. Now, \_\_init\_\_.py imports the setup function (under an alias) and executes it within the `__init__.setup()`

I'm not sure how much use the new directive would be as an independent extension because it largely is based on the CSS inherited from mkdocs-material.

.github/workflows/build.yml
```diff
@@ -1,6 +1,9 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  pull_request:
+    types: [opened, reopened]
```
jbms, line 5: My understanding is that `push` is only triggered when you push to the repository itself.  So if you push more commits to a pull request coming from a different repository, then the workflow won't run as you currently have it configured.

I think it is more important that we don't skip runs than that we avoid duplicate runs.

In general pull requests more commonly come from a different repository, and duplicate runs aren't that big of a deal anyway.

Maybe you could avoid duplicate runs by checking in the case of a pull_request event that it is coming from a different repository.  But then I guess it wouldn't show the results in the pull request itself?

sphinx_immaterial/dropdown.py
2bndy5: It would introduce a bit more testing, but it would also make maintenance easier as we wouldn't need our own directive.

I do wonder how we would support admonitions without a title at that point because the rst directive requires 1 arg (a title), and the arg might be limited to 1 line of input...

.github/workflows/build.yml
2bndy5: Its strange to me that my push events are only executing on my fork. Maybe because it is indeed a fork, but I use the same setup on other forks, and my push events trigger a run on upstream when a PR is open (from my fork).

There might be something up with your repo's settings in this regard.

jbms:
As a general comment, it would be nice if you can rebase your changes to one or more commits, with a parent of the current `main` branch, that correspond to logical changes.

For example:

- Fix setup.py subprocess calls
- Add to OBJECT_INFO_ICON
- Add material-dropdown extension
- Change to confval directive for documenting config options
- etc.

In some cases it isn't practical to split commits up --- e.g. if they touch nearby parts of the same file it may be annoying to split them up.  But when possible it makes the history a lot cleaner and easier to understand, and also easier to review.

sphinx_immaterial/__init__.py
jbms: My thinking was not so much that it would be used independently, but rather that it would be an optional feature.  This way it is fully self-contained.

LICENSE
```diff
@@ -50,3 +50,21 @@ LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
 ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
```
jbms, line 53: Right now you have the short form of the license in this LICENSE file, and the full text in `sphinx_immaterial/dropdown.py`.  Can you instead do the reverse: put the full text in this LICENSE file, and the short form or original copyright notice in `sphinx_immaterial/dropdown.py`?

I apologize for having so many requirements regarding the license --- but I hope to replace the copy of this theme in tensorstore with this version, and therefore must follow Google's license guidelines.

sphinx_immaterial/__init__.py
2bndy5: Hmm, that would also incur optionally `@import`ing the CSS about `details` & `summary`. That sounds nice, but I think it would more work to make it actually self-contained. 

If we remove CSS about `details`/`summary`, then the dropdown directive would be contained in a completely unrelated extension like sphinx-design. I'm all for that because we already have to rely on them for `set-tab` blocks (& hopefully inline icons in the future).

jbms: Sounds like the best option is to actually just remove the details/summary CSS from this theme then.

LICENSE
2bndy5: >I apologize for having so many requirements regarding the license

No worries at all. I admit I can be a little arrogant about these things.

Just to be clear, if we get rid of the CSS that extends makes use of the `details`/`summary` tags, then this new directive (& copyright/license issues) are obsolete. Give me the "ok", and I'll refactor the PR. It would make rebasing easier as well.

jbms: Yes, sounds good to me.

And I didn't pick up on any arrogance, so no worries about that :)

jbms:
To be clear, it is okay to open a new pull request, or separate pull requests for each change, but I wasn't specifically asking for that.  You could just force push to your same branch.

I see that this pull request is coming from your `main` branch so maybe you don't want to force push there, though.

Usually it is good to create a separate branch for each pull request (which you can delete once the pull request goes through).

2bndy5:
Its just easier for me to start a new branch, reset to upstream/main's head, and then start re-factoring. I typically don't use git CLI, and the Github desktop app (& gitkraken) have trouble differentiating between 2 branches of the same name (upstream or not). This will be mostly copy-n-paste anyway

Issue #18, opened 2021-11-04
can't render docs with built/installed theme on windows


2bndy5:
I keep running into a problem when using `npm` on windows to build the theme. It installs fine, but any docs built from the install doesn't properly link to the correct names of compiled CSS files in the installed theme's base.html template.

The generated docs try linking to files named _main.css_ and _pallete.css_. However, they should be named _main.\<sha>.min.css_ and _pallete.\<sha>.min.css_. I suspect it has something to do with the [_tools/build_](https://github.com/jbms/sphinx-immaterial/tree/main/tools/build) JS module.

Looking for JS advice again...
```
(venv) PS C:\path\to\sphinx-immaterial> npm -v
8.1.2
```

-----------
I have resorted to using WSL ubuntu bash (& related npm v8.1.2) to build a wheel and install to a Windows venv. Note that sphinx (& python in general) runs faster in Windows PS than it does in WSL terminal for some reason.

2bndy5:
I realize now (after learning more about npm) that solving this may be dependent on upstream's dev workflow. Closing this.

jbms:
Sorry, I missed this issue when you posted it previously.  There is some logic in the build scripts copied (with some modifications) from mkdocs-material that handles renaming of the files, and it should be possible to fix this.  If there is a bug in the upstream repository it could be fixed there as well.

PR #19, opened 2021-11-04
Customized admonitions

### LINKED ISSUE #5: [question] How to change the icon of an admonition?

### LINKED ISSUE #11: Use custom sphinx object type to document theme config options

### LINKED ISSUE #15: CSS for details/summary html tags is tied to admonitions' CSS

2bndy5:
This resolves #15 and closes #5 and resolves #11 

- reverted the `subprocess` usage in _setup.py_ from changes in #4 . This now allows the theme to be built on windows, but see #18
- abstracted demonstration about admonitions from _additional_samples.rst_ as there were some easy gotchas for people migrating from mkdocs to sphinx.
- added a new directive called `md-admonition` to mimic the mkdocs-material theme's ability to remove an admonition's title. This is a modified version of rST's generic `admonition` directive, thus it can be used as drop-in replacement (title can still be specified as an optional argument). I could've forced sphinx to overwrite the existing `admonition` directive, but I felt that was too intrusive. @jbms thoughts?
- added a demonstration about how to customize admonitions that closely mirrors the mkdocs-material theme's docs in the same regard.
- docs' _customization.rst_ page uses the monkey-patched `confval` directive to document theme config options
- filled out list of API icons (`OBJECT_ICON_INFO`) for secondary toc entries based on all builtin Sphinx domains.
- commented out the mkdocs-material theme's CSS styles for `details` & `summary` tags. The docs' new _admonitions.rst_ page instead refers to using sphinx-design extension (& includes a brief demonstration in a collapsible container element).


2bndy5:
I had to manually deploy docs to my gh-pages branch because the CI wasn't triggered by my main branch. The [updated preview is live](https://2bndy5.github.io/sphinx-immaterial/admonitions.html#collapsable-dropdown).

Previewing the docs would be a lot easier for PRs if you used RTD to host them. They specifically have an option to do this automatically as a CI job. However, I don't know your stance on webhooks that allow RTD to access events, namely creation/deletion of branches and tags, push, and PR events.

2bndy5:
ok I was able to get my fork hosted on RTD, but I noticed the `md-admonition` doesn't seem to render correctly when there is a title specified as an argument.

### Hosted on gh-pages:
image
### Hosted on RTD:
image

--------------
We may need to specify a required version of `docutils`

2bndy5:
fixed it. I forgot to commit 1 line in the md-admonition directive def

jbms:
Hi,

Is this about ready to go or are you still working on it?

2bndy5:
It is ready.

I added a yml for easy RTD integration (if you so choose). As far as the docs about version selector, I have to test it out again before I'm ready to commit changes to the docs. For now the #20 should serve as a reminder.

2bndy5:
> I could've forced sphinx to overwrite the existing admonition directive, but I felt that was too intrusive. @jbms thoughts?

jbms:
As far as modifying the admonition directive, I think making a separate directive makes sense for now.  But that seems like a nice candidate for upstreaming in docutils itself, since it is fully backwards compatible.

sphinx_immaterial/custom_admonition.py
```diff
@@ -0,0 +1,46 @@
...
+title optional."""
+from docutils import nodes
+from docutils.parsers.rst.roles import set_classes
+from docutils.parsers.rst.directives import admonitions
+from sphinx.application import Sphinx
+
+__version__ = "0.0.1"
+
+
+class noTitleAdmonition(admonitions.BaseAdmonition):
```
jbms, line 11: Can you rename this to `NoTitleAdmonition`?

sphinx_immaterial/custom_admonition.py
```diff
@@ -0,0 +1,46 @@
+"""This module inherits from the generic ``admonition`` directive and makes the
```
jbms, line 1: Perhaps rename this file to `md_admonition` to be consistent with the directive name?

sphinx_immaterial/__init__.py
```diff
@@ -271,6 +272,10 @@ def setup(app):
 
     app.add_builder(_get_html_builder(app.registry.builders["html"]), override=True)
     app.add_html_theme("sphinx_immaterial", os.path.abspath(os.path.dirname(__file__)))
+
+    # register our custom adminition directive
+    custom_admonition_setup(app)
```
jbms, line 277: Could just have the user add `sphinx_immaterial.md_admonition` as an additional extension in the `conf.py` if they need this.

Alternatively, if you strongly prefer to load it by default you can call: `app.setup_extension('sphinx_immaterial.md_admonition')`

sphinx_immaterial/custom_admonition.py
2bndy5: Good catch! I'm surprised the linters didn't pick up on that.

```suggestion
class NoTitleAdmonition(admonitions.BaseAdmonition):
```

sphinx_immaterial/__init__.py
2bndy5: I would prefer to have it loaded by default since it is meant to help people migrating from mkdocs-material (& it is documented as part of the theme's featured admonition directives). I can change the function call...

I should probably make a note in the docs about the ability to use it as independent extension in other themes.

2bndy5:
review changes render as expected.

2bndy5:
The build CI failed on the merge commit (building docs step). This means the docs didn't get updated and pypi didn't receive a new release.

IDK why it failed, except for reading the workflow log
```
reading sources... [ 12%] additional_samples

Exception occurred:
  File "/opt/hostedtoolcache/Python/3.9.8/x64/lib/python3.9/site-packages/sphinx_immaterial/apidoc_formatting.py", line 185, in get_signature_prefix
    parts = prefix.strip().split(" ")
AttributeError: 'list' object has no attribute 'strip'
```
The changes in this PR don't include anything about _sphinx_immaterial/apidoc_formatting.py_, so I'm not sure what caused the failure.

naor2013:
@2bndy5 Most likely because the new CI ran on Sphinx 4.3 which had breaking API changes.
Related to #21

2bndy5:
> get_signature_prefix now returns a list of nodes rather than strings.

Yep that's it. I'll have to do some testing, but my first instinct is to use `isinstance(prefix, str)`. However, it should probably involve a temporary check for sphinx v4.3 until that issue is resolved.

In the meantime we should probably adjust the docs/requirements.txt file to avoid sphinx v4.3

naor2013:
@2bndy5 seems like a permanent solution like you suggested (isinstance) shouldn't take too much time, so I don't think there is a need for a temporary solution, at least it seems that way.
I can try to do a PR if you'd like, although it might take me longer because I don't know the code too well.

2bndy5:
My `isinstance()` idea addresses half the problem. The other half would be identifying the version of sphinx being used.

see also https://github.com/jbms/sphinx-immaterial/issues/21#issuecomment-967066460

This discussion should really committee there. I didn't know that issue was raised when I posted my comment here.

Issue #14, opened 2021-10-18
multilined cpp function signatures split between param name and datatype


2bndy5:
I'm getting ready to start exporting docygen XML to sphinx-immaterial (via `breathe` ext), and I noticed that function signatures that are wrapped to multiple lines are confusingly split between the parameter's datatype and name. Observe a test I made without doxygen/breathe (using standard sphinx cpp domain directives):
![image](https://user-images.githubusercontent.com/14963867/137712547-94570da5-1180-4cb3-ab09-8e814545a806.png)
```rst
.. cpp:function:: const MyType Foo(const MyType bar, uint8_t baz, bool flag, uint16_t foobar, int32_t foobaz, unsigned long barbaz)

   Some function description.
```

I'm suspecting the fix should be placed in _apidoc_formatting.py_ of this theme, but I still need to research more about how it detects the directive's domain.

2bndy5:
I was able to solve this in CSS, but it is ugly as hell. I'll keep learning docutils (whose docs are ironically terrible), but for now I think I'll just add custom CSS to my projects until a more robust solution can be realized.

```css
.md-typeset dl.api-field>dt.sig-wrap .sig-param + .sig-param::before,
.md-typeset dl.objdesc>dt.sig-wrap .sig-param + .sig-param::before,
.md-typeset dl.api-field>dt.sig-wrap .n.sig-param + .kt::before,
.md-typeset dl.objdesc>dt.sig-wrap .n.sig-param + .kt::before,
.md-typeset dl.api-field>dt.sig-wrap .n.sig-param + .n:not(.sig-param)::before,
.md-typeset dl.objdesc>dt.sig-wrap .n.sig-param + .n:not(.sig-param)::before {
    content: "\a     ";
    white-space: break-spaces;
}
```
This still doesn't handle template parameters and the first arg to long function signatures... The later wouldn't be hard, but template descriptions is bit more complex since the entire signature is parsed from 1 long line - implying that we'd have to parse more than 1 set of brackets and determine if template/function/both parameters needs multiple lines.

2bndy5:
I've started expanding on the CSS solution currently implemented. I also went through and made sure that syntax highlighting is triggered properly (while still maintaining the `a.reference` override within signatures).

python 
![image](https://user-images.githubusercontent.com/14963867/161869488-f0953284-5687-4ac2-991d-a8a5a2effbe1.png)

c++
[...]

PR #22, opened 2021-11-12
Fix #21

### LINKED ISSUE #21: Sphinx 4.3 Compatability
Firstly, thanks for making this port available - it works and looks amazing!

It seems the newest version of sphinx (4.3) has a few API changes that leads to some breaks in this package. Namely:

* `sphinx.search.IndexBuilder.get_objects()` now returns a list of tuples rather than a dict, where the old dict key has now been included as the last item in the tuple.
* `get_signature_prefix` now returns a list of nodes rather than strings.

I'd be happy to contribute a PR with possible fixes if you'd like.

SimonBoothroyd:
This PR aims to fix #21 in a backwards compatible way.

It 

* updates the index builder to handle either a dict or list of children
* updates the api doc signature patch to handle a list of nodes rather than a list of strings.

@jbms I wasn't sure why the api doc patch is removing properties from the parts list, and this line doesn't seem to be hit when building the example docs in the repo. Any ideas if we need it? 

For now I just return the output of the original function and that seems to yield docs that look identical.

### DIFF
--- a/sphinx_immaterial/apidoc_formatting.py
+++ b/sphinx_immaterial/apidoc_formatting.py
@@ -182,7 +182,7 @@ def _monkey_patch_python_get_signature_prefix(
 
     def get_signature_prefix(self, sig: str) -> str:
         prefix = orig_get_signature_prefix(self, sig)
-        if sphinx.version_info[0] >= 4 and sphinx.version_info[1] >= 3:
+        if sphinx.version_info >= (4, 3):
             return prefix
         parts = prefix.strip().split(" ")
         if "property" in parts:
--- a/sphinx_immaterial/apidoc_formatting.py
+++ b/sphinx_immaterial/apidoc_formatting.py
@@ -182,6 +182,8 @@ def _monkey_patch_python_get_signature_prefix(
 
     def get_signature_prefix(self, sig: str) -> str:
         prefix = orig_get_signature_prefix(self, sig)
+        if sphinx.version_info[0] >= 4 and sphinx.version_info[1] >= 3:
+            return prefix
         parts = prefix.strip().split(" ")
         if "property" in parts:
             parts.remove("property")
--- a/sphinx_immaterial/search_adapt.py
+++ b/sphinx_immaterial/search_adapt.py
@@ -10,7 +10,7 @@
 """
 
 import io
-from typing import Any, Dict, IO, Tuple
+from typing import Any, Dict, IO, List, Tuple
 
 import sphinx.search
 import sphinx.application
@@ -19,15 +19,22 @@ import sphinx.application
 class IndexBuilder(sphinx.search.IndexBuilder):
     def get_objects(
         self, fn2index: Dict[str, int]
-    ) -> Dict[str, Dict[str, Tuple[int, int, int, str, str]]]:
+    ) -> Dict[str, List[Tuple[int, int, int, str, str]]]:
         rv = super().get_objects(fn2index)
         onames = self._objnames
-        for prefix, children in rv.items():
+        for prefix in rv:
             if prefix:
                 name_prefix = prefix + "."
             else:
                 name_prefix = ""
-            for name, (docindex, typeindex, prio, shortanchor) in children.items():
+            if sphinx.version_info[0] >= 4 and sphinx.version_info[1] >= 3:
+                # From sphinx 4.3 onwards the children dict is now a list
+                children = rv[prefix]
+            else:
+                children = [(*values, name) for name, values in rv[prefix].items()]
+            for i, (docindex, typeindex, prio, shortanchor, name) in enumerate(
+                children
+            ):
                 objtype_entry = onames[typeindex]
                 domain_name = objtype_entry[0]
                 domain = self.env.domains[domain_name]
@@ -39,7 +46,23 @@ class IndexBuilder(sphinx.search.IndexBuilder):
                     synopsis = get_object_synopsis(objtype, full_name)
                     if synopsis:
                         synopsis = synopsis.strip()
-                children[name] = (docindex, typeindex, prio, shortanchor, synopsis)
+                if sphinx.version_info[0] >= 4 and sphinx.version_info[1] >= 3:
+                    rv[prefix][i] = (
+                        docindex,
+                        typeindex,
+                        prio,
+                        shortanchor,
+                        synopsis,
+                        name,
+                    )
+                else:
+                    rv[prefix][name] = (
+                        docindex,
+                        typeindex,
+                        prio,
+                        shortanchor,
+                        synopsis,
+                    )
         return rv
 
     def freeze(self):
--- a/sphinx_immaterial/search_adapt.py
+++ b/sphinx_immaterial/search_adapt.py
@@ -10,7 +10,7 @@
 """
 
 import io
-from typing import Any, Dict, IO, List, Tuple
+from typing import Any, Dict, IO, List, Tuple, Union
 
 import sphinx.search
 import sphinx.application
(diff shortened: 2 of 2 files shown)


2bndy5:
> I wasn't sure why the api doc patch is removing properties from the parts list, and this line doesn't seem to be hit when building the example docs in the repo. Any ideas if we need it?

I'm guessing this was a style choice on behalf of @jbms. The line is likely not hit in the docs build because there isn't any instance attributes (only class attributes) in demo python API which was copied from sphinx-rtd-theme docs.


I added an instance attribute to the doc's demo python API to trigger this line.
```py
class Foo:
    # ...
    baz = 2
    """Docstring for class attribute Foo.baz."""

    # not currently present in demo API
    @property
    def is_foobar(self) -> bool:
        """Docstring for instance attribute."""
        return True
```
The results are:

using sphinx v4.2
image
image

using sphinx v4.3
image
image


I'd vote to not remove "properties" from the signature (in sphinx v4.2) because there is a very significant distinction between class vs instance attributes.

SimonBoothroyd:
Ah that makes sense!

> I'd vote to not remove "properties" from the signature (in sphinx v4.2) because there is a very significant distinction between decorated vs un-decorated instance attributes.

+1

sphinx_immaterial/search_adapt.py
```diff
@@ -19,15 +19,22 @@
...
         rv = super().get_objects(fn2index)
         onames = self._objnames
-        for prefix, children in rv.items():
+        for prefix in rv:
             if prefix:
                 name_prefix = prefix + "."
             else:
                 name_prefix = ""
-            for name, (docindex, typeindex, prio, shortanchor) in children.items():
+            if sphinx.version_info[0] >= 4 and sphinx.version_info[1] >= 3:
```
jbms, line 30: Change this to `sphinx.version_info >= (4, 3)` as otherwise the condition will be false for e.g. version 5.3.

sphinx_immaterial/search_adapt.py
```diff
@@ -39,7 +46,23 @@ def get_objects(
                     synopsis = get_object_synopsis(objtype, full_name)
                     if synopsis:
                         synopsis = synopsis.strip()
-                children[name] = (docindex, typeindex, prio, shortanchor, synopsis)
+                if sphinx.version_info[0] >= 4 and sphinx.version_info[1] >= 3:
```
jbms, line 49: Change condition same as above.

sphinx_immaterial/search_adapt.py
```diff
@@ -19,15 +19,22 @@
 class IndexBuilder(sphinx.search.IndexBuilder):
     def get_objects(
         self, fn2index: Dict[str, int]
-    ) -> Dict[str, Dict[str, Tuple[int, int, int, str, str]]]:
```
jbms, line 22: I guess this should be a Union return type, ideally with a comment about the version.

sphinx_immaterial/apidoc_formatting.py
```diff
@@ -182,6 +182,8 @@ def _monkey_patch_python_get_signature_prefix(
 
     def get_signature_prefix(self, sig: str) -> str:
         prefix = orig_get_signature_prefix(self, sig)
+        if sphinx.version_info[0] >= 4 and sphinx.version_info[1] >= 3:
```
jbms, line 185: Change version condition logic as described in other comment.

jbms:
Yes, the reason for removing "property" from the signature line was a stylistic choice.

It wasn't so much to improve the regular full object description, but rather to improve the custom autosummary used in tensorstore:
https://google.github.io/tensorstore/python/api/tensorstore.IndexTransform.html#accessors

I thought it was nicer stylistically, and also easier to read, when all of the names (be they methods or properties) lined up, similar to doxygen class listings.

Additionally, I would agree that the distinction between `static` and instance properties/attributes is important (so static is still retained), but the distinction between a property and regular data attributes did not seem important, as that is really just an implementation detail.

I suppose one problem is that if you have something like:

```python
class Foo:
    x = 10

assert Foo.x == 10
foo = Foo()
assert foo.x == 10
foo.x = 100
assert foo.x == 100
assert Foo.x == 10
```

Then you can access `x` both as a "static" property of `Foo` but it can also serve as a default value for an `x` data attribute of `Foo` instances.

jbms:
In fact currently `static` is retained which interferes with the horizontal alignment of members in the autosummary:

https://google.github.io/tensorstore/python/api/tensorstore.Promise.html

I think it might be better to indicate static methods (and other things like that) on a separate line rather than as a prefix, using the more Python-like syntax

```python
@staticmethod
foo(a: T, b: U) -> V
```

The @staticmethod part would be shown in a different color than the name.

I suppose another option would be to modify the prefixes only for the autosummary, not for full object descriptions, though I think there is an advantage to keeping the styling uniform as well.

2bndy5:
This is what the [python docs say about the builtin `property()`](https://docs.python.org/3/library/functions.html#property)

>Then you can access `x` both as a "static" property of `Foo` but it can also serve as a default value for an `x` data attribute of `Foo` instances.

Ultimately, it is up to the doc's author to provide insight on the behavior of an object's member. Conventionally speaking, read-only properties should indicate (via docstr) that they are read-only. Otherwise, the user could run into a `NotImplementedError`
```py
>>> obj = Foo()
>>> obj.is_foobar
True
>>> obj.is_foobar = False
NotImplementedError: is_foobar does not have a setter
```

The [python docs do steer against using class attributes](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables), but like you said, it is an implementation detail...

I do like the idea of putting the signature prefix on its own line. This could be something to add to the resolution of #14 (for all language syntax).

sphinx_immaterial/search_adapt.py
SimonBoothroyd: Thanks that's much cleaner as well - changed in fac934c

sphinx_immaterial/search_adapt.py
SimonBoothroyd: Added in fac934c

jbms:
Thanks.  I have merged this for now to fix the documentation build, although I am still inclined to remove the `property` prefix since, as mentioned, it seems to me that it is just an implementation detail not really relevant to users of the API.

Strangely it appears that github actions are still not being run for pull requests.

2bndy5:
> it seems to me that it is just an implementation detail not really relevant to users of the API

If its a style choice, then the prefix removal should be abstracted to your customized implementation of autodoc extension. To remove it goes against the norm of documenting python APIs...

I already offered my opinion about how it _is_ relevant to API users.  But, just for clarity, the "property" prefix indicates that is ins't just another variable you can set/get like you would any other python variable. Rather, it should indicate that the attribute's implementation has implications that should considered before using it like any other python variable.

2bndy5:
pylint complains:
> sphinx_immaterial/search_adapt.py:32:8: C0206: Consider iterating with .items() (consider-using-dict-items)

---

>Strangely it appears that github actions are still not being run for pull requests.

I still think there's something up with this repo's settings. Also, note that approval to run workflows for first time contributors is not the same as approval of changes in the PR.

jbms:
I fixed the remaining issues, including the lint error, in #24.

Per your request to retain the `property` prefix, I will see about adding a config option to control that rather than doing it unconditionally.

I intend to add the custom autosummary and autodoc changes from tensorstore, and also the json domain, into this theme as well (as optional features).

jbms:
As far as github actions I'm still not sure what the problem is.  For this pull request I don't think there even was a prompt to approve running actions.

Issue #28, opened 2021-11-29
`font` instead of `fonts` in docs


Vegebutcher:
Found another bug in the theme docs: in the _[New blocks](https://jbms.github.io/sphinx-immaterial/customization.html#new-blocks)_ section there's a description of a block responsible for substituting the default font. But it should read `fonts` instead of `font` – I realised that only after reading the code of the default `layout.html` file here in the repo:

<img width="706" alt="font instead of fonts" src="https://user-images.githubusercontent.com/907743/143959075-56289194-5757-4498-a8f6-f9d48472f4bd.png">

When you use the following `font` block in the /_template/layout.html file that overrides defaults, nothing happens (the text font is still Roboto):

```jinja
{# Import the theme's layout. #}
{% extends '!layout.html' %}

{% block font %}
        <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
        <link
          rel="stylesheet"
          href="https://fonts.googleapis.com/css?family=Lato:300,400,400i,700%7CRoboto+Mono&display=fallback"
        />
        <style>
          :root {
            --md-text-font-family: "Lato";
            --md-code-font-family: "Roboto Mono";
          }
        </style>
{% endblock %}
```

But this code works when you use `{% block fonts %}`.

jbms:
You can actually change the fonts without overriding templates at all, just using:

```python
"font": {"text": "Roboto", "code": "Roboto Mono"},
```

in your `html_theme_options`.

We should adapt the documentation based on this:
https://squidfunk.github.io/mkdocs-material/setup/changing-the-fonts/

2bndy5:
Also I'm pretty sure `fonticon` won't work here.

Vegebutcher:
The trick with `font` in `html_theme_options` indeed works, thanks!

Didn't check `fonticon` at all, so can't tell.

2bndy5:
I don't see anything in the templates that makes use of a `fonticon` block. I think this is a relic from sphinx-material theme's docs.

On the other hand, `footerrel` seems to have been inherited from the mkdocs-material upstream. See footer.html template

Issue #25, opened 2021-11-15
No article TOC/index scrolling for Python elements


Vegebutcher:
In my testing environment the Sphinx-Immaterial version of docs for our [Python API](https://galaxy-integrations-python-api.readthedocs.io/en/latest/galaxy.api.html#module-galaxy.api.plugin) has an issue with the right-side TOC panel: it doesn't scroll and is locked at approx. 1/3rd of the list of available Python methods. As I can see from the template docs in the [Customization section](https://jbms.github.io/sphinx-immaterial/customization.html), this index scrolls smoothly, when its elements are derived from headings.
<img width="1440" alt="Screenshot" src="https://user-images.githubusercontent.com/907743/141759884-de39cfd9-eb7b-41bd-b339-b7e7fe9674ad.png">

2bndy5:
What browser are you testing with?

I noticed that Firefox doesn't scroll certain menus when the mouse first hovers over the menu; I have to wiggle the mouse over the desired menu for it scroll. This experience isn't specific to sphinx-immaterial theme.

The customization page of the docs uses only 2 or 3 explicit headings. Everything else is declared using a custom-defined `confval` directive. So, maybe that page isn't the best to compare to. Try the doc's demo API page for comparison.

Vegebutcher:
Both Chrome and Firefox don't allow scrolling in this pane. Scrolling on your demo API page works, but the index consists of both headings (_C++ API_, _JavaScript API_ etc.) and Python elements. I wonder if it still worked, when no headings were there, only Python elements.

2bndy5:
It would be easier to help you troubleshoot if you shared a copy of your docs. This can be done either via RTD (as a separate branch build) or via uploading a zip of your built docs to a comment here.

2bndy5:
BTW, it looks like your API doc uses a mix of headings and autodoc directives, unless you've modified it for the sphinx-immaterial theme. So, I don't think the problem is specific to python API entries.

Vegebutcher:
The original docs are [hosted on RTD](https://galaxy-integrations-python-api.readthedocs.io/en/latest/galaxy.api.html#module-galaxy.api.plugin) and I did the immaterial build locally, so here it is zipped: [build.zip](https://github.com/jbms/sphinx-immaterial/files/7539778/build.zip)

2bndy5:
[...]

PR #29, opened 2021-12-03
More revisions

### LINKED ISSUE #20: version selector working?
I recently tried the version selector as the docs for this theme instruct, but it didn't work; the JSON file option nor the _conf.py_ dict option made any difference or triggered any unusual build-time output. I know @jbms noted some changes to the version selector dopdown in his initial PR to the sphinx-material theme's repo
> I changed the versions.json format to match the "mike" format supported by mkdocs-material.

So, I tried a json in the form that the mike plugin expects: still no joy.

At this point I think its just broken from a change in the HTML template's context variables. However, I'm actually having trouble finding the section of the template that integrates the version selector dropdown. I see the word "version" mentioned in the bottom of _base.html_ about translations, but I have a feeling the mkdocs-material theme supports the mike plugin via added JS (which is where my expertise stops).

On a related note, I think I found the Sphinx alternative to the mkdocs' mike plugin: [sphinxcontrib-versioning](https://github.com/sphinx-contrib/sphinxcontrib-versioning). It seems that this extension is the key ingredient to versioned docs on RTD as most of the sphinx themes' support it via [the extension's html template API](https://sphinxcontrib-versioning.readthedocs.io/en/latest/context.html).

--------------
My initial intention was to try and have a doxygen rendered version and a sphinx/breathe rendered version of a project's docs I'm working on.

### LINKED ISSUE #23: "language" parameter missing in docs, build fails when there's none in conf.py
Hi,
Recently I stumbled upon your project while looking for a Sphinx theme that would make our Python API docs (hosted on RTD) similarly looking to the general docs built in Material for MkDocs. A bit disappointed with the sphinx-material theme, I decided to test yours. Unfortunately, my first build attempts were unsuccessful:

`writing output... [ 20%] galaxy.api
Theme error:
An error happened in rendering the page galaxy.api.
Reason: TemplateNotFound('partials/languages/None.html')`

I figured out it sth related to the language of the project, but in your docs there's nothing about the need to set the language in the `conf.py` file, otherwise the builder will fail (I guess it takes the `None` value and looks for the respective `None.html` file, which simply doesn't exist). In your `conf.py` file I found `language = "en"`, put it into my `conf.py` and it worked like a charm.

You should either update the docs with some info regarding this parameter or implement an exception, which handles unspecified language by defaulting to EN, for instance.

Apart from that, it's a fantastic work! I'm impressed how smoothly you integrated the great Material for MkDocs theme with Sphinx.

2bndy5:
- resolves #20 by revising the "Version Dropdown" docs
   - docs also now demonstrate the version dropdown selector by offering "ReadTheDocs" version and "Github Pages" version.
- fixes #23 by making English the default language if `language` option was not specified in _conf.py_
- implements suggestions from #25 
   - revises docs about `palette` option
   - uses `favicon_url` and `logo_url` introduced in sphinx 4.0 (& recommended as `favicon`/`logo` are now deprecated)
   - removes some confusing config options like
      - `globaltoc_includehidden` (internally defaults to `True`)
      - `globaltoc_depth` (internally set to `-1` by default)
- addresses #28 and removes doc artifacts from sphinx-material theme, namely `font` & `font_icon`
   - added `font` option to `html_theme_options` docs. This includes an example snippet & a link to [valid font faces](https://fonts.google.com)
- tweaked some other things in the docs' customization page while I was proofreading and adding example snippets.


docs/conf.py
```diff
@@ -126,26 +124,36 @@
             },
         },
     ],
-}
+    "version_dropdown": True,
+    "version_info": [
+        {
+            "version": "https://sphinx-immaterial.rtfd.io",
+            "title": "ReadTheDocs",
+            "aliases": ["RTD"]
```
jbms, line 132: Note: `aliases` are for specifying alternative directory names to match against the URL in order to figure out what the current version is.  They aren't for displaying to the user.

For example, if you have:

```
1.0/
2.0/
3b1/
```

and your webserver additionally redirects `stable/` -> `2.0/` and `latest/` to `3b1/`  then you should make `stable` and alias of `2.0` and `latest` an alias of `3b1`.

I'm guessing in this case you don't intend for `RTD` to be the actual path.

docs/customization.rst
```diff
@@ -174,101 +197,257 @@ Configuration Options
...
 
     .. confval:: google_analytics_account
 
         Set to enable google analytics.
 
-    .. confval:: globaltoc_depth
+    .. confval:: globaltoc_collapse
 
-        The maximum depth of the global TOC; set it to -1 to allow unlimited depth.
+        If true (the default value), TOC entries that are not ancestors of the current page are collapsed.
```
jbms, line 317: Consider adding a warning that setting this to `false` may lead to large generated page sizes since the entire table of contents tree will be duplicated on every page.

docs/customization.rst
```diff
@@ -174,101 +197,257 @@ Configuration Options
...
+
+Supported Approaches
+********************
+
+There are two approaches:
+
+1. The version information is stored in a :confval:`version_info` `list` in the conf.py file.
+   
+   .. note::
+       Notice this is an ordered list. Meaning, aproach 1 will take precedence over approach 2.
```
jbms, line 380: Should be "approach"

docs/customization.rst
```diff
@@ -174,101 +197,257 @@ Configuration Options
...
+           /1.0/index.html
+           /2.0/index.html
+
+
+Version Information Structure
+*****************************
+
+Both approaches use a data structure similar to what is used by the
+`mkdocs mike plugin <https://github.com/jimporter/mike>`_. Contrary to what the mike plugin's
+README says, the ``alias`` field is not optional, but it can be set to an empty list if not using
```
jbms, line 412: Should be `aliases`, also we should perhaps document the meaning of `aliases`.

jbms:
Thanks!

docs/conf.py
2bndy5: I want really sure what aliases we're for (until now).

docs/customization.rst
2bndy5: Agreed. The mike plugin's README doesn't seem to explain the alias in detail (other than exemplifying it's usage). I'm guessing you know how to use this better after modifying the JS for this theme.

docs/conf.py
2bndy5: I used this info in hopes of resolving the comment below about explaining the use of the `aliases` field.

docs/customization.rst
```diff
@@ -175,100 +198,287 @@ Configuration Options
...
+aliases. Other required fields include ``version`` and ``title``.
+
+- The ``version`` field can be set to a relative/absolute path or a URL.
+- The ``title`` field is a string used to describe the version in the selector's dropdown menu.
+- The ``aliases`` field is meant for giving a specific version a surname like "latest" or "stable".
+  This way the alias specified will redirect to the corresponding ``version``'s path. 
+
+  .. note::
+      Aliases are not specified via the dropdown menu. Aliases are used for command line arguments
+      to the mike plugin's CLI.
```
2bndy5, line 426: Should I re-word this to mention how using "URL/stable" will redirect to "URL/2.0"?

Do I even need to mention the mike plugin's CLI at all?

jbms: I don't think we should mention the mike plugin cli since I don't think it applies to sphinx, we are just using the same json format.

2bndy5: I took this note out, re-worded the initial brief explanation of `aliases`, and added a note after the example snippet from these docs that use external URLs

jbms:
Thanks, this looks good to me / ready to merge.

2bndy5:
I should probably mention that I make it a point to not merge my own PRs (unless I'm the only maintainer). If you think it's ready to merge, you can take co-authorship credit by squash-n-merging it.

Issue #31, opened 2022-02-15
`"navigation.tabs"` links to page's first section


mhostetter:
I'm in the process of converting my library over to this theme. Thanks for all the hard work creating this fork!!

I noticed when using navigation tabs that after clicking on the tab, the page `/the-page.html` is loaded with `#first-section` appended. So the URL is `/the-page.html#first-section` and not `/the-page.html`.

This can be observed on @2bndy5's website https://circuitpython-nrf24l01.readthedocs.io/en/latest/index.html.

![image](https://user-images.githubusercontent.com/12112573/154067998-cdb5b360-d566-45c9-b9a2-b140d230a1de.png)

When clicking "Network API Reference", the URL `/topology.html#network-levels` is loaded and not `/topology.html`.

![image](https://user-images.githubusercontent.com/12112573/154068121-fefaa736-d095-48eb-95ba-415b4cc2e9ca.png)

Desired output:

![image](https://user-images.githubusercontent.com/12112573/154068440-00063c51-6444-4858-a7d5-3717841c648e.png)

PS: I noticed that when clicking on a navigation tab while currently on that tab (but halfway down the page), it loads the page correctly without the `#first-section` appended.

2bndy5:
> I noticed when using navigation tabs that after clicking on the tab, the page /the-page.html is loaded with #first-section appended. So the URL is /the-page.html#first-section and not /the-page.html.

I also noticed that a while back, but I didn't think anything of it. Apparently, your desired behavior is consistent with mkdocs-material theme.

>PS: I noticed that when clicking on a navigation tab while currently on that tab (but halfway down the page), it loads the page correctly without the #first-section appended.

You shouldn't be able to see the tabs (let alone click on them) if you've scrolled down at all. I can't reproduce this bit.

mhostetter:
> You shouldn't be able to see the tabs (let alone click on them) if you've scrolled down at all. I can't reproduce this bit.

You're right. I had to scroll back to the top and then click the tab. However, when doing so once already on the tab, it didn't link to the `#first-section` like it does when clicking a tab from a different page entirely.

mhostetter:
This may help in the debugging...

When clicking the "Troubleshooting Info" tab on https://circuitpython-nrf24l01.readthedocs.io/en/latest/index.html, it links to "Attribute dependency" which is a 3rd-level heading. It is also the last heading before normal paragraph text is inserted. Maybe that is related to which section is linked?

[...]

PR #32, opened 2022-02-23
Prevent navigation tabs linking to subsection

mhostetter:
I know nearly no HTML/CSS, so forgive my n00bness.

After the discussed in #31, I tested changes locally with my project. I found the "Main navigation item with nested items" `if` statement doesn't need to run.

https://github.com/jbms/sphinx-immaterial/blob/1597b3e8c81ee2834a152ba77ba833d667f32e8a/src/partials/tabs-item.html#L31-L48

Only the `else` statement needs to run.

https://github.com/jbms/sphinx-immaterial/blob/1597b3e8c81ee2834a152ba77ba833d667f32e8a/src/partials/tabs-item.html#L49-L56

Instead of deleting the `if` statement, I just set it to `false` to minimize merge conflicts with upstream.

### DIFF
--- a/src/partials/tabs-item.html
+++ b/src/partials/tabs-item.html
@@ -29,7 +29,7 @@
 {% endif %}
 
 <!-- Main navigation item with nested items -->
-{% if nav_item.children %}
+{% if false %}
   {% set atitle = atitle | d(nav_item.title) %}
   {% set nav_item = nav_item.children | first %}
 


jbms:
Thanks.  I think this simple fix is indeed the right solution.  I had forgotten that nav_adapt.py already does the necessary propagation of the url, so that toc entries without a corresponding page are already assigned the url of their first descendant, so the recursion in tabs-item.html is unnecessary.

2bndy5:
@mhostetter well done. BTW you just changed some jinja syntax.

I opened a new issue #33 as a reminder about the needed addition to the docs.

2bndy5 (approved):
I just tested this using the CI artifacts (built my CirPy_nRF24 docs), and I can confirm that this solves #31

PR #40, opened 2022-03-02
support `hide` metadata and update docs

### LINKED ISSUE #37: [Reminder] about navigation.indexes not needed
This seems worth mentioning in the docs' listed features.

_Originally posted by @2bndy5 in https://github.com/jbms/sphinx-immaterial/pull/34#r814462188_

### LINKED ISSUE #39: Hide toc or navigation seems not work
When I try to hide table of contents sidebars or navigation, it did not work. I add the following line in markdown files

```
---
hide:
  - navigation
  - toc
---
```

Is this the right method to hide table of contents sidebars or navigation? Thanks.

2bndy5:
This resolves #39 and closes #37 

I updated the docs to guide users on how to hide a certain page's ToC and navigation menu. While I was updating customization.rst, I also added a hint about sphinx auto-implementing `navigation.indexes` feature from mkdocs-material.

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -130,9 +130,8 @@ Each metadata is evaluated as a ``:key: value`` pair.
 
 .. confval:: hide-navigation
 
-    If specified, hides the global navigation sidebar shown on the left side (or right side in
-    ``rtl`` languages `direction`) of the page. By default, the navigation menu is shown if the
-    browser viewport is sufficiently wide.
+    If specified, hides the global navigation sidebar shown on the left side of the page.
+    By default, the navigation menu is shown if the browser viewport is sufficiently wide.
 
     .. code-block:: rst
         :caption: Hide the navigation menu like so:
@@ -141,10 +140,10 @@ Each metadata is evaluated as a ``:key: value`` pair.
 
 .. confval:: hide-toc
 
-    If specified, hides the local table of contents shown on the right side (or left side in
-    ``rtl`` languages `direction`) of the page. By default the local table of contents is shown if
-    the page contains sub-sections and the browser viewport is sufficiently wide. If the
-    ``toc.integrate`` `feature <features>` is enabled, then this option has no effect.
+    If specified, hides the local table of contents shown on the right side of the page.
+    By default the local table of contents is shown if the page contains sub-sections and the
+    browser viewport is sufficiently wide. If the ``toc.integrate`` `feature <features>` is
+    enabled, then this option has no effect.
     
     .. code-block:: rst
         :caption: Hide the Table of Contents like so:
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -224,7 +224,7 @@ Configuration Options
 
         .. hint::
             Sphinx automatically implements the
-            `navigation.index <https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#section-index-pages>`_
+            `navigation.indexes <https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#section-index-pages>`_
             feature.
 
     .. confval:: palette
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -104,11 +104,16 @@
 Customization
 =============
 
-This theme is customized via fields in the `html_theme_options` `dict` in *conf.py*.
-Unlike, newer versions of mkdocs-material theme, this theme also supports the use of a textual
-"hero" section.
+Metadata for a single page
+==========================
+
+Each page can support a set of page-specific options. These are configured using metadata roles.
+Each metadata is evaluated as a ``:key: value`` pair.
 
 .. confval:: hero
+
+    Unlike, newer versions of mkdocs-material theme, this theme also supports the use of a textual
+    "hero" section.
 
     To set the hero's text for an individual page, use the ``:hero:`` metadata field for the desired page.
     If not specified, then the page will not have a hero section.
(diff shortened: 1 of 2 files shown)


docs/customization.rst
```diff
@@ -118,6 +118,24 @@ Unlike, newer versions of mkdocs-material theme, this theme also supports the us
 
         :hero: Configuration options to personalize your site.
 
+.. confval:: hide
```
jbms, line 121: Technically this isn't an `html_theme_options` dict (same applies to `hero`) so it might be better to have a separate directive and separate section of the documentation for sphinx metadata fields.

docs/customization.rst
```diff
@@ -118,6 +118,24 @@ Unlike, newer versions of mkdocs-material theme, this theme also supports the us
...
+        :hide: navigation
+
+    .. note::
+
+        Instead of using ``:hide: toc``, this theme uses the ``:tocdepth:`` metadata to hide the
+        page's Table of Contents.
+
+        .. code-block:: rst
+
+            :tocdepth: 0
```
jbms, line 137: I also noticed that tocdepth is not actually respected by nav_adapt.py, other than handling the special case of `tocdepth == 0`.

Possibly we could support `:hide: toc` as well for consistency with mkdocs and since it is clearer than having to specify `tocdepth`.

jbms: I guess one challenge with `:hide:` as a field element is that it seems that sphinx creates a dictionary --- is there a way to support multiple `:hide:` metadata values?

2bndy5: >is there a way to support multiple :hide: metadata values?

This is why I chose to keep the implemented tocdepth metadata. I suppose we could do
```py
hide_list = meta.get("hide").split(",")
if "navigation" in hide_list:
    # ...
if "toc" in hide_list:
    # ...
```

docs/customization.rst
2bndy5: Yeah that does seem confusing. And it doesn't seem explicitly clear that the metadata only applies to a single page.

docs/customization.rst
jbms: Actually I think we should go with something like `:hide-navigation:` rather than `:hide: navigation` --- that way we are more consistent with how these document metadata fields are normally treated (as key-value assignments).

2bndy5: Would you like to preserve the `tocdepth` option for backward compatibility? Or do you consider it broken enough to get rid of it?

2bndy5: > metadata fields are normally treated (as key-value assignments).

So what value do you propose? A bool? Or does sphinx use a default value in case none is specified?

jbms: I think they are just parsed as strings, but for bool flags the convention seems to be to just leave the value blank:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#special-metadata-fields

So just:

```rst
:hide-navigation:
```

As far as tocdepth --- thinking about it more, I think we should just keep using tocdepth for this purpose, rather than introducing a new `:hide-toc:` field.  At some point we should actually make non-zero values of `tocdepth` have their proper effect, but for now it seems reasonable to leave it as is.

I think currently, on the master document only the toc depth is limited to 2, and on all other pages the toc depth is unlimited.  This difference is due to the fact that nav_adapt.py uses different logic in those two cases for obtaining the local toc.

2bndy5: I can implement both `hide-toc` while preserving `tocdepth`. In testing the new metadata idea, the users would have to specify 
```rst
:hide-naviagation: anyValue
```
otherwise, no value has no effect.
```rst
.. this has no effect
:hide-naviagation:
```

2bndy5: NM, I got it to work if the metadata is specified at all. If it isn't there, then the feature is not enabled for that page.

2bndy5: I'm going forward with the `hide-toc` idea because it better matches mkdocs-material convention (& its more descriptive). People migrating from mkdocs to sphinx may get confused about the toctree directive's option `maxdepth`. To be clear, `:tocdepth: 0` will still work (just like `hide-toc`).

docs/customization.rst
```diff
@@ -104,12 +104,17 @@
 Customization
 =============
 
-This theme is customized via fields in the `html_theme_options` `dict` in *conf.py*.
-Unlike, newer versions of mkdocs-material theme, this theme also supports the use of a textual
-"hero" section.
+Metadata for a single page
+==========================
+
+Each page can support a set of page-specific options. These are configured using metadata roles.
```
jbms, line 110: Add link to sphinx documentation:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#file-wide-metadata

Could probably use intersphinx as well rather than hard-coding the URL.

Should be called "metadata fields" rather than roles.

docs/customization.rst
```diff
@@ -118,6 +123,34 @@ Unlike, newer versions of mkdocs-material theme, this theme also supports the us
 
         :hero: Configuration options to personalize your site.
 
+.. confval:: hide-navigation
+
+    This theme uses the ``:hide-navigation:`` metadata to hide the page's navigation menu (on
+    large enough viewports only).
```
jbms, line 129: If specified, hides the global navigation sidebar shown on the left side of the page.  By default, the navigation menu is shown if the browser viewport is sufficiently wide.

docs/customization.rst
2bndy5: we are using intersphinx for the sphinx-doc master docs. I had to change a link about `html_logo` because it was defaulting to this theme's <code>:confval:\`html_logo\`</code>

docs/customization.rst
```diff
@@ -118,6 +123,34 @@ Unlike, newer versions of mkdocs-material theme, this theme also supports the us
...
+
+    .. code-block:: rst
+        :caption: Hide the navigation menu like so:
+
+        :hide-navigation:
+
+.. confval:: hide-toc
+
+    This theme uses ``:hide-toc:`` metadata to hide the page's Table of Contents (on large enough
```
jbms, line 138: If specified, hides the local table of contents shown on the right side of the page.  By default the local table of contents is shown if the page contains sub-sections and the browser viewport is sufficiently wide.  If the toc.integrate feature is enabled, this option has no effect.

docs/customization.rst
2bndy5: "left" is relative to the `direction` setting.

jbms: Oh, I wasn't aware of that.  I guess we could say "left (or right in rtl langauge mode)" or leave out the side altogether.  The main thing was that I wanted to clarify that it was the global toc rather than the local toc, as users may not know the distinction between nav vs toc.

docs/customization.rst
2bndy5: I'm not seeing a way to use intersphinx for that link (its not a `confval` or of a domain).

2bndy5: I did find a way to force the `html_logo` link to default to the sphinx docs using intersphinx 🤓

2bndy5:
woah! I just tried switching the `direction` to `rtl`, and there's a bunch of problems... It would seem that the side nav menus don't swap sides despite the direction specified, but I think there's more than a few bugs happening here.

I'll remove the "( or XXX side in `rtd` languages direction)" from the last commit's change to docs. We can work that out when the `direction` is fixed.

jbms (approved):

jbms:
Thanks!

PR #42, opened 2022-03-03
fix autosummary tables and add a note about tabs' toctree

### LINKED ISSUE #33: Tabs require first item to be a page header
I also noticed that a tab item in which the toc entry is just RST links will forward to only the first link in the list. This is exemplified in my CirPy_nRF24 docs' "other links" tab.
[image](https://circuitpython-nrf24l01.readthedocs.io/en/latest/index.html)

RST source code:
```rst
.. toctree::
    :caption: Other Links
    :hidden:

    Download <https://github.com/2bndy5/CircuitPython_nRF24L01/releases/latest>
    CircuitPython Reference Documentation <https://circuitpython.readthedocs.io>
    CircuitPython Support Forum <https://forums.adafruit.com/viewforum.php?f=60>
    Discord Chat <https://adafru.it/discord>
    Adafruit Learning System <https://learn.adafruit.com>
    Adafruit Blog <https://blog.adafruit.com>
    Adafruit Store <https://www.adafruit.com>
```
We should probably put a note in the docs that this convention cannot be supported with navigation tabs. Although it is supported in the side navigation menu.
image

_Originally posted by @2bndy5 in https://github.com/jbms/sphinx-immaterial/issues/31#issuecomment-1048378986_

### LINKED ISSUE #35: Autosummary tables have strange line wrapping
I found that sometimes autosummary tables have very strange line wrapping. Below the functions are wrapped across three lines, even though the second column (function summary) hasn't yet reached the end of the table.

I'm willing to help resolve the issue (to the best of my ability). Can @jbms and/or @2bndy5 point me in the right direction? Is this an HTML decision of when to wrap text in each column?

Source file:

```rst
.. rubric:: Polynomial tests
.. autosummary::
   :toctree:

   is_monic
   is_irreducible
   is_primitive
   is_square_free
```

`sphinx-immaterial` (https://galois.readthedocs.io/en/latest/api/polys.html):

image

`sphinx_rtd_theme` (https://galois.readthedocs.io/en/v0.0.24/api/polys.html):

image

2bndy5:
add rule to fix #35

I also added a note to close #33 (it was just a reminder issue anyway).

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -240,6 +240,16 @@ Configuration Options
         - `navigation.sections <https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-sections>`_
         - `navigation.tabs <https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-tabs>`_
           (only shows for browsers with large viewports)
+
+          .. note::
+              Due to a difference in general document structure between mkdocs and sphinx, any
+              `toctree` entry should start with a page entity (that contains a section header).
+              Sphinx does allow `toctree` entries to be a list of only hyperlinks, but a
+              navigation tab created from such an entry will only lead to the first hyperlink in
+              the `toctree`.
+
+              See `issue #33 <https://github.com/jbms/sphinx-immaterial/issues/33>`_ for a
+              problematic demonstration.
         - `navigation.tabs.sticky <https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#sticky-navigation-tabs>`_
         - `navigation.top <https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#back-to-top-button>`_
         - `navigation.tracking <https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#anchor-tracking>`_
--- a/src/assets/stylesheets/main/_api.scss
+++ b/src/assets/stylesheets/main/_api.scss
@@ -220,7 +220,11 @@ $objinfo-icon-size: 16px;
 // a rule specifically designed for autosummary tables
 // this overrides `.md-typeset code { word-break: break-word; }`
 table.longtable.docutils.data.align-default {
-  tbody>tr>td>p>a.reference.internal>code.xref.py.py-obj.docutils.literal.notranslate>span.pre {
-    word-break: normal;
+  tbody>tr>td>p>a.reference.internal {
+    &>code.xref.py.py-obj.docutils.literal.notranslate{
+      &>span.pre {
+        word-break: normal;
+      }
+    }
   }
 } 
--- a/src/assets/stylesheets/main/_api.scss
+++ b/src/assets/stylesheets/main/_api.scss
@@ -210,3 +210,9 @@ $objinfo-icon-size: 16px;
   border: 1px solid var(--md-default-fg-color--light);
   border-radius: 2px;
 }
+
+// a rule specifically designed for autosummary tables
+.reference.internal .xref.py.py-obj.docutils.literal.notranslate {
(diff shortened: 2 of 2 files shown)


jbms:
It looks like your css rule may match all references to Python objects, not just ones in autosummary tables.  That may not cause any problems, but it might be nicer to only target autosummary tables.

Unfortunately it appears that autosummary does not add any class that may be used to identify it, so we would have to monkey patch e.g. https://github.com/sphinx-doc/sphinx/blob/b60caca9408bb778ec0b277eed105dab8cfae0bb/sphinx/ext/autosummary/__init__.py#L152 to add such a class.

2bndy5:
I could add much more to the selector, but I think a monkey patch would be overkill since other themes don't take that approach.

I'll re-examine the generated autosummary HTML tomorrow and modify the rule. As of right now it only looks for the last 2 matching children (without specifying element tag names).

2bndy5:
I added more to the selector. I could could keep going by adding `.md-typeset` and `code.classes:not(.highlight)`

jbms:
Ultimately I think the issue is that autosummary adds no special styling other than `longtable` which might be used elsewhere, so there is no way to distinguish it from any other table that happens to contain a reference to a Python object in one of its cells.

In general when working on the api documentation styling for this theme/for tensorstore, I noticed that sphinx output lacks appropriate semantic classes in a lot of cases and had to fix that by monkey patching.  I think in large part this is because extensions like autosummary that produce docutils output do not normally coordinate with themes, so they see no immediate benefit from adding extra classes.

2bndy5:
In the case of autosummary, users could use a custom template to add CSS classes aor skip the rubric tables all together,

jbms:
Is your objection to the monkey patching just that you want to avoid added complexity in this theme?

2bndy5:
yes, but I'm lazy like that. I can't really object to your valid points.

2bndy5:
Just pointing out that other theme's don't have a problem with autosummary tables because (as you pointed out) they tend to inherit from a basic CSS - unlike this theme. The real problem this PR aims to fix is overriding the CSS selector from mkdocs-material (which was written in a broad way).

I see this as a problem with mkdocs-material - not as a problem with autosummary. It would seem that `longtable` is mostly for latex writers, but [a repo-wide search](https://github.com/sphinx-doc/sphinx/search?l=Python&q=longtable) reveals that autosummary is the only other python source that uses `longtable` (not sure if that helps).

There's a warning in the [`tabularcolumns` directive doc](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-tabularcolumns) about why longtables are used in latex.

jbms:
Okay, fair enough.

2bndy5:
I still raised this issue upstream. I'm now subscribed to https://github.com/sphinx-doc/sphinx/issues/10234

2bndy5:
The CI skipped upload to pypi (actual) because the commit wasn't tagged.  Is that something I'm allowed to do? I'd probably go with v0.3.1

The [latest is v0.3 on pypi](https://pypi.org/project/sphinx-immaterial/#history); the test pypi index still is getting these updates.

jbms:
Yes feel free to create tagged commits when you think a release is appropriate.

2bndy5:
It looks like there is a CSS fix (class exclusive to autosummary tables) slated for Sphinx v4.5.0. I'm weary of pinning this theme to Sphinx v4.5.0 just to satisfy this CSS specificity problem because it seems like an edge case to me. I guess we could recommend users require Sphinx v4.5.0 (when it is released). However, we would need to alter the rule to somehow optionally look for the new CSS class `autosummary`.

Would it make more sense to monkey patch autosummary tables for older extension versions prior to Sphinx v4.5.0 and adjust the CSS rule accordingly?

Issue #43, opened 2022-03-13
Content tabs


mhostetter:
Are [content tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/) supported in this theme? I searched in the docs here, but didn't find any mention. Thanks!

2bndy5:
No. Not natively. To use content tabs in sphinx-immaterial, we recommend the sphinx-design extension. There are other extensions that get similar results, we use sphinx-design in our docs.

2bndy5:
We can't support content tabs from the theme because the mkdocs-material theme uses a markdown extension to support content tabs. Since this markdown extension isn't used at all in the sphinx world (myRST parser doesn't even use it), we are forced to use a different solution (like the sphinx-design extension).

The drawback, as you've already noted, is that the CSS styling doesn't align well with the theme's CSS.

jbms:
We could potentially add some type of sphinx-design integration so that the content tabs are styled consistently.

2bndy5:
> We could potentially add some type of sphinx-design integration so that the content tabs are styled consistently.

I've been avoiding this idea because it would pin us to using sphinx-design exclusively (there are other/simpler implementations out there). And, there's the matter of overriding CSS from the extension; I'm not sure where extensions' CSS are in the cascade relative to the theme's CSS.

mhostetter:
Is it easier/better to add a Sphinx directive in this theme, such as `.. content-tabs::` (which could have similar syntax to `sphinx-design`), that would produce a tabbed window with the Material for Mkdocs theming?

2bndy5:
> Is it easier/better to add a Sphinx directive in this theme, such as `.. content-tabs::` (which could have similar syntax to `sphinx-design`), that would produce a tabbed window with the Material for Mkdocs theming?

I'm more in favor of this approach since we had to do something like that to support admonitions without a title. We're also considering something like this for line number styles in literal code blocks

2bndy5:
Just want to make note that synchronizing the common tabs on a certain page requires JS (which we may already have from upstream).

jbms:
Providing our own implementation of content tabs seems reasonable assuming it is not too complicated.

I think in principle it might be possible to synchronize the tabs on a single page without JavaScript using fancy hidden form elements or something, but I think certainly synchronizing across multiple pages would require JavaScript in any case.

[...]

PR #44, opened 2022-03-17
Merge upstream v8.2.5

2bndy5:
This merges in updates from upstream up to latest release (not including the bleeding edges).

The only major conflict I had to resolve was in _src/assets/javascripts/components/search/query/index.ts_

<details><summary>Here was the conflicted diff</summary>

```diff
diff --git a/src/assets/javascripts/components/search/query/index.ts b/src/assets/javascripts/components/search/query/index.ts
index fa360ad63..951b6f2c0 100644
--- a/src/assets/javascripts/components/search/query/index.ts
+++ b/src/assets/javascripts/components/search/query/index.ts
@@ -43,7 +43,8 @@ import { translation } from "~/_"
 import {
   getLocation,
   setToggle,
-  watchElementFocus
+  watchElementFocus,
+  watchToggle
 } from "~/browser"
 
 import { Component } from "../../_"
@@ -84,12 +85,41 @@ export function watchSearchQuery(
   let param$: Observable<string>
   if (searchParams.has("q")) {
     setToggle("search", true)
+<<<<<<< ours
     const value = el.value = searchParams.get("q")!
     el.focus()
     param$ = of(value)
   } else {
     param$ = of()
   }
+=======
+
+  /* Intercept query parameter (deep link) */
+  const param$ = rx$
+    .pipe(
+      filter(isSearchReadyMessage),
+      take(1),
+      map(() => searchParams.get("q") || "")
+    )
+
+  /* Remove query parameter when search is closed */
+  watchToggle("search")
+    .pipe(
+      filter(active => !active),
+      take(1)
+    )
+      .subscribe(() => {
+        const url = new URL(location.href)
+        url.searchParams.delete("q")
+        history.replaceState({}, "", `${url}`)
+      })
+
+  /* Set query from parameter */
+  param$.subscribe(value => { // TODO: not ideal - find a better way
+    if (value)
+      el.value = value
+  })
+>>>>>>> theirs
 
   /* Intercept focus and input events */
   const focus$ = watchElementFocus(el)
```

</details>

Given how you mentioned altering the search feature, I opted to not merge in the changes to this file. I have no idea what the upstream changes were doing, but I have a feeling it had something to do with URI search query parameter. The [git diff on upstream](https://github.com/squidfunk/mkdocs-material/compare/8.2.1...8.2.5) only shows [changes related to `watchToggle`](https://github.com/squidfunk/mkdocs-material/compare/8.2.1...8.2.5#diff-36117e5373aef262c0fc3e0d37c56bec5d51de771f9d91a7100aa5def753dfb8). So, I can add that in if it is ok.

---
## The merger script
- a bit more verbose when
   1. `git` fails in `subprocess.run()` calls
   2. informing user of which task is being performed
- ignores some large folders when copying the current sphinx-immaterial repo. These are folders that are in the gitignore anyway.
- formatted with `black` and compliant with pylint. Most changes here were `'` into `"` and whitespace stuff. I also sequestered the main portion of the script into a `main()` function to avoid any problems with global variables (which seem to share the same name as some scoped variables) being changed by pre-existing functions.

---

Lastly, I added your very helpful explanation comment to the contributing guideline after mentioning the merger script.


jbms:
Thanks.

That conflict in query/index.ts we do want to resolve.  That is to address this issue I filed:
https://github.com/squidfunk/mkdocs-material/issues/3622

You should add back in this part:

```
+
+  /* Remove query parameter when search is closed */
+  watchToggle("search")
+    .pipe(
+      filter(active => !active),
+      take(1)
+    )
+      .subscribe(() => {
+        const url = new URL(location.href)
+        url.searchParams.delete("q")
+        history.replaceState({}, "", `${url}`)
+      })
```

Note: Due to the way the merge script works, if we leave out a change, it won't get merged in later either.

2bndy5:
> Note: Due to the way the merge script works, if we leave out a change, it won't get merged in later either.

Yeah that's why I posted the diff and mentioned it first. I will add the `watch Toggle` stuff in.

jbms:
Thanks, looks good to me.

Have you tested if mermaid actually works?

Note: I missed it before, but I see that mkdocs-material loads mermaid.js directly from a CDN with no way to override.  It would be better to bundle the mermaid code directly, or at least have the option to do so, so that we aren't loading code from third-party servers.  We could take care of that in the future, though.

2bndy5:
I did a quick test of the mermaid feature on a different branch (locall only), but it didn't work (probably due to the needed changes from v8.2.5 related to hosting from `file://`). I did find that there is [a sphinx extension that adds a `.. mermaid::` directive](https://github.com/mgaitan/sphinxcontrib-mermaid), but it has several issues about loading the mermaid.js multiple times 👎🏼 . If/when that extension matures, I wouldn't mind just using that instead of maintaining our own (without autodoc API support) as long as the upstream CSS is still used.

As for the loading of mermaid.js from a CDN, I also found this strange, but I think it was done like this to reduce the bundled JS. I also want to point out that all the google fonts are fetched from CDN as well, so I don't really have a problem with the current approach (its just not what I would've done). Given the youth of mermaid.js, I wouldn't mind not doing any changes to the script loading - at least not for a couple years.

jbms:
We could put mermaid.js in a separate bundle so that it is only loaded when needed.  It would be better to also not use the CDN for the fonts, though they are safer in that we are only loading CSS, not JavaScript,.  It is possible to extract some information from a website using malicious CSS but I think it would be impossible to steal login credentials, etc.

2bndy5:
> It would be better to also not use the CDN for the fonts

I did try doing this a while back on a different project for the material icon font(s). I foresee some added problems:

1. If the user changes the font for either `code` or `text`, then we'd have to make sure the specified font is available from the theme's assets. Note there are hundreds of fonts hosted from google (many just duplicates with different weight/DPI).
2. We'd have to keep our local copies updated (and preferably converted to woff format which significantly saves on bundle size).

2bndy5:
I just realized I forgot to update the package.lock on remote. I can do this on my next PR about #43 & #15

jbms:
Oh yeah, I noticed that at one point but forgot about it.

Unless you have the next PR ready soon it might be better to fix package-lock.json first so that the repository is in a consistent state.

We should probably also add a github actions check to verify it.

(i.e. that package-lock.json is unmodified after running npm install)

2bndy5:
> We should probably also add a github actions check to verify it.

Good idea!

I can open a PR to resolve #43 today, but I wanted to also squash #15. Unfortunately, I got no immediate response from the https://github.com/tk0miya/sphinxcontrib-details-directive/issues/4, so maybe I'll just roll my fix for that directive into the theme (for now) and wade through the Apache License reqs.

jbms:
I think tk0miya is mostly only available for sphinx work on the weekends.

You could fix that directive via monkey patching rather than copying all of the code, and then there wouldn't really be any license issues.

2bndy5:
Hmm, my initial idea to monkey patch it may require the directive not be optional. Unless I peek at the extensions list during `__init__.py` then apply the patch as needed...

jbms:
You could try to import it, and catch the ImportError if it fails to import.

PR #45, opened 2022-03-18
Custom extensions

### LINKED ISSUE #15: CSS for details/summary html tags is tied to admonitions' CSS

### LINKED ISSUE #26: line numbers could use better visual distinction
At first, this felt like a personal style choice, but I've been adding custom CSS to all my docs that use this theme.

```css
.linenos {
  background-color: var(--md-default-bg-color--light);
  margin-right: 0.5rem;
}
```
which transforms 
image
into something a little more "legible" like so
image

---
I was wondering if this CSS change should be considered for this theme. Additionally, I kinda like how line numbers look in the mkdocs-material upstream
image

### LINKED ISSUE #43: Content tabs

2bndy5:
# New Directives
These new directives are automatically added to doc builds because they are tied to our CSS:

- `md-tab-set` and `md-tab-item` which resolves #43 (includes a new doc page to demonstrate CSS `class`/`id` usage)
- `md-mermaid` to expose the implementation merged in from upstream (includes a new doc page to demonstrate CSS `class`/`id` usage).

# Patches
- monkey-patch for the `sphinxcontrib-details-directive` which is applied only when the directive is installed. This patch allows us to use the `:class:` option to specify styling for collapsible admonitions. resolves #15 again
- updates _package.lock_ from changes in #44 
- auto-asserts line numbers as a table's column in code-blocks - resolves #26 until sphinx v6.0 is released
- removes dependency on `sphinx-design` in docs' builds, but adds `sphinxcontrib-details-directive` to the docs' extensions/reqs.

## Known bug (very minor annoyance)
Upon page refresh, the tabs' highlighted bottom border always resets to the first tab (unexpected behavior), but the displayed tab content and highlighted tab label do not reset (expected behavior). I suspect it has something to do with how the JS loads on page refresh, but I'm not too concerned at this point.

### before refresh
image

### after refresh
image

### DIFF
--- a/docs/content_tabs.rst
+++ b/docs/content_tabs.rst
@@ -1,7 +1,7 @@
 .. _sphinx-design tabs: https://sphinx-design.readthedocs.io/en/furo-theme/tabs.html
 
-Using tabbed content
-====================
+Content tabs
+============
 
 .. note::
     This document focused on content tabs, not navigation tabs.
--- a/docs/content_tabs.rst
+++ b/docs/content_tabs.rst
@@ -1,0 +1,128 @@
+.. _sphinx-design tabs: https://sphinx-design.readthedocs.io/en/furo-theme/tabs.html
+
+Using tabbed content
+====================
+
+.. note::
+    This document focused on content tabs, not navigation tabs.
+
+Use of `content tabs in mkdocs-material <https://squidfunk.github.io/mkdocs-material/reference/content-tabs/>`_
+theme relies on a markdown extension that isn't used in the world of Sphinx. Instead,
+the sphinx-immaterial theme provides its own directives to make use of content tabs.
+
+.. code-block:: python
+    :caption: These directives require an added entry in conf.py's extension list:
+
+    extensions = ["sphinx_immaterial", "sphinx_immaterial.content_tabs"]
+
+.. admonition:: Linked Tabs
+    :class: missing
+
+    The `linked content tabs <https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#linked-content-tabs>`_
+    seen in mkdocs-material is not supported until that feature transitions from the mkdocs-material theme's insider
+    releases to it's public releases.
+
+    However, you can use other sphinx extensions (like `sphinx-design tabs`_) to achieve this functionality.
+    Although, other extensions will require some custom CSS styling to match the mkdocs-material
+    theme's styling for content tabs.
+
+.. confval:: md-tab-set
+
+    Each set of tabs on a page must begin with a `md-tab-set` directive. This directive
+    only accepts children that `md-tab-item` directives.
+
+    This directive supports ``:class:`` and ``:name:`` options to use custom CSS classes
+    and reference links (respectively)
+
+    .. code-block:: rst
+
+        .. md-tab-set::
+            :class: custom-tab-set-style
+            :name: ref_this_tab_set
+
+            .. md-tab-item:: Local Ref
+
+                A reference to this tab set renders like so:
+                `tab set description <ref_this_tab_set>`.
+                
+                This syntax can only be used on the same page as the tab set.
+
+            .. md-tab-item:: Cross-page Ref
+
+                To cross-reference this tab set from a different page, use
+                :ref:`tab set description <ref_this_tab_set>`
+
+                Use this syntax to give a tab set a custom description.
+                Clearly this also works on the same page.
+
+            .. md-tab-item:: Custom CSS
+
+                .. literalinclude:: _static/extra_css.css
+                    :language: css
+                    :start-at: /* ************************ custom-tab-set-style
+                    :end-before: /* *********************** custom-tab-item-style
+
+    .. md-tab-set::
+        :class: custom-tab-set-style
+        :name: ref_this_tab_set
+
+        .. md-tab-item:: Local Ref
+
+            A reference to this tab set renders like so:
+            `tab set description <ref_this_tab_set>`.
+            
+            This syntax can only be used on the same page as the tab set.
+
+        .. md-tab-item:: Cross-page Ref
+
+            To cross-reference this tab set from a different page, use
+            :ref:`tab set description <ref_this_tab_set>`
+
+            Clearly this also works on the same page as the tab set.
+
+        .. md-tab-item:: Custom CSS
+
+            .. literalinclude:: _static/extra_css.css
+                :language: css
+                :start-at: /* ************************ custom-tab-set-style
+                :end-before: /* *********************** custom-tab-item-style
+
+.. confval:: md-tab-item
+
+    This directive is used to create a tab within a set of content tabs. It requires a
+    label as it's argument. Additionally, it also supports the ``:class:`` option, to
+    optionally provide custom CSS classes to the tab's content (not the tab's label).
+
+    .. code-block:: rst
+
+        .. md-tab-set::
+
+            .. md-tab-item:: Customized content
+                :class: custom-tab-item-style
+
+                This content could be styled differently from other page content.
+
+            .. md-tab-item:: Custom CSS
+
+                .. literalinclude:: _static/extra_css.css
+                    :language: css
+                    :start-at: /* *********************** custom-tab-item-style
+                    :end-before: /* ************************* inline icon stuff
+
+    .. md-tab-set::
+
+        .. md-tab-item:: Customized content
+            :class: custom-tab-item-style
+
+            This content could be styled differently from other page content.
+
+        .. md-tab-item:: Custom CSS
+
+            .. literalinclude:: _static/extra_css.css
+                :language: css
+                :start-at: /* *********************** custom-tab-item-style
+                :end-before: /* ************************* inline icon stuff
+
+Typical examples are seen in this documentations'
+`Custom admonitions <admonitions.html#custom-admonitions>`_ and
+:ref:`Version Information Structure <version_info_example>` sections.
--- a/docs/content_tabs.rst
+++ b/docs/content_tabs.rst
@@ -4,9 +4,9 @@ Content tabs
 ============
 
 .. note::
-    This document focused on content tabs, not navigation tabs.
+    This document discusses content tabs, not navigation tabs.
 
-Use of `content tabs in mkdocs-material <https://squidfunk.github.io/mkdocs-material/reference/content-tabs/>`_
+Use of `content tabs in the mkdocs-material <https://squidfunk.github.io/mkdocs-material/reference/content-tabs/>`_
 theme relies on a markdown extension that isn't used in the world of Sphinx. Instead,
 the sphinx-immaterial theme provides its own directives to make use of content tabs.
 
@@ -14,17 +14,30 @@ the sphinx-immaterial theme provides its own directives to make use of content t
     :class: missing
 
     The `linked content tabs <https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#linked-content-tabs>`_
-    seen in mkdocs-material is not supported until that feature transitions from the mkdocs-material theme's insider
-    releases to it's public releases.
+    feature seen in mkdocs-material is not supported until that feature transitions from the mkdocs-material theme's insider
+    releases to its public releases.
 
-    However, you can use other sphinx extensions (like `sphinx-design tabs`_) to achieve this functionality.
-    Although, other extensions will require some custom CSS styling to match the mkdocs-material
+    You can use other sphinx extensions (like `sphinx-design tabs`_) to achieve this functionality.
+    However, other extensions will require some custom CSS styling to match the mkdocs-material
     theme's styling for content tabs.
 
-.. confval:: md-tab-set
+.. rst:directive:: md-tab-set
 
     Each set of tabs on a page must begin with a `md-tab-set` directive. This directive
     only accepts children that are `md-tab-item` directives.
+
+    .. rst:directive:option:: class
+        :type: string
+        
+        A space delimited list of qualified names that get used as the HTMl element's
+        ``class`` attribute.
+
+    .. rst:directive:option:: name
+        :type: string
+        
+        A qualified name that get used as the HTML element's ``id`` attribute.
+    
+        Use the `ref` role to reference the element by name.
 
     This directive supports ``:class:`` and ``:name:`` options to use custom CSS classes
     and reference links (respectively)
@@ -81,13 +94,35 @@ the sphinx-immaterial theme provides its own directives to make use of content t
                 :start-at: /* ************************ custom-tab-set-style
                 :end-before: /* *********************** custom-tab-item-style
 
-.. confval:: md-tab-item
+.. rst:directive:: md-tab-item
 
     This directive is used to create a tab within a set of content tabs. It requires a
-    label as it's argument. Additionally, it also supports the ``:class:`` option, to
-    optionally provide custom CSS classes to the tab's content (not the tab's label).
+    label as it's argument. 
 
-    .. code-block:: rst
+    .. rst:directive:option:: class
+        :type: string
+        
+        A space delimited list of qualified names that get used as the HTMl element's
+        ``class`` attribute.
+
+        Use the ``:class:`` option to optionally provide custom CSS classes to the tab's content
+        (not the tab's label).
+
+        .. code-block:: rst
+
+            .. md-tab-set::
+
+                .. md-tab-item:: Customized content
+                    :class: custom-tab-item-style
+
+                    This content could be styled differently from other page content.
+
+                .. md-tab-item:: Custom CSS
+
+                    .. literalinclude:: _static/extra_css.css
+                        :language: css
+                        :start-at: /* *********************** custom-tab-item-style
+                        :end-before: /* ************************* inline icon stuff
 
         .. md-tab-set::
 
@@ -102,20 +137,6 @@ the sphinx-immaterial theme provides its own directives to make use of content t
                     :language: css
                     :start-at: /* *********************** custom-tab-item-style
                     :end-before: /* ************************* inline icon stuff
-
-    .. md-tab-set::
-
-        .. md-tab-item:: Customized content
-            :class: custom-tab-item-style
-
-            This content could be styled differently from other page content.
-
-        .. md-tab-item:: Custom CSS
-
-            .. literalinclude:: _static/extra_css.css
-                :language: css
-                :start-at: /* *********************** custom-tab-item-style
-                :end-before: /* ************************* inline icon stuff
 
 Typical examples are seen in this documentations'
 `Custom admonitions <admonitions.html#custom-admonitions>`_ and
--- a/docs/content_tabs.rst
+++ b/docs/content_tabs.rst
@@ -24,7 +24,7 @@ the sphinx-immaterial theme provides its own directives to make use of content t
 .. confval:: md-tab-set
 
     Each set of tabs on a page must begin with a `md-tab-set` directive. This directive
-    only accepts children that `md-tab-item` directives.
+    only accepts children that are `md-tab-item` directives.
 
     This directive supports ``:class:`` and ``:name:`` options to use custom CSS classes
     and reference links (respectively)
@@ -47,7 +47,7 @@ the sphinx-immaterial theme provides its own directives to make use of content t
                 To cross-reference this tab set from a different page, use
                 :ref:`tab set description <ref_this_tab_set>`
 
-                Clearly this also works on the same page.
+                Clearly, this also works on the same page.
 
             .. md-tab-item:: Custom CSS
 
@@ -72,7 +72,7 @@ the sphinx-immaterial theme provides its own directives to make use of content t
             To cross-reference this tab set from a different page, use
             :ref:`tab set description <ref_this_tab_set>`
 
-            Clearly this also works on the same page as the tab set.
+            Clearly, this also works on the same page as the tab set.
 
         .. md-tab-item:: Custom CSS
 
--- a/docs/content_tabs.rst
+++ b/docs/content_tabs.rst
@@ -52,7 +52,6 @@ the sphinx-immaterial theme provides its own directives to make use of content t
                 To cross-reference this tab set from a different page, use
                 :ref:`tab set description <ref_this_tab_set>`
 
-                Use this syntax to give a tab set a custom description.
                 Clearly this also works on the same page.
 
             .. md-tab-item:: Custom CSS
--- a/docs/content_tabs.rst
+++ b/docs/content_tabs.rst
@@ -10,11 +10,6 @@ Use of `content tabs in mkdocs-material <https://squidfunk.github.io/mkdocs-mate
 theme relies on a markdown extension that isn't used in the world of Sphinx. Instead,
 the sphinx-immaterial theme provides its own directives to make use of content tabs.
 
-.. code-block:: python
-    :caption: These directives require an added entry in conf.py's extension list:
-
-    extensions = ["sphinx_immaterial", "sphinx_immaterial.content_tabs"]
-
 .. admonition:: Linked Tabs
     :class: missing
 
--- a/docs/content_tabs.rst
+++ b/docs/content_tabs.rst
@@ -40,7 +40,7 @@ the sphinx-immaterial theme provides its own directives to make use of content t
         Use the `ref` role to reference the element by name.
 
     This directive supports ``:class:`` and ``:name:`` options to use custom CSS classes
-    and reference links (respectively)
+    and reference links (respectively).
 
     .. code-block:: rst
 
--- a/docs/mermaid_diagrams.rst
+++ b/docs/mermaid_diagrams.rst
@@ -8,16 +8,29 @@ Mermaid diagrams
 
 The mkdocs-material theme is equipped to make use of diagrams generated (during page load time)
 with `mermaid.js`_. Although, its implementation relies on a markdown extension that does not get
-used by this sphinx-immaterial theme. Thus, the sphinx-immaterial theme provides with an optional
+used by this sphinx-immaterial theme. Thus, the sphinx-immaterial theme provides an optional
 directive that exposes the underlying implementation in mkdocs-material theme.
 
-.. confval:: md-mermaid
+.. rst:directive:: md-mermaid
 
-    The `md-mermaid` directive does support the ``:class:`` and ``:name:`` options which can used
+    .. rst:directive:option:: class
+        :type: string
+        
+        A space delimited list of qualified names that get used as the HTMl element's
+        ``class`` attribute.
+
+    .. rst:directive:option:: name
+        :type: string
+        
+        A qualified name that get used as the HTML element's ``id`` attribute.
+    
+        Use the `ref` role to reference the element by name.
+
+    The `md-mermaid` directive's ``:class:`` and ``:name:`` options can be used
     as respective class and id specifiers in custom CSS.
 
-    This theme comes with CSS styling that conforms to the chosen primary & accent colors
-    (based on the selected scheme).
+    This theme comes with CSS styling that conforms to the chosen `primary` & `accent` colors
+    (based on the selected `scheme`).
 
     .. md-admonition::
         :class: missing
--- a/docs/mermaid_diagrams.rst
+++ b/docs/mermaid_diagrams.rst
@@ -16,7 +16,7 @@ directive that exposes the underlying implementation in mkdocs-material theme.
     .. rst:directive:option:: class
         :type: string
         
-        A space delimited list of qualified names that get used as the HTMl element's
+        A space delimited list of qualified names that get used as the HTML element's
         ``class`` attribute.
 
     .. rst:directive:option:: name
--- a/docs/mermaid_diagrams.rst
+++ b/docs/mermaid_diagrams.rst
@@ -19,29 +19,27 @@ directive that exposes the underlying implementation in mkdocs-material theme.
 
         extensions = ["sphinx_immaterial", "sphinx_immaterial.mermaid_diagrams"]
 
-    The `md-mermaid` directive does support the ``:class:`` option, but it has no affect after
-    the graph is generated.
+    The `md-mermaid` directive does support the ``:class:`` and ``:name:`` options which can used
+    as respective class and id specifiers in custom CSS.
 
     This theme comes with CSS styling that conforms to the chosen primary & accent colors
-    (based on the selected scheme). To modify the CSS for a particular graph, you should use the
-    mechanics available in `mermaid.js`_. For example, each graph's node has a unique ``id``
-    and generic ``class`` attribute that can be used in a CSS rule's selector.
+    (based on the selected scheme).
 
     .. md-admonition::
         :class: missing
     
         While all `mermaid.js`_ features should work out-of-the-box, This theme will currently only
-        adjust the fonts and colors for `flowcharts`_, `sequence diagrams`_, `class diagrams`_,
-        `state diagrams`_ and `entity-relationship diagrams`_.
+        adjust the fonts and colors for `flowcharts`_, `sequence diagrams <sequence-diagrams>`,
+        `class diagrams <class-diagrams>`, `state diagrams <state-diagrams>`, and
+        `entity-relationship diagrams <entity-relationship-diagrams>`.
 
-.. _flowcharts:
-
-flowcharts
+Using flowcharts
 ----------------
 
 .. code-block:: rst
 
     .. md-mermaid::
+        :name: flowcharts
 
         graph LR
             A[Start] --> B{Error?};
@@ -51,6 +49,7 @@ flowcharts
             B ---->|No| E[Yay!];
 
 .. md-mermaid::
+    :name: flowcharts
 
     graph LR
         A[Start] --> B{Error?};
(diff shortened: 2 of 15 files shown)


2bndy5:
I've started seeing a deprecation warning when executing `pip setup.py bdist_wheel` cmd:
```
/path/to/sphinx-immaterial/setup.py:21: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils.command.build
```
Seems to only appear in python v3.10+ or could be caused by pip v22.0.4

docs/content_tabs.rst
```diff
@@ -0,0 +1,122 @@
+.. _sphinx-design tabs: https://sphinx-design.readthedocs.io/en/furo-theme/tabs.html
+
+Content tabs
+============
+
+.. note::
+    This document focused on content tabs, not navigation tabs.
```
jbms, line 7: "This document discusses content tabs, not navigation tabs."

docs/content_tabs.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Use of `content tabs in mkdocs-material <https://squidfunk.github.io/mkdocs-material/reference/content-tabs/>`_
```
jbms, line 9: "in the mkdocs-material"

docs/content_tabs.rst
```diff
@@ -0,0 +1,122 @@
...
+theme relies on a markdown extension that isn't used in the world of Sphinx. Instead,
+the sphinx-immaterial theme provides its own directives to make use of content tabs.
+
+.. admonition:: Linked Tabs
+    :class: missing
+
+    The `linked content tabs <https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#linked-content-tabs>`_
```
jbms, line 16: "feature seen in mkdocs-material"

"to its public releases."

docs/content_tabs.rst
```diff
@@ -0,0 +1,122 @@
...
+    seen in mkdocs-material is not supported until that feature transitions from the mkdocs-material theme's insider
+    releases to it's public releases.
+
+    However, you can use other sphinx extensions (like `sphinx-design tabs`_) to achieve this functionality.
```
jbms, line 20: "You can use other..."  "However, other extensions "

docs/content_tabs.rst
```diff
@@ -0,0 +1,122 @@
...
+    Although, other extensions will require some custom CSS styling to match the mkdocs-material
+    theme's styling for content tabs.
+
+.. confval:: md-tab-set
```
jbms, line 24: Rather than confval, we should use the rst domain provided by sphinx:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-restructuredtext-domain

docs/mermaid_diagrams.rst
```diff
@@ -0,0 +1,207 @@
+.. _mermaid.js: https://mermaid-js.github.io/mermaid/
+
+Mermaid diagrams
+================
+
+.. note::
+    Use of this feature has no affect or affiliation with sphinx's graphviz implementation.
+
+The mkdocs-material theme is equipped to make use of diagrams generated (during page load time)
```
jbms, line 9: Perhaps put this as a `.. note` at the bottom, since mkdocs-material's implementation isn't terribly relevant to users of this theme, except I suppose if they are migrating.

docs/mermaid_diagrams.rst
```diff
@@ -0,0 +1,207 @@
...
+        adjust the fonts and colors for `flowcharts`_, `sequence diagrams <sequence-diagrams>`,
+        `class diagrams <class-diagrams>`, `state diagrams <state-diagrams>`, and
+        `entity-relationship diagrams <entity-relationship-diagrams>`.
+
+Using flowcharts
+----------------
+
+.. code-block:: rst
+
+    .. md-mermaid::
```
jbms, line 35: Note: it might be nice to add a directive for us in our documentation that avoids having to duplicate the code for samples like this.

setup.py
```diff
@@ -17,20 +17,19 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-import setuptools
```
jbms, line 20: setuptools should always be imported first because it does some monkey patching of distutils.

We should add a comment to indicate that.

sphinx_immaterial/mermaid_diagrams.py
```diff
@@ -0,0 +1,59 @@
...
+    attributes = {"class": "mermaid"}
+    self.body.append(self.starttag(node, "pre", **attributes))
+
+
+def depart_mermaid_node(self, node: mermaid_node):
+    self.body.append("</pre>")
+
+
+def visit_mermaid_node_latex(self, node: mermaid_node):
+    self.body.append('<pre class="mermaid">')
```
jbms, line 50: This doesn't work, right?

This theme isn't intended to be used except with the html builder, so I don't think we need to worry about latex.

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,220 @@
...
+
+            tab_set.children = children
+            tab_set += tab_label_div
+            tab_set += tab_content_div
+
+
+def setup(app: Sphinx) -> None:
+    app.add_directive("md-tab-set", MaterialTabSetDirective)
+    app.add_directive("md-tab-item", MaterialTabItemDirective)
+    app.add_post_transform(MaterialTabSetHtmlTransform)
```
jbms, line 217: I think it would be best to avoid using `add_post_transform`, since that means an additional full traversal of every document.  I think the main thing you are using the post transform for is assigning a unique index to each tab set --- for that you could instead use `env.temp_data` (which is cleared for each document) or handle it in the writer (i.e. the `visit_` functions), where you could store the counter as an attribute on the writer.

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,220 @@
...
+            type="",
+            classes=["tabbed-block"] + self.options.get("class", []),
+        )
+        self.state.nested_parse(self.content, self.content_offset, tab_content)
+        tab_item += tab_content
+
+        return [tab_item]
+
+
+class content_tab_input(nodes.Element, nodes.General):
```
jbms, line 122: I think it would make more sense to have a `content_tab_item` node, and then handle the input and label stuff in the html translation only.

jbms:
Overall, great work here!

I think we should try to fix the indicator line beneath the tab.  Probably adding some console.log statements to the javascript code may help in figuring out what is going on.

jbms:
I looked into the indicator line issue -- turns out it is a bug in mkdocs-material which appears to have been fixed in the insiders version.

I reported it here:
https://github.com/squidfunk/mkdocs-material/issues/3744

sphinx_immaterial/mermaid_diagrams.py
2bndy5: It does work with latex writer. That's why I have the CI building both html and latex. In fact the RTD builds are making PDFs from the latex builds.

The mermaid feature won't work with latex since it relies on JS. In fact if the JS in a browser isn't loaded, then the mermaid directives look like non-highlighted code blocks. This is what I'm aiming for with the latex writer.

This function is important because the `starttag()` used in html doesn't exist in the latex builder.

jbms: Oh I see ---  putting the mermaid code literally in the output seems reasonable enough for latex (rendering the diagrams would be better but of course that would be a lot more trouble to accomplish), but why include HTML like `<pre class="mermaid">`?

sphinx_immaterial/content_tabs.py
2bndy5: >  I think the main thing you are using the post transform for is assigning a unique index to each tab set

Not true. The transformer, takes the tab item and splits it up into a label, input, and content's div elements. For each tab set the labels and and content's div elements go in their own div containers, while the input elements go directly after the div for the tab set but before the label and content groups.

I couldn't think of a way to organize these elements from the `MaterialTabItemDirective.run()` And, there was the additional pain of figuring out the `id`, `for`, and `name` attributes value; I didn't know about the `env.temp_data`.

sphinx_immaterial/mermaid_diagrams.py
2bndy5: that's just what it does for html. I don't really know latex syntax...

Actually the mermaid directive results in a
```html
<div class="mermaid-diagram">
  <pre class="mermaid"> <!--this part is reuired for the mermaid.js to work-->
    <code>
      <!-- all special markup language characters are converted -->
    </code>
  </pre>
</div>
```

sphinx_immaterial/content_tabs.py
jbms: Okay, I see.  Thinking about this more, I think this could all be handled in `visit_tab_set`:

- Store the counter for ids in the HTMLTranslator (i.e. `self`)
- In `visit_tab_set`, emit the opening div, then emit the input tag for each child, then emit the opening tag for the labels.  Then emit the opening label tag for each child, and call `child_label.walkabout(self)` to render the label.
- After emitting the labels, do a similar thing for the content.
- Emit the closing tag.
- raise docutils.nodes.SkipNode to prevent docutils from itself recursing on the children.

sphinx_immaterial/content_tabs.py
2bndy5: I need the writer to explicitly use different tag names for the input & label elements. Are you saying there's a way to do that without the `stattag()` in a visitor function?

Also the attributes of a node aren't directly output as attributes of an element. These visitor functions take care of adding peculiar attributes like `for`,

jbms: starttag just formats an HTML tag:
https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/writers/_html_base.py#l411

But why can't you use it?

setup.py
2bndy5: distutils is going away; see https://github.com/jbms/sphinx-immaterial/pull/45#issuecomment-1072288302

until we phase out the use of distutils, I could add a pylint comment to ignore the import ordering, here.

sphinx_immaterial/content_tabs.py
2bndy5: Interesting challenge. Much of this code is heavily modified from the `sphinx-design` ext. I didn't bother with attribution because the `sphinx-design` uses MIT lisence, and the code changed so much from the initial copy.

Clearly, I have a lot to learn about `docutils` - I wish they had better/complete docs.

sphinx_immaterial/content_tabs.py
2bndy5: > But why can't you use it?

I think I worded my previous comment wrong. I need to use the `starttag()`. Is there a way to change the html tag without using `starttag()`?

docs/mermaid_diagrams.rst
2bndy5: The `sphinx-design` docs use external files to store the duplicated snippets. The external files are also used in test runs via `pytest`. Their docs are a bit more laborious because they also support the myRST parser (which involves duplicated markdown versions of rst snippets).

sphinx_immaterial/content_tabs.py
jbms: Basically the way the document gets converted to HTML is that the HTMLTranslator class (defined in docutils but extended by sphinx) has a `self.body` member which is just a list of strings that will get concatenated together to form the final HTML output..

Then docutils traverses the document using the walkabout function, using the HTMLTranslator object as a visitor.

In these visit methods you can just append any arbitrary text that you like, there isn't any restriction.  Additionally, in the `visit_xxx` function if you have already handled the entire content of the element you can raise `docutils.nodes.SkipNode` to prevent docutils from recursing on the children and calling the `depart_xxx` function.  So that allows you to have full control over the output.

The `starttag` method is just a utility function for formatting an HTML start tag so you can use it as you like, but you can also just format the HTML directly.  There is also a `self.attval` function for escaping an HTML attribute.

setup.py
jbms: I think distutils isn't exactly going away, it is just being moved out of the Python standard library into setuptools itself.

Furthermore, I think setuptools modifies `sys.modules` so that you can still import distutils using the normal path:
https://github.com/pypa/setuptools/blob/main/_distutils_hack/__init__.py

Does the deprecation warning still appear if you fix the import order?

sphinx_immaterial/content_tabs.py
jbms: Yeah unfortunately the docs are rather limited --- I have found the only way to really understand docutils and sphinx is to study the source code.

docs/mermaid_diagrams.rst
jbms: Yeah it was just a suggestion --- I didn't imagine it as an extension provided by this theme to users, but perhaps just something defined for our own documentation, like `confval`.  But it is true it is widely useful for other sphinx extensions, though maybe some customization is needed for different uses.

setup.py
2bndy5: pylint realizes the setuptools is not in std lib, so it complains about the order. Using
```python

import setuptools  # pylint: disable=wrong-import-order

import atexit
import distutils.command.build
import os
import subprocess
import tempfile

import setuptools.command.build_py
import setuptools.command.develop
import setuptools.command.install
import setuptools.command.sdist
```
doesn't seem to trigger the deprecatin warning. 🤷🏼

sphinx_immaterial/content_tabs.py
2bndy5: Agreed, its just **_a lot_** to study. Also, its not a good sign when a docs-based library lacks on its own docs... 🤣

docs/content_tabs.rst
2bndy5: Done.

This is the first time I've actually used the rst domain, so let me know if there are any tweaks you'd like to see,

docs/mermaid_diagrams.rst
2bndy5: Its more of an introductory note. I could hide it into a collapsible admonition titled "Implementation Note".

There is a similar introductory statement in practically all supported MD extensions (in mkdocs) that get special attention for this port of the theme. This was added specifically for those migrating from mkdocs.

setup.py
2bndy5: After reading [backward compatibility in PEP 632](https://peps.python.org/pep-0632/#backwards-compatibility), I found that `setuptools` (as of v60.0) has import paths that help mitigate the removal of distutils. According to [the setuptools docs](https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools), we should use 
```ini
# pyproject.toml
[build-system]
requires = ["setuptools>=60", "wheel", "setuptools_scm"]
```
```python
# setup.py
try:
    import setuptools.command.build
except ImportError:  # for setuptools<60.0
    import distutils.command.build
```

jbms: I don't think `setuptools.command.build` is present in setuptools 60.  It isn't listed on that webpage and I tried installing setuptools 60.10.0 and get an import error.

However, as long as we import setuptools first we are in fact using the version of distutils from setuptools, which is why there is no deprecation warning.

2bndy5: Strange, that deprecation warning is completely useless then. All it did was cause confusion here.

sphinx_immaterial/content_tabs.py
2bndy5: `temp_data` persists across all pages, so storing the tab set count in `temp_data` yields an `id` of `__tabbed_4` on the second set of tabs in content_tabs.rst. This is not consistent with upstream HTML generation. It doesn't seem to be a problem, but it is a less precise re-implementation. My main concern here is having to retool it when linked tabs moves to public releases in upstream.

### un-committed mod
Currently, I'm not using the `walkabout()` idea. I basically moved all the algorithm from the transformer into `visit_tab_set()`. Is there an objection to having separate `visit_*()` functions? I'm uneasy about writing all the raw HTML in a single `visit_*()` function.

jbms: temp_data should be cleared after reading each document, but note that it only applies to the initial reading stage, i.e. when the directive is first being processed.  During later stages I don't think it gets reset, so it shouldn't be used for per-document state.  During writing, you can instead use the HTMLTranslator object itself (i.e. the `self` parameter of the visit methods) to store per-document state, because a separate HTMLTranslator object is created per document.

sphinx_immaterial/content_tabs.py
2bndy5: There seems to be a lot of convenient translation (like `<` into `&lt;`) and string normalization (for class names and id attributes) that I don't want to circumvent.

jbms: If you want to use starttag and don't have a suitable node object to pass it, you could always just create a docutils node object on the fly.

sphinx_immaterial/content_tabs.py
2bndy5: I can't treat the translator like a dict; that results in a error. I'm currently doing
```py

def visit_tab_set(self: HTMLTranslator, node: content_tab_set):
    # increment tab set counter
    temp_data = self.builder.env.temp_data
    temp_data["tab_set_count"] = temp_data.get("tab_set_count", 0) + 1
```
Even dynamic attribute creation has the same effect
```python
def visit_tab_set(self: HTMLTranslator, node: content_tab_set):
    # increment tab set counter
    self.tab_set_count = getattr(self, "tab_set_count", 0) + 1
```
The old way i was using was a global 1-item list in content_tabs.py that did get reset for each page.

2bndy5: NM, dynamic attribute creation worked I was just misreading the HTML... commit incoming.

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,207 @@
...
+    for tab_item in node.children:
+        try:
+            tab_label, tab_block = tab_item.children
+        except ValueError as exc:
+            raise ValueError(f"md-tab-item has no children:\n{repr(tab_item)}") from exc
+        tab_total += 1
+        tab_item_identity = tab_set_identity + f"_{tab_total}"
+
+        # create: <input checked="checked" id="id" type="radio">
+        input_node = content_tab_input(
```
jbms, line 170: You could do this tag more simply as:

```
self.body.append(f'<input id="{self.attval(tab_item_identity)}" name="{self.attval(tab_set_identity)}" type="radio"{" checked" if tab_total == 1 else ""}>')
```

I think that more clearly indicates what is happening and avoids the need for the extra content_tab_input node.

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,207 @@
...
+    self.body.append("</label>")
+
+
+def visit_tab_set(self: HTMLTranslator, node: content_tab_set):
+    # increment tab set counter
+    self.tab_set_count = getattr(self, "tab_set_count", 0) + 1
+
+    # configure tab set's div attributes
+    tab_set_identity = "__tabbed_" + str(self.tab_set_count)
+    attributes = {"data-tabs": str(self.tab_set_count) + node["data-tabs"]}
```
jbms, line 152: Potentially rather than have an extra data-tabs attribute you could just compute that here from `len(node.children)`

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,207 @@
...
+    node["set_id"] = tab_set_identity
```
jbms, line 153: What is `set_id` used for?

Normally the HTMLTranslator would not modify the document itself, though I am not sure if that is a problem --- e.g. I'm not clear if multiple translators would ever be run on the same copy of the document.

But in any case I think it would be simpler not to modify the node itself here or below.

Instead you can accomplish the same thing by calling e.g.:

`label_node.walkabout(self)`

That will cause all of the visit/depart methods to be called for the label_node.

sphinx_immaterial/content_tabs.py
2bndy5: Again, this assumes that what the user entered is well formed. I think by doing it the current way, it is clearer, and string normalization (including exceptions raised about invalid characters) from docutils is used.

I understand that there are other methods that can be called to take advantage of docutils conveniences, but this just seems like the simplest and conventional.

sphinx_immaterial/content_tabs.py
2bndy5: `set_id` turns into the element's id attribute. IIRC, in docutils source, it gets concatenated with `ids` attribute. Though I think `set_id` takes precedence when given more the an 1 id (other ids - if they exist- get out into another attribute for the HTML element).

This function is setting up the necessary structure by re-writing the children nodes of a tab-set. There isn't an my need to call `walkabout()` (which would require another for loop at the end followed by a raised `SkipNode` exception). I feel like that just takes control away from docutils.

sphinx_immaterial/content_tabs.py
2bndy5: I've found that there are multiple ways to count the children. I have no objection to this suggestion. I just set the attribute since I had to traverse the child nodes (in `Material TabSetDirective.run()`) to verify they were of `md-tab-item` directives (so I tallied the number of children there).

sphinx_immaterial/content_tabs.py
2bndy5: After giving this more thought, The conveniences of docutils in this instance shouldn't be needed. I applied this suggestion, but I still feel it isn't the most robust.

sphinx_immaterial/content_tabs.py
2bndy5: If I call `label_node.walkabout(self)` here then how do I use `SkipNode` on the child of a child from here?

2bndy5: ok I'm now calling
```py
    tab_label_div.walkabout(self)
    tab_content_div.walkabout(self)
    raise nodes.SkipNode()
```
instead of appending new children to the tab-set node.

This should completely satisfy this request.

sphinx_immaterial/mermaid_diagrams.py
2bndy5: I installed all the `tex-*` stuff used in the RTD docker and test built a pdf from the latex output... The HTML syntax used here (for latex writers) is now replaced with latex syntax to better imitate a literal code block in the generated PDF.

As a side note, I don't think that PDFs are built on RTD because of other numerous unknown references (seems related to `.. _ref_id:` tactic). A PDF can be built from this theme's latex output, but you have to accept that there's no ToC written. We can address this in the future if someone is so inclined, but it isn't the primary goal of this theme. So, for now, I'm just doing the bare minimum to allow a latex builder to complete its job.

docs/mermaid_diagrams.rst
2bndy5: After giving this more thought, IDK if an external extension is practical for everyone's needs (multiple example snippets can rendered identically). I'll try to keep the idea on my radar, but I doubt I'll ever implement something like this.

sphinx_immaterial/content_tabs.py
2bndy5: Its hard to track what the request/comment here is concerning. Recent developments have improved the `visit_tab_set()` to factor out the `content_tab_input` class entirely. I think this should be resolved.

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,182 @@
...
+        try:
+            tab_label, tab_block = tab_item.children
+        except ValueError as exc:
+            raise ValueError(f"md-tab-item has no children:\n{repr(tab_item)}") from exc
+        tab_item_identity = tab_set_identity + f"_{tab_total + 1}"
+
+        # create: <input checked="checked" id="id" type="radio">
+        self.body.append(
+            "<input "
+            + ('checked="checked"' if not tab_total else "")
```
jbms, line 151: It is also valid to just specify 'checked' --- for HTML boolean attributes it is not necessary to specify 'checked="checked"', though that is allowed:
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

In any case, though, I think you need a space afterwards, so that you don't end up with `<input checked="checked"type="radio" ...>`

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,182 @@
...
+    # configure tab set's div attributes
+    tab_set_identity = f"__tabbed_{self.tab_set_count}"
+    attributes = {"data-tabs": f"{self.tab_set_count}:{len(node.children)}"}
+    self.body.append(self.starttag(node, "div", **attributes))
+
+    # walkabout the children
+    tab_total = 0
+    tab_label_div = nodes.container("", is_div=True, classes=["tabbed-labels"])
+    tab_content_div = nodes.container("", is_div=True, classes=["tabbed-content"])
+    for tab_item in node.children:
```
jbms, line 141: Might be slightly clearer to do:

`for i, tab_item in enumerate(node.children)`

rather than manually increment `tab_total`.

docs/mermaid_diagrams.rst
jbms: I'll leave this up to your judgement then.

sphinx_immaterial/content_tabs.py
```diff
@@ -0,0 +1,182 @@
...
+            tab_label, tab_block = tab_item.children
+        except ValueError as exc:
+            raise ValueError(f"md-tab-item has no children:\n{repr(tab_item)}") from exc
+        tab_item_identity = tab_set_identity + f"_{tab_total + 1}"
+
+        # create: <input checked="checked" id="id" type="radio">
+        self.body.append(
+            "<input "
+            + ('checked="checked"' if not tab_total else "")
+            + f'type="radio" id="{tab_item_identity}" name="{tab_set_identity}">'
```
jbms, line 152: Note: You mentioned you were concerned about the robustness --- it is true that properly escaping attributes is an important issue when generating HTML.  Here both of these attributes are generated by this extension themselves and are known not to need escaping.  However, to make it extra robust you could use `self.attval(tab_item_identity)` and `self.attval(tab_set_identity)` --- `attval` is a method implemented by docutils that is used by the `starttag` method to properly escape attribute values.

jbms:
Thanks for making the changes --- it is true there are many ways to implement things and it is not always clear what the best approach is --- I think it is true that your initial approach of using a post transform was perfectly valid.  I just was recently trying to improve the efficiency of documentation generation for tensorstore, where due to the large number of pages generated for the api documentation, it can be rather slow, so I was acutely aware of efficiency concerns.  My thinking is also that ideally the docutils nodes represent semantically meaningful things, and then the HTML translator converts that to HTML.  In this case, the input elements aren't really a semantically meaningful thing on their own, they are just how it happens to be implemented in HTML, so it seems nice to encapsulate that in the HTML translator if possible.

jbms:
Regarding the indicator line issue -- if that doesn't get fixed in upstream mkdocs-material soon that should be easy to fix here.

I also think it would be fairly easy to implement the linked tabs feature if we want it, though I wonder if that should be enabled via a `:linked:` option or similar, rather than a global "feature" option.

sphinx_immaterial/content_tabs.py
2bndy5: oh I forgot to change that.

sphinx_immaterial/content_tabs.py
2bndy5: I can add the space which doesn't seem to be needed as the generated docs perform just the same (probably because the ending quotation mark).

As for the `check="checked"`, I'm just trying to mimic what the HTML looks like on upstream. If you find that the JS doesn't care about this attribute's value, then I can change it to a boolean value.

Notice also that the attribute never changes when selecting different tabs.

sphinx_immaterial/content_tabs.py
2bndy5: thanks for the explanation! Adding it now.

sphinx_immaterial/content_tabs.py
2bndy5: Testing reveals that the quoted value is more resistant to missing spaces, but the boolean value works with the upstream JS just as well. I wish I could see the insider code so I won't have to retool this when that linked tabs feature goes public (~$900 away from that goal).

docs/mermaid_diagrams.rst
2bndy5: I favor leaving it as is because a collapsible admonition will still draw attention (even if collapsed by default). If the implementation detail was much longer or more specific, then it would be a good idea to hide that in a collapsible admonition.

Notice that this paragraph exists before the directive's docs, so a link directly to the directive's docs will scroll past the introductory note(s).

2bndy5: That and I don't want to retrofit all the other statements that talk about the differences between upstream and this port.

sphinx_immaterial/content_tabs.py
2bndy5: now using boolean value

2bndy5:
> I also think it would be fairly easy to implement the linked tabs feature if we want it, though I wonder if that should be enabled via a :linked: option or similar, rather than a global "feature" option.

Whatever upstream does to do it would be preferable. If it ends up being harder to implement here, then I would also favor a `:linked:` option as that also _slightly_ resembles how sphinx-design does it.
- Rather than use the label, the sphinx-design uses a `:key: linked_label_x` that it assign to each `tab-item`. This is of course more long-winded, but using RST is more flexible than MD in adding a specific flag to a block.

jbms:
It wouldn't be hard to support it exactly as in mkdocs-material --- logically though it seems like it may be more useful to allow the user more flexibility.  As you note --- it is easier to support additional options in rst than in markdown.

jbms:
Thanks!

2bndy5:
Either way, we'd have to brush up on rxjs syntax if the `:linked:` approach is added in with upstream's global implementation (in the same TS file). If I'm being honest, I feel like you're better equipped to make that kind of change. 😉

2bndy5:
I still don't fully understand pushing tags directly. I tagged the merge commit https://github.com/jbms/sphinx-immaterial/commit/f925a444a9ca724d46cbacdf909e376ee2c68364 with v0.4.0, but the CI didn't seem to run on that commit.

jbms:
Yeah I can take a look at adding the linked support.

I am bit confused by what is going on with github actions --- we have seen issues before with it not triggering on this repository.  The config is identical to what I have on other repositories where it seems to work reliably, so I'm not sure what is going on here.

jbms:
Actually I'm not seeing your tag.  After you create a tag locally, you have to specifically push the tags to the remote repository:

```
git push --tags origin
```

2bndy5:
Its got to related to PRs coming from other forks. I just switched my remotes to only use this repo from here on out. They recently added [a new event (for security concerns) specific to this situation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).

You can also use `gh` CLI to [manually trigger a workflow](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch).

RTD updated the "stable" build on the merging of this PR, so I think its related to event.

>Actually I'm not seeing your tag. After you create a tag locally, you have to specifically push the tags to the remote repository:

I did, but then I removed it when I realized the CI wasn't triggered for main.

---

I'm going to try adding a `workflow_dispatch` to the CI's list of triggered events. then tag it with v0.4.0 again.

2bndy5:
well I got it to publish, but it keeps triggering twice
1. for commit
2. for the tag

This might be problematic for some scenarios of bumped version numbers. As witnessed when I pushed v0.3.1 tag:
```
Checking dist/sphinx_immaterial-0.3.post1.dev2-py3-none-any.whl: PASSED, with warnings
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
Uploading distributions to https://test.pypi.org/legacy/
Uploading sphinx_immaterial-0.3.post1.dev2-py3-none-any.whl

  0%|          | 0.00/4.54M [00:00<?, ?B/s]
 85%|████████▌ | 3.86M/4.54M [00:00<00:00, 14.8MB/s]
100%|██████████| 4.54M/4.54M [00:01<00:00, 4.19MB/s]
HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
Error during upload. Retry with the --verbose option for more details.
File already exists. See https://test.pypi.org/help/#file-name-reuse for more information.
```

jbms:
Yeah, it will trigger twice, but I think that isn't too big of a deal.  The error you are seeing is for an upload to the *test* PyPI instance.  Only the action run triggered by the tag push will upload to the non-test PyPI instance.

2bndy5:
Unfortunately, such an error will cause the badge on the readme to indicate the failure to upload to test-pypi. I believe we can tell the workflow to [ignore the error for that specific step](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)

PR #57, opened 2022-04-01
Properly configure Google Analytics

### LINKED ISSUE #48: Google Analytics search query support
I am receiving no search queries in Google Analytics from my webpage using `sphinx-immaterial`. ~~Full disclosure, I wasn't receiving them using `sphinx_rtd_theme` either.~~ I was, however, able to see the search terms on the readthedocs.org Admin tab. I don't believe I get any search terms in the Admin tab with `sphinx-immaterial`.

My question still stands, though. It seems that it should be possible to receive Google Analytics search results as is discussed [here](https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-analytics/) with `sphinx-immaterial`.

I believe I'm configuring Google Analytics the correct way. And I can see that the JS is loaded and is emitting page references (which I can see in my GA dashboard).

```python
html_theme_options = {
    "google_analytics": ["G-***********", "auto"],
}
```

However, the webpage is not emitting anything when search terms are typed or selected. I did some investigating and comparison with what the Material for Mkdocs webpage does. I found the Material for Mkdocs website does invoke `analytics.js` when searching and `sphinx-immaterial` does not, see below.

Thoughts on this? Can we add some JS to invoke the `analytics.js` after typing a search query (begin scrolling through results) or when clicking on a search result?

### Material for Mkdocs

On https://squidfunk.github.io/mkdocs-material/, I typed the search term and nothing was emitted on the network. Once I scrolled down onto the first result, or clicked somewhere, the `analytics.js` was triggered (see the last line).

image

Examining the last packet, we see the `?q=a custom search query`.

image

Google Analytics uses the `q` to parse the search result.

image

### Sphinx Immaterial

I noticed no such `?q=` emitted from my website https://galois.readthedocs.io/en/v0.0.25/index.html using `sphinx-immaterial`.

When arriving on a page, there is a `analytics.js` trigger just saying where we are.

image

However, after typing a search query, scrolling down through the results, and then clicking a page there is no additional `analytics.js` trigger with a `?q=` included.

mhostetter:
This PR fixes #48 by properly assigning the `config` dictionary with the Google Analytics property value.

After some digging, I realized the Material for Mkdocs analytics code (partial) was not being added during local testing. The Google Analytics signaling I get on RTD is from custom JS added by RTD at build time.

Below is the partial looking for `config.extra.analytics`, which currently is not set.

https://github.com/jbms/sphinx-immaterial/blob/3b1a1a097d2923c7cd8736eb6509a0aca7d96aba/src/partials/integrations/analytics/google.html#L23-L30

I learned that the `google_analytics` configuration option has been deprecated in favor of `extra/analytics`. See [here](https://squidfunk.github.io/mkdocs-material/upgrade/?h=google+ana#google_analytics).

### Current PR

This PR reads the old configuration style and assigns it properly in the `config` dictionary passed to the Material for Mkdocs pages. I tested this locally and it works -- the partial is added to the page and I can receive GA reports in my account, including search terms! 

```python
html_theme_options = {
    ...
    "google_analytics": ["G-XXXXXXXXXXXX", "auto"],
    ...
}
```

### Future PR

However, I think it would be wise to adjust the `html_theme_options` hierarchy to match Material for Mkdocs configuration hierarchy. That is, I believe the `html_theme_options` should be:

```python
html_theme_options = {
    ...
    "extra": {
        ...
        "analytics": {
            "provider": "google",
            "property": "G-XXXXXXXXXXX",
        }
        ...
    }
    ...
}
```

I noticed `"social"`, `"disqus"`, and `"pwa_manifest"` also don't have the property hierarchy. I believe they should be nested in the `"extra"` dictionary within `html_theme_options`.

https://github.com/jbms/sphinx-immaterial/blob/3b1a1a097d2923c7cd8736eb6509a0aca7d96aba/sphinx_immaterial/__init__.py#L228-L248

I'm willing to make this change and document it in a subsequent PR if you agree.

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -372,9 +372,18 @@ Configuration Options
 
         You can specify any of the available `Google Fonts <https://fonts.google.com/>`_.
 
-    .. confval:: google_analytics_account
+    .. confval:: analytics
 
-        Set to enable google analytics.
+        Set to enable site analytics.
+
+        .. code-block:: python
+
+            html_theme_options = {
+                "analytics": {
+                    "provider": "google",
+                    "property": "G-XXXXXXXXXX"  # Or "UA-XXXXXXXX-X"
+                }
+            }
 
     .. confval:: globaltoc_collapse
 
--- a/sphinx_immaterial/__init__.py
+++ b/sphinx_immaterial/__init__.py
@@ -1,4 +1,4 @@
-"""Sphinx Material theme."""
+"""Sphinx-Immaterial theme."""
 
 import os
 from typing import List, Type, Dict, Mapping
@@ -227,10 +227,13 @@ def html_page_context(
 
     analytics = None
     if theme_options.get("google_analytics"):
+        # Parse old-style analytics config for backwards compatibility
         analytics = {
             "provider": "google",  # Google is the only provider currently supported
             "property": theme_options.get("google_analytics")[0],
         }
+    if theme_options.get("analytics"):
+        analytics = theme_options.get("analytics")
 
     context.update(
         config=dict_merge(
--- a/sphinx_immaterial/__init__.py
+++ b/sphinx_immaterial/__init__.py
@@ -229,7 +229,7 @@ def html_page_context(
     if theme_options.get("google_analytics"):
         analytics = {
             "provider": "google",  # Google is the only provider currently supported
-            "property": theme_options.get("google_analytics")[0]
+            "property": theme_options.get("google_analytics")[0],
         }
 
     context.update(
--- a/sphinx_immaterial/__init__.py
+++ b/sphinx_immaterial/__init__.py
@@ -225,6 +225,13 @@ def html_page_context(
             "versionPath": theme_options.get("version_json"),
         }
 
+    analytics = None
+    if theme_options.get("google_analytics"):
+        analytics = {
+            "provider": "google",  # Google is the only provider currently supported
+            "property": theme_options.get("google_analytics")[0]
+        }
+
     context.update(
         config=dict_merge(
             context.get("config", {}),
@@ -239,9 +246,9 @@ def html_page_context(
                     "social": theme_options.get("social"),
                     "disqus": theme_options.get("disqus"),
                     "manifest": theme_options.get("pwa_manifest"),
+                    "analytics": analytics,
                 },
                 "plugins": theme_options.get("plugins"),
-                "google_analytics": theme_options.get("google_analytics"),
             },
         ),
         base_url=base_url,


2bndy5:
Thanks for looking into this! As for the config hierarchy, we're not going for exact duplication of the mkdocs theme. This is because of differences in implementation details. Generally, with sphinx, extensions get their own config variables (not a sub-section/field of a another theme-specific variable). But, there is going to be confusion about where the theme stops and the patchwork of extensions begins.

mhostetter:
@2bndy5 how should I resolve the Black issues in the CI?

The GitHub Action says

```
Run black --check sphinx_immaterial/*.py
would reformat sphinx_immaterial/__init__.py

Oh no! 💥 💔 💥
1 file would be reformatted, 11 files would be left unchanged.
```

but locally it says this for me

```
matt@DESKTOP-I79NAGP:/mnt/c/Users/matth/repos/sphinx-immaterial$ black --check sphinx_immaterial/*.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/mermaid_diagrams.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/postprocess_html.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/inlinesyntaxhighlight.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/search_adapt.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/object_toc.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/content_tabs.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/__init__.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/apidoc_formatting.py
would reformat /mnt/c/Users/matth/repos/sphinx-immaterial/sphinx_immaterial/nav_adapt.py
All done! 💥 💔 💥
9 files would be reformatted, 3 files would be left unchanged.
```

Is there a config file I need to explicitly pass to `black`?

2bndy5:
I think if you run black from the repo root it should automatically use the settings in the toml file

2bndy5:
I don't see any specific black configuration in this repo. It might just be using a different version of black in the CI runner than what you're using locally.

mhostetter:
Gotcha. I literally just installed `black`, so I'm sure I'm running the latest. And FWIW, I did run it in the repo root.

mhostetter:
The CI says its uses Black 22.3.0 on Python 3.9. I have Python 3.8 and (the latest) Black 19.3b0.

It might be nice if there's a way to streamline this, or make it more consistent. Perhaps a CI action that appends a commit with the black formatting? Just a thought...

If I knew the `black` complaint, I'd manually reformat it...

2bndy5:
We could have the output show a diff. I'm not very fluent on pre-commit (which is how you would go about automating black formatting). Also, it requires the dev to do frequent `git pull`s in the event that there was a pre-commit amendment...

2bndy5:
> The CI says its uses Black 22.3.0 on Python 3.9. I have Python 3.8 and (the latest) Black 19.3b0.

Unless I missed something, you're using an older version of black. Seems it left beta finally. Since they're using a time based symantic versioning, I'd wager 22.3.0 is newer than 19.3b0.

I suppose we could also pin to a specific black version in the requirements.txt file...

mhostetter:
I would have thought the same... However, this is what I get. I'm guessing 19.3b0 is the latest for Python 3.8 and 22.3.0 is the latest for Python 3.9??

```
matt@DESKTOP-I79NAGP:/mnt/c/Users/matth/repos/sphinx-immaterial$ sudo apt install --upgrade black
[sudo] password for matt: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
black is already the newest version (19.3b0-1).
```

I can run the auto-formatter with the Black version I have, but I fear the newer(?) version will complain lol

mhostetter:
Oops. I just realized that the Debain package isn't going to be the latest. I ran `pip install black`, as is done in the CI, and I'm getting the same version. Sorry for the confusion, @2bndy5 !!

2bndy5:
oh yeah, getting updated resources from `apt` typically isn't a thing. 🤣

mhostetter:
I think this PR is ready. Does this look good to you, @jbms?

jbms:
Maybe we should just have an `analytics` theme option that just matches mkdocs-material rather than using a different syntax?

I don't see a particular advantage to doing it differently.

mhostetter:
I added an `"analytics"` options to `html_theme_options`, as suggested. I also left the parser of the old style of configuration in case it's passed, for backwards compatibility.

I also updated the docs. Let me know your thoughts, or if you'd me like to change anything.

jbms:
Thanks!

PR #59, opened 2022-04-07
Speed up TOC generation and other minor fixes

jbms:


2bndy5:
Could you add a summary of the changes to the description? It's bad enough that we're not keeping a change log, and publishing releases via non-annotated tags leaves users with only the git history as a source of logged changes.

Maybe we can start a change log when we tag the theme with a stable major version... I see github added the ability to auto generate a list of changes when publishing a release via the [site](github.com/jbms/sphinx-immaterial/releases)

jbms:
I updated the individual commits to have descriptions.

jbms:
You have a good point about a change log --- some projects use a naming convention for the first line of the commit messages, e.g.:

feat: Description of feature

fix: Description of bug fix

style: Style cleanup

docs: Documentation change

chore: Build/CI change

Potentially we could generate a changelog page in our documentation automatically based on the git history.

Another option would be to use a different syntax within the commit message specifically to indicate the changelog entry.

E.g.

BEGIN_CHANGELOG
Description of change goes here
END_CHANGELOG

And then extract those automatically.

Alternatively we could just manually maintain a changelog file.  One downside of getting the changelog from individual commit messages is that it is tricky if you forget to add one or need to modify it later.  But not sure if that will come up.

2bndy5:
I found a github action a while back that generates a change log from git history (never tried it though).

I like the idea of commit name prefixes. Such an idea deserves mention in the contributing guidelines.

sphinx_immaterial/nav_adapt.py
```diff
@@ -258,6 +272,8 @@ class ObjectIconInfo(NamedTuple):
     ("cpp", "var"): ObjectIconInfo(icon_class="alias", icon_text="V"),
     ("cpp", "type"): ObjectIconInfo(icon_class="alias", icon_text="T"),
     ("cpp", "namespace"): ObjectIconInfo(icon_class="alias", icon_text="N"),
+    ("cpp", "templateParam"): ObjectIconInfo(icon_class="data", icon_text="P"),
```
2bndy5, line 275: I recently wondered if we could wrap the template param names in angle brackets for the secondary toc.
image
Could be less garbled like "\<T A B C N>std::array". Screenshot is from my more-extensions branch's changes in which I added a bunch of template params to see if they could be wrapped into multiple lines in lengthy conditions.

sphinx_immaterial/nav_adapt.py
```diff
@@ -359,20 +375,141 @@ def _collapse_children_not_on_same_page(entry: MkdocsNavEntry) -> MkdocsNavEntry
...
     theme_options = app.config["html_theme_options"]
-    global_toc_node = sphinx.environment.adapters.toctree.TocTree(
-        app.env
-    ).get_toctree_for(
-        pagename,
-        app.builder,
+    global_toc = _get_global_toc(
+        app=app,
+        pagename=pagename,
         collapse=theme_options.get("globaltoc_collapse", False),
```
2bndy5, line 511: I think these are positional args, so there's no need to use kwarg syntax here.

2bndy5 (approved):
Your knowledge of docutils is most impressive. Thank you for letting me be apart of this project!

sphinx_immaterial/nav_adapt.py
jbms: For the tensorstore docs I am using a custom C++ autosummary extension (kind of like the custom Python autosummary extension I'm using), and then I manually set a `toc_title` attribute on all of the signatures, which then gets picked up by `object_toc` so this default name generation doesn't apply.

The reason the template parameter names are included is that the current logic is to concatenate all of the desc_name elements, but Sphinx generates the template parameter names as desc_name as well.  I have a fix for that (which would just exclude the template parameter names).

If we wanted to include them, we should presumably do it in the normal way, e.g. `std::array<T, A, B, C, N>`.  Not sure if that is desirable, though --- may make the TOC entry too long in some cases; it might be better to only do that for template specializations.

2bndy5: I'm ok with excluding them for template declarations. I also like the idea of appending the template names for template specializations.

> For the tensorstore docs I am using a custom C++ autosummary extension 

Was breathe not a viable option due to the babel build system or something? (just curious)

jbms: I did look into breathe --- in fact when I first began creating the TensorStore documentation I selected sphinx with the intention of using breathe.

However, there were a number of things I wanted that breathe didn't really help with:

- generating a custom autosummary similar to my python autosummary
- custom transformations of declarations, to hide internal types, convert enable_if uses to c++20 requires clauses, convert certain patterns to c++20 explicit(bool).

Additionally, doxygen doesn't parse c++ perfectly, and it has its own markup --- I want to just use rST in my doc comments without having to put \rst and \endrst.

Plus doxygen is an additional dependency to have to deal with.

I found that there is a python libclang package, in pypi that is kept up to date and statically links clang, and using that it is pretty easy to extract declarations and comments from c++ source code, and it is guaranteed to parse correctly since the code is tested using clang.

Using that plus the c++ parser included in the sphinx c++ domain I have been able to accomplish what I want.

sphinx_immaterial/nav_adapt.py
jbms: In Python there isn't a clear separation of positional vs keyword args, in that most of the time people don't use `/` or `*` to require one or the other.  Generally I prefer keyword args since it leaves less room for confusion, although granted in this case it doesn't add much.

Issue #56, opened 2022-04-01
Resolve roles in navigation


mhostetter:
I noticed that Sphinx roles are not applied in the navigation tab or content tab headings. Is there anything we can do about that? Is there a way to run a Sphinx role processor on the the titles or captions?

### Content tabs

Here I tried to add an inline literal to a content tab heading. I also tried to add an octicon from the `sphinx-design` extension.
Potentially the octicon did get converted to a SVG but isn't displaying as such?

```rst
    Examples
    --------
    .. md-tab-set::

        .. md-tab-item:: n = 14

            Content one...

        .. md-tab-item:: ``n = 15``

            Content two...

        .. md-tab-item:: :octicon:`graph` Graphs

            Content three...
```

renders as

![image](https://user-images.githubusercontent.com/12112573/161286076-e3122890-400e-4993-9b04-d48216b67f60.png)

<details>
<summary>HTML</summary>

```html
<div class="tabbed-labels docutils">
    <label class="tabbed-label" for="__tabbed_1_1">
        n = 14
    </label>
    <label class="tabbed-label" for="__tabbed_1_2">
        &lt;literal&gt;n = 15&lt;/literal&gt;
    </label>
    <label class="tabbed-label" for="__tabbed_1_3">
        &lt;raw format=”html” xml:space=”preserve”&gt;&lt;svg version=”1.1” width=”1.0em” height=”1.0em”
        class=”sd-octicon sd-octicon-graph” viewBox=”0 0 16 16” aria-hidden=”true”&gt;&lt;path fill-rule=”evenodd”
        d=”M1.5 1.75a.75.75 0 00-1.5 0v12.5c0 .414.336.75.75.75h14.5a.75.75 0 000-1.5H1.5V1.75zm14.28 2.53a.75.75 0
        00-1.06-1.06L10 7.94 7.53 5.47a.75.75 0 00-1.06 0L3.22 8.72a.75.75 0 001.06 1.06L7 7.06l2.47 2.47a.75.75 0
        001.06 0l5.25-5.25z”&gt;&lt;/path&gt;&lt;/svg&gt;&lt;/raw&gt; Graphs
    </label>
</div>
```
</details>

### Navigation tabs

Here I tried to add an octicon to a navigation tab, but it didn't render as expected.

```rst
.. toctree::
   :caption: Getting Started
   :hidden:

   getting-started.rst

.. toctree::
   :caption: :octicon:`info` Basic Usage
   :hidden:

[...]

PR #60, opened 2022-04-11
More extensions

### LINKED ISSUE #47: Support for task lists
Currently sphinx/rst do not provide a way that generates HTML that takes advantage of mkdocs-material's task lists (configured [using a md extension again](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#tasklist)).

<details><summary>Example HTML output from <a href="https://squidfunk.github.io/mkdocs-material/reference/lists/#using-task-lists">mkdocs-material docs</a></summary>

```html
<ul class="task-list">
  <li class="task-list-item">
    <label class="task-list-control">
      <input type="checkbox" disabled="" checked="">
      <span class="task-list-indicator"></span>
    </label>
    Lorem ipsum dolor sit amet, consectetur adipiscing elit
  </li>
  <li class="task-list-item">
    <label class="task-list-control">
      <input type="checkbox" disabled="">
      <span class="task-list-indicator"></span>
    </label>
    Vestibulum convallis sit amet nisi a tincidunt
    <ul class="task-list">
      <li class="task-list-item">
        <label class="task-list-control">
          <input type="checkbox" disabled="" checked="">
          <span class="task-list-indicator"></span>
        </label>
        In hac habitasse platea dictumst
      </li>
      <li class="task-list-item">
        <label class="task-list-control">
          <input type="checkbox" disabled="" checked="">
          <span class="task-list-indicator"></span>
        </label>
        In scelerisque nibh non dolor mollis congue sed et metus
      </li>
      <li class="task-list-item">
        <label class="task-list-control">
          <input type="checkbox" disabled="">
          <span class="task-list-indicator"></span>
        </label>
        Praesent sed risus massa
      </li>
    </ul>
  </li>
  <li class="task-list-item">
    <label class="task-list-control">
      <input type="checkbox" disabled="">
      <span class="task-list-indicator"></span>
    </label>
    Aenean pretium efficitur erat, donec pharetra, ligula non scelerisque
    </li>
</ul>
```
</details>

I'm not sure how popular the feature is, but I figured I'd discuss the implementation here before submitting a proposal.

### About ideal syntax
I would prefer to be able to use the GitHub-flavored MD style of task lists (as used in upstream). This way users migrating from mkdocs to sphinx would be faced with less friction.

```rst
.. md-task-list::
    - [x] Lorem ipsum dolor sit amet, consectetur adipiscing elit
    - [ ] Vestibulum convallis sit amet nisi a tincidunt
        .. md-task-list::
            * [x] In hac habitasse platea dictumst
            * [x] In scelerisque nibh non dolor mollis congue sed et metus
            * [ ] Praesent sed risus massa
    - [ ] Aenean pretium efficitur erat, donec pharetra, ligula non scelerisque
```
There may be nothing stopping the user from implementing this on their own, but the styling from upstream (& thus this theme) is desirable.

### About persistence
As noted in the upstream's docs, checkbox states aren't persistent. Although this could easily be done with a little extra JS from the user's end.

2bndy5:
- added port of `pymdownx.keys` and `pymdownx.tasklist` to the theme.
   - `sphinx_immaterial.keys` is optional since proper usage requires pymdownx installed (for aliasing with `pymdownx.keymap_db` module) which is outlined in the new addition to the docs. closes #47 
   - `sphinx_immaterial.task_lists` is auto enabled since it requires no external dependencies.
- added a theme specific extension called `theme_result` in which a directive (`.. rst-result::`) is fed an RST snippet, and it is output as a literal code-block (with RST syntax highlighting) and the expected rendered output.
   - While this could be useful to other project's docs, the extension relies on some of the CSS provided by mkdocs-material theme. Some CSS was added to conform docutils output with the mkdocs-material's output.
   - Captions for `code-block`s are patched to conform with mkdocs-material's code-blocks' captions (dubbed `filename` in the CSS classes).
   - docs have been modified to make use of the new `rst-result` directive where applicable.
- Revised the CSS solution for wrapping multilined API signatures.
- partial fix for #56 concerning content tabs' label accepting RST roles


.gitattributes
```diff
@@ -0,0 +1,8 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
```
jbms, line 2: I never use Windows, but why do we want any conversion of line endings at all? It seems like it would be be better to just have them always be LF at all times.

sphinx_immaterial/theme_result.py
```diff
@@ -0,0 +1,53 @@
+"""A directive designed to reduce example snippets duplication."""
+from docutils.parsers.rst import directives
+from docutils import nodes
+from sphinx.util.docutils import SphinxDirective
+from sphinx.directives.code import container_wrapper
+
+
+class ResultsDirective(SphinxDirective):
```
jbms, line 8: Maybe call this `rst-example` instead since `result` is rather generic.

sphinx_immaterial/theme_result.py
```diff
@@ -0,0 +1,53 @@
...
+            out_prefix = nodes.container("", classes=["results-prefix"])
+            prefix = self.options.get("output-prefix", "")
+            if not prefix:
+                prefix = "Which renders the following content:"
+            textnodes, _ = self.state.inline_text(prefix, self.lineno)
+            out_prefix += textnodes
+            container_node += out_prefix
+
+        # the `result` CSS class is theme specific (as used by mkdocs-material theme)
+        results_div = nodes.container("", classes=["result"])
```
jbms, line 43: Why are custom CSS rules needed?

sphinx_immaterial/kbd_keys.py
```diff
@@ -0,0 +1,112 @@
+from typing import List, Tuple
```
jbms, line 1: Sphinx already has a `kbd` role:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#other-semantic-markup

sphinx_immaterial/task_lists.py
```diff
@@ -0,0 +1,95 @@
...
+    self.body.append(">")
+    if node["custom"]:
+        self.body.append('<span class="task-list-indicator"></span>')
+
+
+def depart_checkbox_label(self: HTMLTranslator, node: nodes.Node):
+    self.body.append("</label>")
+
+
+class checkboxLabel(nodes.container):
```
jbms, line 29: docutils and sphinx use `snake_case` for node names --- best to be consistent here.

src/assets/stylesheets/main/_api.scss
```diff
@@ -222,3 +240,9 @@ table.longtable.docutils.data.align-default {
     }
   }
 }
+
+// a rule to reset margin for code-blocks with a caption
+.md-typeset .code-block-caption + .notranslate pre,
```
jbms, line 245: Where does this issue occur?

sphinx_immaterial/kbd_keys.py
2bndy5: It does. But the pymdownx ext allows for greater customization. That's why I chose `keys` as the rule name.

src/assets/stylesheets/main/_api.scss
2bndy5: anytime the code block has a caption.

sphinx_immaterial/theme_result.py
2bndy5: done

sphinx_immaterial/theme_result.py
2bndy5: Several reasons. You'll notice that the directive has a `output-prefix` option and takes an optional arg (which gets used as the snippet's caption). This diverges from upstream's usage of example snippets, and builds on the fact that we can do better with this dedicated directive. Not to mention, the docutils output structure is different from mkdocs output.

src/assets/stylesheets/main/_api.scss
```diff
@@ -65,18 +65,36 @@ $objinfo-icon-size: 16px;
         color: var(--md-accent-fg-color);
       }
 
+      // Ensure each parameter starts on a new line and is indented.
       &.sig-wrap {
-        .sig-param {
-          // Ensure each parameter starts on a new line and is indented.
+        .sig-param,
+        .n {
+          + .kt,
```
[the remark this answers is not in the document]
2bndy5: Are you objecting to `.n` or `.kt` or both? I was just going off of what pygments would be looking for.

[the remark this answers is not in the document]
jbms: I just mean that it is kind of incidental whether the parameter happens to start with elements that have those pygments styles --- depending on the specific way the parameter is declared, or depending on which language it is, that might or might not work.  It is also not easy for someone reading it to understand what is going on.

Can you remind me why just using `.sig-param` doesn't work here?

src/assets/stylesheets/main/_api.scss
jbms: Oh okay.  Perhaps we should move this to _typeset.scss instead since it doesn't pertain to API documentation specifically?  (I intended this file for styling sphinx "object descriptions".)

src/assets/stylesheets/main/_api.scss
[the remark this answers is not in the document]
2bndy5: `sig-param` work for the python domain, but it doesn't work for others (like the C family). In fact, parameters are not always encapsulated in a `sig-param` classed div in the python domain (may have something to do with member's scoping).

src/assets/stylesheets/main/_api.scss
2bndy5: I can move it to there. I was more concerned with merging conflict from upstream...

src/assets/stylesheets/main/_sphinx.scss
```diff
@@ -79,4 +79,33 @@
   .rubric {
     font-weight: 700;
   }
+
+  // theme-results directive special rules
```
2bndy5, line 83: Would you like me to move this into _typeset.scss as well?

docs/_static/extra_css.css
```diff
@@ -16,11 +16,51 @@
...
+
+.md-typeset kbd.key-my-special-key:hover,
+.md-typeset kbd.key-git:hover {
+  color: var(--md-accent-fg-color);
+}
+
+/* **************************** custom-task-list style rules *************************** */
+
+/* style for the unchecked checkboxes */
+.custom-task-list-style>ul>.task-list-item>.task-list-control>.task-list-indicator::before {
```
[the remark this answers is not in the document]
2bndy5: >  For example we might add a way to customize the styling from conf.py that doesn't require copying any SVG files, etc.

I'd feel more confident in this idea after we have all the SVG data exported in an optional CSS file. For now, it doesn't get any simpler than using the `:class:` option and adding some extra CSS.

src/assets/stylesheets/main/_api.scss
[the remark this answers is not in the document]
jbms: For Python, if `_parse_arglist` is successful then I believe every parameter will be properly enclosed in a `desc_parameter` docutils node.

However, if the signature cannot be parsed as a real Python signature (e.g. it uses the "[" syntax to indicate optional parts) then Sphinx falls back to `_pseudo_parse_arglist` which does not always enclose arguments in `desc_parameter` nodes.

It looks like the C domain also should enclose each parameter in a `desc_parameter` node, so I'm a bit confused why that isn't working.

https://github.com/sphinx-doc/sphinx/blob/b4276edd848d112b4e981011c334d27cbcb20018/sphinx/domains/c.py#L720

Can you show an example (and e.g. the output of `node.pformat()`) of where that breaks down?  I think it might be better to just fix the domains to produce the proper node structure rather than try to accommodate it in CSS.

sphinx_immaterial/kbd_keys.py
```diff
@@ -0,0 +1,112 @@
...
+        self.body.append(">" + node["text"])
+    else:
+        self.body.append(">")
+
+
+def depart_span(self: HTMLTranslator, node: spanNode):
+    self.body.append("</span>")
+
+
+class kbdNode(nodes.container):
```
jbms, line 41: I think you could just have a single `kbd` node that inherits from nodes.Text and holds the unparsed text.  Then in visit_kbd you can put all of the logic for parsing it and outputting the spans.  You wouldn't need the extra `spanNode` that way.  (The idea is that the docutils nodes would better reflect the semantic structure of the document, and you would leave the HTML-specific stuff to the HTML translation step.)

sphinx_immaterial/kbd_keys.py
```diff
@@ -0,0 +1,112 @@
...
+        keys_out.append(keys_key)
+        if i + 1 != len(keys):
+            keys_sep = spanNode("", classes=[])
+            keys_sep["text"] = KEYS_OPTS["keys_sep"]
+            keys_out.append(keys_sep)
+    keys_div += keys_out
+    return [keys_div], []
+
+
+KEYS_OPTS = {"keys_map": ({} if keys_db is None else keys_db.keymap.copy())}
```
jbms, line 92: From the visit_kbd method you can access `self.builder.config` --- that allows you to avoid needing to use a global variable for `KEYS_OPT`.

sphinx_immaterial/theme_result.py
jbms: Previously we just used existing directives and rST syntax to show examples, e.g.:


```rst
.. details:: Open by default
   :class: example
   :open:

   .. code-block:: rst

      .. details:: Open by default
         :class: example
         :open:
```

This didn't require any custom CSS.  I imagined that this `rst-example` directive would basically just serve as a shorthand to produce exactly the same docutils nodes as before, just without the duplication.

I see that you have added some additional features, like `output-prefix`, but it seems like that could also be reasonably formatted just using existing nodes.

In general it just seems like a good idea to try to be minimal in the use of additional CSS and custom nodes if we can accomplish the same thing just by composing existing directives, as that keeps the theme simpler and there is less to maintain and fewer interactions between things to worry about.

docs/_static/extra_css.css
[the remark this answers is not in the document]
jbms: I think we can avoid having to export all of the SVG data into CSS (as that would waste space by including SVGs that are not actually needed).  Instead as part of the sphinx build we could collect all of the SVGs that are referenced, and output them to a new CSS file.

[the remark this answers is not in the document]
jbms: (Currently all of the CSS for our theme is static, but it doesn't have to be that way.)

.gitattributes
jbms: I guess you have `core.autocrlf` enabled?

I think we can disable any line ending conversion with:

```
* -text
```

sphinx_immaterial/kbd_keys.py
2bndy5: Thank you I had a hell of a time trying to figure out how to do that. I will give that a try.

sphinx_immaterial/kbd_keys.py
```diff
@@ -0,0 +1,112 @@
...
+    KEYS_OPTS["keys_class"] = app.config["keys_class"]
+    KEYS_OPTS["keys_sep"] = app.config["keys_separator"]
+    KEYS_OPTS["keys_strict"] = app.config["keys_strict"]
+    if app.config["keys_map"].keys():
+        KEYS_OPTS["keys_map"].update(**app.config["keys_map"])
+
+
+def setup(app: Sphinx):
+    app.add_config_value("keys_class", "keys", rebuild=True, types=str)
+    app.add_config_value("keys_strict", False, rebuild=True, types=bool)
```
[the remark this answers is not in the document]
2bndy5: This theme does not expect the containing element to be a `kbd` tag. The CSS is styling any kbd element as a single key. So, if there's a kbd element within another kbd element, then it looks weird.

With option off:
<kbd>Ctrl</kbd> + <kbd>v</kbd>

With option on:
<kbd><kbd>Ctrl</kbd>+<kbd>v</kbd></kbd>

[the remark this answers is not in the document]
jbms: Oh I see --- yes, nested `<kbd>` elements doesn't make sense --- is there a reason to even keep that option then?

sphinx_immaterial/kbd_keys.py
2bndy5: There might be an instance where a user doesn't want to take advantage of the `keys_map` config option (which by default includes many aliases for English words/acronyms). That's where the original kbd role could be useful. Of course, if they don't want to use the `keys_map` at all, then they could simply not install pymdownx. But that is a global approach (using `kbd` role vs `keys` role would be context specific).

jbms: It seems kind of unlikely that someone would want to display key bindings in two different ways in one project, though.  If they just want to avoid a particular alias, perhaps there can just be a way for them to modify it from `conf.py`?

sphinx_immaterial/theme_result.py
2bndy5: If you look at the mkdocs-material docs, you'll notice there is a border around the example snippets' output. I wanted to mimic that style (which was apparently inherited). 

However, the fact that literal blocks don't automatically have a `highlight` class (rather it gets a `highlight-xxx` class) caused the upstream CSS to not get used.

.gitattributes
2bndy5: It seems risky without that safety net. It is a major pain in the ass to revert CRLF to LF manually if any get introduced (git diffs don't really show the distinction of different line endings - so I've been bogged down by this in the past).

I read yesterday (on the git docs) that the value `input` would convert CRLF to LF ("but not the other way around"). The docs are terribly worded to the point where the config option's explanation is actually confusing.

sphinx_immaterial/theme_result.py
jbms: Oh I see.  I looked at the mkdocs-material docs source code and I see that there is not in fact any special markdown support for the `results` class -- the mkdocs-material docs just use literal HTML to specify that.

In that case, rather than duplicate the styles, perhaps just modify the existing styles in pymdownx/_highlight.scss so that they work for your extension.

sphinx_immaterial/kbd_keys.py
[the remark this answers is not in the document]
2bndy5: It was part of the `pymdownx.keys` options, so I thought it was easy enough to throw it in there (in case it might be useful to a theme that does not have the same CSS limitation that we have). 

The only option I'm not willing to support (from the pymdownx ext) is the one about experimental support for case sensitive key names (`left-ctrl` vs `LeftCtrl`).

.gitattributes
jbms: It is not too hard to convert line endings if you find a problem --- e.g. you can use the dos2unix program.

I think the way you currently have it, autocrlf behavior will apply to all files by default if the user has `core.autocrlf` enabled, except that `.py`, `.rst`, and `.scss` files will always have `LF` line endings when checked out (even on Windows).

I'm not sure why we want some files to have autocrlf mode and some files always LF.

sphinx_immaterial/theme_result.py
2bndy5: That's reasonable. Again, I was trying to avoid merge conflicts there.

jbms: Yeah makes sense, but if the upstream CSS changes it is actually better to have a merge conflict so that we know about it and make sure our modifications work with it.

sphinx_immaterial/kbd_keys.py
2bndy5: I haven't tried it, but theoretically one could try to override a default alias...
```py
keys_map = {"delete": "Delete"} # default for this is "Del"
```

---
If that works, then I'll happily override the kbd role.

sphinx_immaterial/kbd_keys.py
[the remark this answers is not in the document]
jbms: Oh I see.  Okay.

sphinx_immaterial/kbd_keys.py
2bndy5: I looked into how the original kbd role works, and it looks like the `nodes.container` is used. IIRC,  when `visit_container()` is called, the tag is changed (from div to span or kbd) based on the node's context/flags.

jbms: Yeah, you could still create a custom `kbd` node type but just override the existing `kbd` role.

sphinx_immaterial/theme_result.py
2bndy5: I also considered adding the directive to the RST domain, but it seemed like overkill.

sphinx_immaterial/kbd_keys.py
2bndy5: I remember why I resorted to a global var. The parsing of the interpreted text (`given_text.split(separator)`) is done from the role function. I couldn't find a way to use the config options from the role function Is there an role-function arg that provides access to the config vals?

The parsing should be done at the role function because it creates the necessary nodes. Maybe this can be done from the `visit_*()`, but I haven't tried it.

sphinx_immaterial/kbd_keys.py
2bndy5: Tried overriding the `delete` alias in `keys_map`... It worked, but the theme CSS is still using the CSS classes to add symbols before/after certain keys, 

`:keys:`delete`` results in  image

Using the `:kbd:`Delete`` results in image

sphinx_immaterial/kbd_keys.py
2bndy5: ok its super hard to differentiate segments of a `keys` interpreted text at the `visit_kbd()` context. You should be aware of the HTML output for this role:

```rst
:keys:`ctrl+alt+tab`
```
results in
```html
<span class="keys">
  <kbd class="key-control">Ctrl</kbd>
  <span>+</span>
  <kbd class="key-alt">Alt</kbd>
  <span>+</span>
  <kbd class="key-tab">Tab</kbd>
</span>
```
If I simply pass the given text `ctrl+alt+tab` to the kbd node's text attribute, then I'm having much difficulty getting `visit_kbd()` to differentiate between what should be a
1. \<span>+\</span>
2. \<span class="keys">...\</span>
3. \<kbd class="key-KEY_NAME">displayed text\</kbd>

I'm also trying to appease the request to consolidate the span_node and kbd_node here.

jbms: What I had in mind is that the `kbd` role should just generate a single `kbd` node that contains the plain text without any parsing, i.e. something like: `kbd(text, text)`.  Then all of the parsing happens in `visit_kbd` --- so you would do all of the parsing and generate the full HTML all inside `visit_kbd` called on the single node generated.

Note: You can access the config from the role function via `inliner.document.settings.env.config`.  I figured that out from here: https://github.com/sphinx-doc/sphinx/blob/b4276edd848d112b4e981011c334d27cbcb20018/sphinx/util/docutils.py#L430

Note though that I don't recommend actually using SphinxRole, even though it seems like it would be the correct thing to do, because SphinxRole is badly designed (https://github.com/sphinx-doc/sphinx/pull/10251#issuecomment-1073159787).

jbms: (But anyway if you do the parsing in `visit_kbd` I think you won't need to access the config from the role function.)

2bndy5: My initial approach was to create kbd elements (nodes) from within `visit_kbd()`. I'll try it using raw HTML.

sphinx_immaterial/kbd_keys.py
2bndy5: Just to re-cap this thread, I won't be overriding the original `kbd` role unless there's a way to allow the user to not use the theme's CSS (related to the `pymdownx.keys` ext).

Technically, there are several ways to do this globally (user could set `keys_class = <anything but "keys">` in conf.py), but no way (that I know of) to do this per inline parsing.

2bndy5: I'd also like to point out that listing CLI options in RST ends up generating `<kbd>` elements. See the last row [in this table](https://sphinx-immaterial.readthedocs.io/en/latest/rst-cheatsheet/rst-cheatsheet.html#lists). But, that's different from using the kbd role (syntax must start with a `-` or `/`).

sphinx_immaterial/kbd_keys.py
2bndy5: `kbd_node` is now the only new node introduced by kbd_keys.py. I had to inherit from `nodes.TextElement` because `nodes.Text` was too generic.

As for latex, I just have the key's displayed text inserted as plain text (with separators when applicable).

sphinx_immaterial/kbd_keys.py
2bndy5: This is satisfied now using `self.builder.config`

docs/_static/extra_css.css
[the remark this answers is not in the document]
2bndy5: > Instead as part of the sphinx build we could collect all of the SVGs that are referenced

How do you imagine doing this? I'm willing to hack this together, but I want to share your vision about implementation (how the user would go about specifying which SVGs are used). Initially, I imagine a list (in conf.py - probably part of `html_theme_options`) of filenames (with path relative to .icons folder).

In upstream, squidfunk uses the `pymdownx.emoji` ext to specify inline SVGs (as exemplified in [icons & emojis page](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/#using-icons)). That approach is very similar to what I submitted to sphinx-design (for inline icon roles - still waiting on that release though 😞 ).

src/assets/stylesheets/main/_api.scss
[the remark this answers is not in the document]
2bndy5:
```rst
.. cpp:function:: const MyType Foo(const MyType bar, uint8_t* arr, unsigned int len = DEFAULT_LENGTH, bool baz= false)
```

<details><summary>output from <code>node.pformat()</code></summary>

```xml
<desc_signature classes="sig sig-object cpp highlight" ids="_CPPv43FooK6MyTypeP7uint8_tjb _CPPv33FooK6MyTypeP7uint8_tjb _CPPv23FooK6MyTypeP7uint8_tjb Foo__MyTypeC.uint8_tP.unsigned-i.b" is_multiline="True">
    <desc_signature_line add_permalink="True" xml:space="preserve">
        <inline classes="k">
            const
        <inline classes="w">

        <reference internal="True" refid="_CPPv46MyType" reftitle="MyType">
            <inline classes="n">
                MyType
        <inline classes="w">

        <desc_name classes="sig-name descname" xml:space="preserve">
            <inline classes="n">
                Foo
        <desc_parameterlist xml:space="preserve">
            <desc_parameter noemph="True" xml:space="preserve">
                <inline classes="k">
                    const
                <inline classes="w">

                <reference internal="True" refid="_CPPv46MyType" reftitle="MyType">
                    <inline classes="n">
                        MyType
                <inline classes="w">

                <inline classes="n sig-param">
                    bar
            <desc_parameter noemph="True" xml:space="preserve">
                <inline classes="n">
                    uint8_t
                <inline classes="w">

                <inline classes="p">
                    *
                <inline classes="n sig-param">
                    arr
            <desc_parameter noemph="True" xml:space="preserve">
                <inline classes="kt">
                    unsigned
                <inline classes="w">

                <inline classes="kt">
                    int
                <inline classes="w">

                <inline classes="n sig-param">
                    len
                <inline classes="w">

                <inline classes="p">
                    =
                <inline classes="w">

                <inline classes="n">
                    DEFAULT_LENGTH
            <desc_parameter noemph="True" xml:space="preserve">
                <inline classes="kt">
                    bool
                <inline classes="w">

                <inline classes="n sig-param">
                    baz
                <inline classes="w">

                <inline classes="p">
                    =
                <inline classes="w">

                <inline classes="k">
                    false
```

</details>

<details><summary>HTML output</summary>

```html
<dt class="sig sig-object cpp highlight sig-wrap" id="_CPPv43FooK6MyTypeP7uint8_tjb">
  <span id="_CPPv33FooK6MyTypeP7uint8_tjb"></span>
  <span id="_CPPv23FooK6MyTypeP7uint8_tjb"></span>
  <span id="Foo__MyTypeC.uint8_tP.unsigned-i.b"></span>
  <span class="k"><span class="pre">const</span></span>
  <span class="w"> </span>
  <a class="reference internal" href="#_CPPv46MyType" title="MyType">
    <span class="n">
      <span class="pre">MyType</span>
    </span>
  </a>
  <span class="w"> </span>
  <span class="sig-name descname">
    <span class="n">
      <span class="pre">Foo</span>
    </span>
  </span>
  <span class="sig-paren">(</span>
  <span class="k">
    <span class="pre">const</span>
  </span>
  <span class="w"> </span>
  <a class="reference internal" href="#_CPPv46MyType" title="MyType">
    <span class="n">
      <span class="pre">MyType</span>
    </span>
  </a>
  <span class="w"> </span>
  <span class="n sig-param">
    <span class="pre">bar</span>
  </span>, 
  <span class="n">
    <span class="pre">uint8_t</span>
  </span>
  <span class="w"> </span>
  <span class="p">
    <span class="pre">*</span>
  </span>
  <span class="n sig-param">
    <span class="pre">arr</span>
  </span>, 
  <span class="kt">
    <span class="pre">unsigned</span>
  </span>
  <span class="w"> </span>
  <span class="kt">
    <span class="pre">int</span>
  </span>
  <span class="w"> </span>
  <span class="n sig-param">
    <span class="pre">len</span>
  </span>
  <span class="w"> </span>
  <span class="p">
    <span class="pre">=</span>
  </span>
  <span class="w"> </span>
  <span class="n">
    <span class="pre">DEFAULT_LENGTH</span>
  </span>, 
  <span class="kt">
    <span class="pre">bool</span>
  </span>
  <span class="w"> </span>
  <span class="n sig-param">
    <span class="pre">baz</span>
  </span>
  <span class="w"> </span>
  <span class="p">
    <span class="pre">=</span>
  </span>
  <span class="w"> </span>
  <span class="k">
    <span class="pre">false</span>
  </span>
  <span class="sig-paren">)</span>
  <a class="headerlink" href="#_CPPv43FooK6MyTypeP7uint8_tjb" title="Permalink to this definition">¶</a>
  <br />
</dt>
<dd>
  <p>Some function type thing</p>
</dd>
```

</details>

---

```rst
.. py:function:: func(long: int, param: str, args: None, flags: bool, lists: Union[list, tuple])
```

<details><summary>output from <code>node.pformat()</code></summary>

```xml
<desc_signature class="" classes="sig sig-object py highlight" fullname="func" ids="test_py_module.test.func" module="test_py_module.test">
    <desc_addname classes="sig-prename descclassname" xml:space="preserve">
        test_py_module.test.
    <desc_name classes="sig-name descname" xml:space="preserve">
        func
    <desc_parameterlist xml:space="preserve">
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                long
            <inline classes="p">
                :
            <inline classes="w">

            <inline classes="n">
                <desc_type xml:space="preserve">
                    <reference internal="False" reftitle="(in Python v3.10)" refuri="https://docs.python.org/3/library/functions.html#int">
                        int
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                param
            <inline classes="p">
                :
            <inline classes="w">

            <inline classes="n">
                <desc_type xml:space="preserve">
                    <reference internal="False" reftitle="(in Python v3.10)" refuri="https://docs.python.org/3/library/stdtypes.html#str">
                        str
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                args
            <inline classes="p">
                :
            <inline classes="w">

            <inline classes="n">
                <desc_type xml:space="preserve">
                    <reference internal="False" reftitle="(in Python v3.10)" refuri="https://docs.python.org/3/library/constants.html#None">
                        None
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                flags
            <inline classes="p">
                :
            <inline classes="w">

            <inline classes="n">
                <desc_type xml:space="preserve">
                    <reference internal="False" reftitle="(in Python v3.10)" refuri="https://docs.python.org/3/library/functions.html#bool">
                        bool
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                lists
            <inline classes="p">
                :
            <inline classes="w">

            <inline classes="n">
                <desc_type xml:space="preserve">
                    <reference internal="False" reftitle="(in Python v3.10)" refuri="https://docs.python.org/3/library/typing.html#typing.Union">
                        Union
                    <inline classes="p">
                        [
                    <reference internal="False" reftitle="(in Python v3.10)" refuri="https://docs.python.org/3/library/stdtypes.html#list">
                        list
                    <inline classes="p">
                        ,
                    <inline classes="w">

                    <reference internal="False" reftitle="(in Python v3.10)" refuri="https://docs.python.org/3/library/stdtypes.html#tuple">
                        tuple
                    <inline classes="p">
                        ]
    <reference internal="True" refuri="_modules/test_py_module/test.html#func">
        <inline classes="viewcode-link">
            [source]
```

</details>

<details><summary>HTML output</summary>

```html
<dt class="sig sig-object py highlight sig-wrap" id="test_py_module.test.func">
  <span class="sig-prename descclassname">
    <span class="pre">test_py_module.test.</span>
  </span>
  <span class="sig-name descname">
    <span class="pre">func</span>
  </span>
  <span class="sig-paren">(</span>
  <em class="sig-param">
    <span class="n">
      <span class="pre">long</span>
    </span>
    <span class="p">
      <span class="pre">:</span>
    </span>
    <span class="w"> </span>
    <span class="n">
      <span class="desctype">
        <a class="reference external" href="https://docs.python.org/3/library/functions.html#int"
          title="(in Python v3.10)">
          <span class="pre">int</span>
        </a>
      </span>
    </span>
  </em>,
  <em class="sig-param">
    <span class="n">
      <span class="pre">param</span>
    </span>
    <span class="p">
      <span class="pre">:</span>
    </span>
    <span class="w"> </span>
    <span class="n">
      <span class="desctype">
        <a class="reference external" href="https://docs.python.org/3/library/stdtypes.html#str"
          title="(in Python v3.10)">
          <span class="pre">str</span>
        </a>
      </span>
    </span>
  </em>,
  <em class="sig-param">
    <span class="n">
      <span class="pre">args</span>
    </span>
    <span class="p">
      <span class="pre">:</span>
    </span>
    <span class="w"> </span>
    <span class="n">
      <span class="desctype">
        <a class="reference external" href="https://docs.python.org/3/library/constants.html#None"
          title="(in Python v3.10)">
          <span class="pre">None</span>
        </a>
      </span>
    </span>
  </em>,
  <em class="sig-param">
    <span class="n">
      <span class="pre">flags</span>
    </span>
    <span class="p">
      <span class="pre">:</span>
    </span>
    <span class="w"> </span>
    <span class="n">
      <span class="desctype">
        <a class="reference external" href="https://docs.python.org/3/library/functions.html#bool"
          title="(in Python v3.10)">
          <span class="pre">bool</span>
        </a>
      </span>
    </span>
  </em>,
  <em class="sig-param">
    <span class="n">
      <span class="pre">lists</span>
    </span>
    <span class="p">
      <span class="pre">:</span>
    </span>
    <span class="w"> </span>
    <span class="n">
      <span class="desctype">
        <a class="reference external" href="https://docs.python.org/3/library/typing.html#typing.Union"
          title="(in Python v3.10)">
          <span class="pre">Union</span>
        </a>
        <span class="p">
          <span class="pre">[</span>
        </span>
        <a class="reference external" href="https://docs.python.org/3/library/stdtypes.html#list"
          title="(in Python v3.10)">
          <span class="pre">list</span>
        </a>
        <span class="p">
          <span class="pre">,</span>
        </span>
        <span class="w"> </span>
        <a class="reference external" href="https://docs.python.org/3/library/stdtypes.html#tuple"
          title="(in Python v3.10)">
          <span class="pre">tuple</span>
        </a>
        <span class="p">
          <span class="pre">]</span>
        </span>
      </span>
    </span>
  </em>
  <span class="sig-paren">)</span>
  <a class="reference internal" href="_modules/test_py_module/test.html#func">
    <span class="viewcode-link">
      <span class="pre">[source]</span>
    </span>
  </a>
  <a class="headerlink" href="#test_py_module.test.func" title="Permalink to this definition">¶</a>
</dt>
<dd>
  <p>A function with many parameters.</p>
</dd>
```

</details>

---

```rst
.. py:function:: format_exception(etype, value, tb[, limit=None])
```

<details><summary>output from <code>node.pformat()</code></summary>

```xml
<desc_signature class="" classes="sig sig-object py highlight" fullname="format_exception" ids="format_exception" module="True">
    <desc_name classes="sig-name descname" xml:space="preserve">
        format_exception
    <desc_parameterlist xml:space="preserve">
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                etype
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                value
        <desc_parameter xml:space="preserve">
            <inline classes="n">
                tb
        <desc_optional xml:space="preserve">
            <desc_parameter xml:space="preserve">
                <inline classes="n">
                    limit=None
```

</details>

<details><summary>HTML output</summary>

```html
<dt class="sig sig-object py highlight" id="format_exception">
  <span class="sig-name descname">
    <span class="pre">format_exception</span>
  </span>
  <span class="sig-paren">(</span>
  <em class="sig-param">
    <span class="n">
      <span class="pre">etype</span>
    </span>
  </em>,
  <em class="sig-param">
    <span class="n">
      <span class="pre">value</span>
    </span>
  </em>,
  <em class="sig-param">
    <span class="n">
      <span class="pre">tb</span>
    </span>
  </em>
  <span class="optional">[</span>,
  <em class="sig-param">
    <span class="n">
      <span class="pre">limit=None</span>
    </span>
  </em>
  <span class="optional">]</span>
  <span class="sig-paren">)</span>
  <a class="headerlink" href="#format_exception" title="Permalink to this definition">¶</a>
</dt>
<dd>
  <p>Format the exception with a traceback.</p>
  <dl class="field-list simple">
    <dt class="field-odd">Parameters</dt>
    <dd class="field-odd">
      <dl class="api-field simple">
        <dt><span class="sig-name descname">etype</span></dt>
        <dd>
          <p>exception type</p>
        </dd>
        <dt><span class="sig-name descname">value</span></dt>
        <dd>
          <p>exception value</p>
        </dd>
        <dt><span class="sig-name descname">tb</span></dt>
        <dd>
          <p>traceback object</p>
        </dd>
        <dt>
          <span class="sig-name descname">limit</span>
          <span class="p"> : </span>
          <span class="desctype">
            <span class="desctype">integer or None</span>
          </span>
        </dt>
        <dd>
          <p>maximum number of stack frames to show</p>
        </dd>
      </dl>
    </dd>
    <dt class="field-even">Return type</dt>
    <dd class="field-even">
      <p>list of strings</p>
    </dd>
  </dl>
</dd>
```

</details>

docs/_static/extra_css.css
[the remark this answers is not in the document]
jbms: I imagined that we might have various Python functions or `conf.py` options that control customization of icons --- e.g. specifying custom admonition types with custom icons, or custom task list styles, since the actual CSS required to define those customizations ends up "leaking" a lot of implementation details of the theme, so it would be best if the user didn't need to know about that.

Then there might be a lower level API for registering specific icons that need to be added to the CSS.  The implementation of those higher-level customization options would make use of this lower-level API to ensure the icons they need are available.  The user could also directly make use of this lower level API / `conf.py` option if they have written custom CSS that requires certain icons as well.

src/assets/stylesheets/main/_api.scss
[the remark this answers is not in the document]
jbms: Thanks for gathering all of that --- that is really helpful for understanding what is going on.

So it looks like for the C and C++ domains we can count on every parameter being contained in a desc_parameter node, and for Python we mostly can as well as long as we don't care about the not-real-Python optional syntax.

If `desc_parameter` resulted in `<span class="sig-param">` then we would be all set.  Unfortunately, that is not always the case:

In fact `desc_parameter` only results in a `sig-param` class when using the html5 writer (which our theme always uses) and additionally when the `noemph` attribute is not set:
https://github.com/sphinx-doc/sphinx/blob/b4276edd848d112b4e981011c334d27cbcb20018/sphinx/writers/html5.py#L178

The `noemph` attribute is set by the C and C++ domains on all of the `desc_parameter` nodes that they create, so we never get `sig-param` classes on the `desc_parameter` nodes:
https://github.com/sphinx-doc/sphinx/blob/b4276edd848d112b4e981011c334d27cbcb20018/sphinx/domains/c.py#L720

In contrast the Python domain never sets `noemph=True` so Python parameters are always wrapped in `<em class="sig-param">`.

We can see though that the  C++ domain does still output `class="sig-param"` on the inner name portion of the parameter.  That is because it manually adds that class:
https://github.com/sphinx-doc/sphinx/blob/b4276edd848d112b4e981011c334d27cbcb20018/sphinx/domains/cpp.py#L650

In summary, we clearly can't rely on the `sig-param` class as provided by the existing HTML output for much of anything.

I think we should just add custom `visit_desc_parameter` and `depart_desc_parameter` methods here:
https://github.com/jbms/sphinx-immaterial/blob/4b7f27ae05542cbcca979be4f0dfe905d78f141b/sphinx_immaterial/apidoc_formatting.py#L167

that will wrap every parameter in:

`<span class="sig-param-decl">`

Then we can change our CSS rules to reference this new `sig-param-decl` class rather than `sig-param`.  We can also have `visit_desc_parameter` set `noemph=True` so that we don't get the redundant `<em class="sig-param">` also wrapping Python parameters.

We would no longer have any style rules for `sig-param` so the extra class added by the C++ domain would have no effect.


However, I see that `desc_parameter` does not always result in a `sig-param` class (added by sphinx/writers/html5.py), as that depends on the `noemph` property not being set.  The C++ domain always sets `noemph`, so we never get a sig-param class.

sphinx_immaterial/kbd_keys.py
jbms: Hmm --- yeah I noticed in the HTML spec also that there seemed to be multiple different interpretations of the `<kbd>` element --- sometimes as an individual key, other times just to indicate some form of keyboard input.

I'm not sure what we should do about that.

sphinx_immaterial/theme_result.py
2bndy5: I'm resolving this as I've moved the custom CSS to _highlight.scss. Please review changes there, and let me know if there are any further requests about the custom CSS regarding `.results`, `.results` or `.results-prefix` (directive-specific div containers), etc.

sphinx_immaterial/kbd_keys.py
2bndy5: My initial instinct is do nothing. The immediate question is: Do you still want to override the `kbd` role knowing that would invoke theme-specific CSS (which was solely designed for `pymdownx.keys`)?

If you decide that the CLI args list shouldn't use the same CSS for kbd role, then I guess we could override the CSS on our end - its not a problem upstream because MD doesn't have a native syntax for that scenario like RST does.

docs/_static/extra_css.css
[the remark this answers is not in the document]
2bndy5: As for the custom admonition icons, I can add an option (call it "custom-icon") to the md-admonition directive that lets users specify the relative path to the SVG within the .icons folder. 

Thinking about it further, I suppose the task-list directive could use the same treatment, probably using the directive's `:custom:` option or the `custom_checkboxes` conf.py option (which are both currently boolean values).

Outside of special cases like that I think I'd rather defer to using the sphinx-design ext to use inline icons in text.

I noticed that [upstream relies on page metadata](https://squidfunk.github.io/mkdocs-material/reference/#setting-the-page-icon) to give the page title a prefixed icon. It also recommends altering the [color of icons with a user-specified CSS](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/#with-colors) class (& corresponding CSS rules), so eventually the user may have to understand fundamental CSS at some point.

---
BTW, I did simplify this CSS selector by using the elements' tag names instead of using primarily ext-specific CSS classes:
```css
.custom-task-list-style > ul > li > label .task-list-indicator::before {
  /* ... */
}
```

src/assets/stylesheets/main/_api.scss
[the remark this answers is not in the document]
2bndy5: Great analysis! I've locally tested the `visit/depart_desc_parameter()` idea with _api.scss adjusted accordingly, and it worked like a charm. Committed solution for review.

> We can also have visit_desc_parameter set noemph=True so that we don't get the redundant `<em class="sig-param">` also wrapping Python parameters.

I opted not to do this since I wasn't sure how to determine node's domain. Since we aren't using the `sig-param` class in CSS anymore, it seemed redundant to modify the original behavior.

### Side-effects
- All parameters (albeit multiple or single line signature) will be wrapped in a `span.sig-param-decl` element. So users could extend the multi-line approach to single line signatures if they so choose.
- The non-pythonic optional syntax is not wrapped with the `*_desc_parameter()`
   ```html
   <span class="sig-name descname"><!--  --></span>
   <span class="sig-paren">(</span>
   <span class="sig-param-decl"><!-- -->, </span>
   <span class="optional">[</span>
   <span class="sig-param-decl"><!-- -->, </span>
   <span class="optional">]</span>
   <span class="sig-paren">)</span>
   ```
   This ends up looking a little weird for multi-lined signatures, but not at all alien:
   image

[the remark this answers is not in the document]
jbms: yeah the "optional" syntax doesn't look great but I think we can just encourage users not to use it

sphinx_immaterial/kbd_keys.py
jbms: Yes I agree --- let's just do nothing for now.

docs/_static/extra_css.css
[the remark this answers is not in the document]
jbms: I don't think it makes sense to worry about it as part of this pull request, but yeah, adding an option to `md-admonition` and other directives seems like a good idea.

As I see it, in most cases users wouldn't directly specify such an option within their rST sources.  Instead, they would define their own custom admonition directives by inheriting from the class that defines `md-admonition`, and would specify the icon there.

E.g. they could define an `example` admonition or something and specify a custom icon to use for it.  That way they can maintain a separation between the semantic content of their documents and any style information.

I think it is not so bad for users to have to specify CSS as long as they can specify their styles in a simple way that doesn't depend on implementation details of the theme.  E.g. in the example you posted (https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/#with-colors) they are just defining a CSS class and specifying a `color` --- nothing complicated about that.

sphinx_immaterial/apidoc_formatting.py
```diff
@@ -134,6 +134,7 @@ def visit_desc(self, node: sphinx.addnodes.desc) -> None:
 
         # Augment the list of classes with `objdesc` to make it easier to
         # style these without resorting to hacks.
+        # add highlight to invoke syntax highlighting in CSS
```
jbms, line 137: this comment belongs in the `visit_desc_signature` method below instead.

src/assets/stylesheets/main/extensions/pymdownx/_highlight.scss
```diff
@@ -322,22 +322,43 @@
...
-
-    // Clearfix, because we can't use overflow: auto
-    &::after {
-      display: block;
-      clear: both;
-      content: "";
+  // sphinx-immaterial: altered to support
+  // 1. docutils classes (.literal-block-wrapper, div[class^="highlight-"])
+  // 2. rst-result directive's output-prefix (.results-prefix)
+  :is(.highlight, .highlighttable, .literal-block-wrapper),
```
jbms, line 328: I think you can put all of the conditions inside `:is`, e.g. `:is(.highlight, .highlighttable, div[class^="highlight-"], .results-prefix) + .result`

That way the indentation doesn't change so merging will be easier

sphinx_immaterial/apidoc_formatting.py
2bndy5: This placement is actually a artifact from local dev testing. Good catch!

src/assets/stylesheets/main/extensions/pymdownx/_highlight.scss
2bndy5: I think this request violates the stylelint line length though

2bndy5: Nope. It seems to have passed stylelint checks 👍🏼

docs/_static/extra_css.css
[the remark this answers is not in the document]
2bndy5: ok, resolving this thread since the selectors have been simplified. I'm glad to have started discussing how to implement user specified icons though.

.gitattributes
2bndy5: well, I did some experimenting.
- Setting my global config `core.autocrlf=input` provides LF endings in the local workspace on checkout. But this overrides the `eol` config option.
- When `eol` is set to `lf`, then the workspace uses LF endings (which is the desired result - even on Windows), but this means `autocrlf` should be set to `false` (which means no normalizaton of line endings on checkout).
- If `core.autocrlf=false` and `core.eol=lf` are set via .gitattributes, they seem to have no affect when the git global config is not `core.autocrlf=false`.

So, I think I found a compromise to override a user's global config for this repo:
1. `* text autocrlf=false` prevents line ending normalization
2. `* text eol=lf` ensures workspace uses LF on checkout

This way if a CRLF is introduced, then `npm run check` will likely fail. At that point the user should be aware of the line endings' preference for this repo and convert them manually as needed.

### git docs links
This problem seems to have been addressed so many times (in various ways) that the docs are very nuanced and damn confusing.
- [core.eol](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreeol)
- [core.autocrlf](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreautocrlf)
- [gitattributes specs](https://git-scm.com/docs/gitattributes)

src/assets/stylesheets/main/_api.scss
```diff
@@ -45,7 +45,7 @@ $objinfo-icon-size: 16px;
         font-weight: 700;
       }
 
-      a.reference > .sig-name {
+      a.reference > span {
```
jbms, line 48: I guess this is okay but can you explain more why this is needed?

jbms:
Note: Going forward, it would probably simplify things if you try to keep unrelated changes in separate pull requests.  Sometimes that isn't easy to do, but e.g. this PR could have been split into a number of separate ones:

- rst-example directive
- cpp domain fixes
- task_list support
- keys support
- content tabs bug fix

src/assets/stylesheets/main/_api.scss
2bndy5: Its related to the differences in domain. I found that reference links in C++ domain did not get special coloring (the intention of this CSS rule) when the `highlight` class (which is needed for other syntax highlighting like `class`) is added to the signature.

2bndy5:
Busted! I'm lazy about branch management in this regard. I have been trying to be better about this, but clearly it is a bad habit.

jbms:
I will often create multiple changes in the same working tree, and then split them into a sequence of commits.  I often do this by staging individual files or individual hunks/line ranges.  This works well as long as the changes don't affect nearby lines of the same file.

I use the Emacs magit package to stage individual hunks, but if using vscode it looks like you can do that as described here:

https://stackoverflow.com/questions/34730585/how-can-i-commit-some-changes-to-a-file-but-not-others-in-vscode

If you already have a sequence of commits, and then have made some changes in your working tree that you want to incorporate into that commit sequence, you can do that by staging the changes applicable to one of the commits, creating a fixup commit (e.g. `git commit --fixup=commithash`), not sure if there is a convenient way to do that with vscode, you can also just create a new commit, then use interactive rebase to apply it as a "fixup" to the existing commit.

jbms:
This all looks good to me, thanks!  I just had one comment about the `.sig-name` CSS change.

Issue #58, opened 2022-04-07
Usability issues with long TOCs


jbms:
For API documentation you can easily end up with a very long table of contents in the left and/or right side panels.

This results in two usability issues:

- The left TOC does not start out scrolled to have the current page in view, and the right TOC does not automatically keep the current section in view.  That makes it challenging to understand the context of the current page/section.  This feature is apparently implemented in the insiders-only version of mkdocs-material (https://github.com/squidfunk/mkdocs-material/issues/2155).
- When scrolling the left or right TOC, if there are multiple levels of nesting, the ancestors of the current page/section may be scrolled out of view and not visible, which also makes it difficult to understand the context of the current page/section.  This could be solved by making the section headers "sticky", so that the headers for each nesting level "collect" at the top of the TOC.  An implementation of this is explained here: https://stackoverflow.com/a/55941740

PR #64, opened 2022-04-21
Cpp domain improvements

jbms:
This adds a number of improvements for the C++ domain:

- Improved styling
- clang-format-based signature formatting
- Support for resolving cpp domain references to external links, with specific support for cppreference.com
- Support for `#include` directives in C++ signatures to indicate the required header
- Support for cross-linking parameters in C and C++ domains
- Support for "synopses" of C/C++ domain objects shown in tooltips and search results
- Option for including object description fields in TOC


docs/format_signatures.rst
```diff
@@ -0,0 +1,122 @@
...
+   The default value is :python:`70`.
+
+.. rst-example::
+
+   .. py:function:: long_function_signature_example(\
+                      name: int, \
+                      other_param: list[str], \
+                      yet_another: bool = False, \
+                      finally_another: str = "some long string goes here"\
+                    ) -> Tuple[A, Very, Long, Return, Type]
```
2bndy5, line 48: I'm the RTD CI, you'll see that this return type is styled like a hyperlink, but it doesn't actually behave like a hyperlink (only the color changes when clicking it - nothing else happens).

I would expect that `Tuple` and `Type` (a built-in python function) lead to the python docs, but it doesn't.

docs/format_signatures.rst
```diff
@@ -0,0 +1,122 @@
...
+    extensions = [
+        # other extensions...
+        "sphinx_immaterial.format_signatures",
+    ]
+
+    clang_format_signatures_domain_styles = {
+        "cpp": """{
+        BasedOnStyle: LLVM,
+        ColumnLimit: 70,
+        }""",
```
2bndy5, line 72: Are the curly brackets necessary here?

I understand this is a documented example, but my question is more specific to the extension.

docs/format_signatures.rst
```diff
@@ -0,0 +1,122 @@
...
+          "cpp": """{
+          BasedOnStyle: Google,
+          BreakInheritanceList: BeforeColon,
+          ColumnLimit: 70,
+          }""",
+      }
+
+   .. tip::
+
+      It is recommended to include ``ColumnLimit: 70`` as a style option.
```
2bndy5, line 93: Should probably mention that this is the default value for the CSS styling as well. That way there's an explicit distinction between the 2 implemented column limits.

docs/format_signatures.rst
```diff
@@ -0,0 +1,122 @@
...
+
+   .. warning::
+
+      Due to how this extension is implemented, style options that change
+      non-whitespace characters, such as setting ``QualifierAlignment`` to a
+      value other than ``Leave``, must not be used.
```
2bndy5, line 99: What is the consequence? Shouldn't the user be made aware of the consequence?

docs/index.rst
```diff
@@ -73,6 +73,9 @@ or ``theme.conf`` for more details.
     mermaid_diagrams
     task_lists
     keys
+    format_signatures
+    cpp
+    external_cpp_references
```
2bndy5, line 78: I think the domain customizations should be part of a separate `toctree` in the docs (may include the formatting extension doc also).

docs/conf.py
```diff
@@ -156,6 +158,44 @@
...
+clang_format_signatures_domain_styles = {
+    "cpp": """{
+    BasedOnStyle: LLVM,
+    ColumnLimit: 70,
+    }""",
+}
+
+html_wrap_signatures_with_css = ["py"]
+
+external_cpp_references = {
```
2bndy5, line 170: If you prepend comments here that describe which extension this option is related to, then it would be easier to use a `literalinclude` block in the docs.

As an example, I did this for the `kbd_keys` extension's `keys_map` config option.

docs/conf.py
```diff
@@ -156,6 +158,44 @@
...
+        "desc": "C++ type alias",
+    },
+    "nlohmann::basic_json": {
+        "url": "https://json.nlohmann.me/api/basic_json/",
+        "object_type": "class",
+        "desc": "C++ class",
+    },
+}
+
+cpp_strip_namespaces_from_signatures = ["my_ns1", "my_ns2"]
```
2bndy5, line 183: Same idea here. Specify the extension that this is related to and optionally use a `literalinclude` block in the docs.

package.json
```diff
@@ -31,6 +31,7 @@
     "build:noclean": "ts-node -T tools/build --verbose",
     "docs": "npm run build && rimraf docs/_build && cd docs && sphinx-build . _build -a",
     "docs:dev": "npm run build:dev && rimraf docs/_build/* && cd docs && sphinx-build . _build -a",
+    "docs:watch": "ts-node -T tools/build --verbose --dirty --watch --docs",
```
2bndy5, line 34: ? Is this intended to mimic `mkdocs serve`?

I don't really use these commands, but I'm trying to learn to.

2bndy5 (changes requested):
### First glance
I'll dive into the src changes later. This is just what I noticed in the docs changes.

docs/cpp.rst
```diff
@@ -0,0 +1,22 @@
...
+
+      cpp_strip_namespaces_from_signatures = ["my_ns1", "my_ns2"]
+
+   .. rst-example:: C++ function definition with stripped namespaces
+
+      .. cpp:type:: my_ns1::A
+
+      .. cpp:type:: my_ns2::B
+
+      .. cpp:function:: void my_ns1::MyFunction(my_ns1::A x, my_ns2::B y, my_ns3::C);
```
2bndy5, line 22: Given the `cpp_strip_namespaces_from_signatures` config value's setting, we could also document a `.. cpp:type:: my_ns3::C` and the namespace won't get stripped in this function's 3<sup>rd</sup> parameter, correct?

sphinx_immaterial/apidoc_formatting.py
```diff
@@ -225,92 +111,80 @@ def depart_caption(self, node: docutils.nodes.Element) -> None:
...
-    GoogleDocstring = sphinx.ext.napoleon.docstring.GoogleDocstring
 
-    def _add_admonition_class(method_name: str, class_name: str) -> None:
-        orig_method = getattr(GoogleDocstring, method_name)
+def _normalize_wrap_signatures_predicate(value):
+    if value is None:
+        value = True
+    if isinstance(value, bool):
+        return lambda _domain: value
+    if callable(value):
```
2bndy5, line 158: What's the use case for accepting a callable. Is it meant to be a dynamic filter?

sphinx_immaterial/cpp_domain_fixes.py
```diff
@@ -0,0 +1,444 @@
...
+            continue
+        index = grandparent.index(parent)
+        if index + 1 >= len(grandparent.children):
+            continue
+        sibling = grandparent[index + 1]
+        if (
+            not isinstance(sibling, sphinx.addnodes.desc_sig_punctuation)
+            or sibling.astext() != "::"
+        ):
+            continue
```
2bndy5, line 373: Does this account for nested namespaces as well? It's not important to me, but it might be to very "picky" users.

src/assets/stylesheets/main/_api.scss
```diff
@@ -3,93 +3,96 @@ $api-color-keyword: var(--md-code-hl-keyword-color);
...
+      }
 
-        .sig-paren:last-of-type {
-          // Ensure the final paren starts on a new line
-          &::before {
-            white-space: pre;
-            content: "\A";
-          }
-        }
+      &:hover,
```
2bndy5, line 84: I think this selector is what is causing a non-hyperlinked return type look like hyperlink when hovered over (or tapped on in mobile browsers) as reported by [my other comment](https://github.com/jbms/sphinx-immaterial/pull/64/files#r855674950))

sphinx_immaterial/format_signatures.py
```diff
@@ -0,0 +1,301 @@
...
+        source = io.StringIO()
+
+        for sig_id, signature in signatures.items():
+            source.write(f"// {sig_id}\n")
+            source.write(signature.strip().strip(";"))
+            source.write(";\n")
+
+        style = domain_styles[domain]
+        result = subprocess.run(
+            [app.config.clang_format_command, f"-style={style}"],
```
2bndy5, line 261: Instead of using an f-str, I think you could concatenate the style arg with the `{`/`}` (expected by clang-format's CLI) surrounding the arg's value. This way the user doesn't need to add the curly brackets to a style definition (which follows yaml syntax) in the conf.py option.

There may also be a way to escape curly brackets in an f-str, but I don't know it off-hand.

docs/format_signatures.rst
jbms: This is just exactly following the syntax required by the clang-format `-style` command-line option.

Allowed values are:

- `file`
- `LLVM`
- `GNU`
- `Google`
- `Chromium`
- `Microsoft`
- `Mozilla`
- `WebKit`
- `{key:value, ...}`

Requiring the user to include the curly braces in the style means they could theoretically use one of the other plain-string values, and makes it consistent with the clang-format command-line syntax.

However, it is true that given the recommendation to use a column limit of 68 (previously 70), none of these options except the last one is likely to be used, so we could just add the curly braces.

2bndy5: Technically speaking the curly brackets are an implementation detail that I don't think should be burdened on the user. It might be better to encourage `LLVM`, `GNU`, `Google`, `Chromium`, `Microsoft`, `Mozilla`, `WebKit`, or `file` (or `file:path/relative/to/docs/src/.clang-format` for v14+). That way a complex style definition is properly abstracted from being in a dict's key's value. If users are going to define their own style definition, then it will not likely be as simple as a few lines (based on my experience and recommendations from others).

Coincidentally, I also coded this `subprocess.run()` call identically [in a github action I re-wrote in python](https://github.com/shenxianpeng/cpp-linter-action/blob/9f2caa8fa2e3e7a5bf39103e634a7d754c36c056/cpp_linter/run.py#L455). However, that action explicitly supports the predefined options (which can be clang version dependent) & implicitly supports the curly bracket option if the user is clever enough (though its not at all encouraged).

docs/external_cpp_references.rst
```diff
@@ -0,0 +1,116 @@
+External C++ symbol links
```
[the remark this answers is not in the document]
jbms: Done.

docs/format_signatures.rst
jbms: Done.

docs/format_signatures.rst
jbms: Added comment.

package.json
jbms: This doesn't actually start an http server for hosting the docs --- you still have to do that separately if you need it.  But it does auto-rebuild the docs if you make any changes to any of the sources.  That way you just have to refresh the web browser to see the changes.

sphinx_immaterial/apidoc_formatting.py
jbms: Removed.  I originally documented that supported a dynamic filter, but then removed that to simplify things.

sphinx_immaterial/cpp_domain_fixes.py
jbms: No --- would be nice, but not done at the moment.  I added a comment to the documentation.

docs/format_signatures.rst
jbms: I'm not sure I understand what you are proposing --- if we always surround the user's style with curly braces, then the user can't specify one of those other options, like LLVM, file, etc.

sphinx_immaterial/cpp_domain_fixes.py
2bndy5: would it be possible (in the meantime) to specify a nested namespace like so:
```py
cpp_strip_namespaces_from_signatures =["abc", "abc::my_ns1"]
```

BTW, I don't see the newly added example stripping the namespace like the admonition describes. It still renders as
image
Does the type need to be documented for the stripping to happen?

docs/cpp.rst
```diff
@@ -0,0 +1,62 @@
...
+----------------------------------------
+
+This theme extend the C and C++ domains to allow signatures to specify required
+:cpp:`#include` directives.
+
+.. rst-example:: Specifying :cpp:`#include` directives in signatures
+
+   .. cpp:function:: #include "my_header.h"
+                     #include "another_header.h"
+                     void foo(int param);
```
2bndy5, line 60: The user shouldn't be surprised if clang-format re-orders `#include` statements when using that ext.

docs/conf.py
jbms: Done.

docs/conf.py
jbms: Done.

docs/format_signatures.rst
jbms: I fixed the hover effect per your suggestion, and also fixed the issue with the close paren not being on a separate line --- `:last-of-type` matches based on the tag name, not the class.

I think fixing the cross-linking of Tuple will require pulling in some other changes I have in the tensorstore docs.

docs/format_signatures.rst
2bndy5: I'm ok with leaving it as is because adding curly brackets would be only apply if a predefined style isn't used. So if you wanted to change it, then there may need to be a 
```py
if style not in (
    "LLVM", "GNU", "Google", "Chromium", "Microsoft", "Mozilla", "WebKit"
) or not style.startswith("file"):
    style = "{" + style + "}"
```
But, that would less sustainable because clang-format may choose to add or remove a predefined style at any time in the future.

By encouraging the use of predefined styles (or `file`), then we might be able to avoid testing for predefined styles and still implicitly support inline curly bracket styles.

docs/cpp.rst
2bndy5: Oops, I may be actually thinking of clang-tidy concerning the ordering, but clang-format also does some grouping and indenting with `#include`s.

sphinx_immaterial/cpp_domain_fixes.py
jbms: No it doesn't, since it happens before reference resolution anyway.

My comment was actually wrong.  In fact you are right that nested namespaces are supported --- I revised the example.

docs/cpp.rst
jbms: The format_signatures extension detects `#include` statements and ignores them in part because they are not normal signatures and also because as you mention reordering would be a problem (due to how it works, it cannot support reordering anyway).

2bndy5: It would be cool if std libs like `cstdio` could be cross-referenced from include statements...

docs/format_signatures.rst
jbms: The thing is a predefined style will not have the right column limit --- and similarly the project's own .clang-format also won't have the right column limit.

docs/cpp.rst
jbms: Yeah, perhaps in a later commit --- would require some thought to decide how to cross link header files in a general way.

docs/format_signatures.rst
2bndy5: Actually, some styles do define a `ColumnLimit`. It sounds like the preferred method is to use custom styles. If curly brackets are not going to be automatic, then a note about that might help (since the YAML syntax doesn't really use curly brackets the same as JSON).

sphinx_immaterial/format_signatures.py
2bndy5: resolving as this discussion is happening on another review comment.

docs/format_signatures.rst
jbms: Yeah I think they all specify a column limit, but they don't use a column limit of 68, which is what works best for this theme.  Probably they mostly use 80.

I added a note about curly braces being required.

Note that the style syntax is not yaml, but rather a custom syntax used by clang-format.

sphinx_immaterial/format_signatures.py
```diff
@@ -0,0 +1,297 @@
...
```
2bndy5, line 257: Wouldn't it make sense to override clang-format's syntax guessing based on filename?

From `clang-format -h`:
```
  --assume-filename=<string> - Override filename used to determine the language.
                               When reading from stdin, clang-format assumes this
                               filename to determine the language.
```
This option is consistently available for clang-format v9+ (possibly earlier, but I think v9 is about to become deprecated).

docs/format_signatures.rst
2bndy5: From clang-format docs:
> The `.clang-format` file uses YAML format

But maybe there's a difference in .clang-format file vs CLI value.

jbms: Oh, I think you are right -- sorry.

docs/conf.py
```diff
@@ -156,6 +158,44 @@
...
+   :language: python
+   :class: highlight
+
+.. role:: cpp(code)
+   :language: cpp
+   :class: highlight
+
+.. role:: json(code)
+   :language: json
+   :class: highlight
```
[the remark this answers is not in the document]
2bndy5: Actually, I think this is also used in specimen.rst for inline JS code.

docs/format_signatures.rst
```diff
@@ -0,0 +1,127 @@
...
+
+   When specifying an inline style (as opposed to a predefined style), it is
+   necessary to enclose the style in curly braces, as in the example above.
+   Since the predefined styles (such as ``Google``, ``LLVM``, etc.) do not use a
+   column limit of 68, it is not recommended to use a predefined style.
+
+   .. tip::
+
+      It is recommended to include ``ColumnLimit: 68`` as a style option, as
+      this is the default for the CSS-based wrapping, and is the number of
```
2bndy5, line 94:
```suggestion
      this is the default for the CSS-based wrapping, and it is the number of
```
Conjunctions that follow a comma should involve a subject. This rule of English is a bit hidden here because of the prepositional anecdote (which are also delimited by commas).

docs/format_signatures.rst
```diff
@@ -0,0 +1,127 @@
...
+
+   .. cpp:function:: template <typename T, \
+                               typename U = void, \
+                               int AnotherParameter = 42> \
+                     requires std::is_const_v<T> \
+                     const MyType LongFunctionSignatureExample(\
+                       const MyType bar, \
+                       uint8_t* arr, \
+                       unsigned int len = DEFAULT_LENGTH, \
+                       bool baz = false);
```
2bndy5, line 125: I'm guessing the ending semi-colon is required. If it isn't included, I wonder if clang-format would add it (and possibly crash the docs build since its a non-whitespace change).

jbms: The extension always appends a semicolon to the input provided to clang-format.  However when the docutils nodes are later transformed, any unmatched trailing text in the clang-format output is ignored, so it doesn't cause a problem.

docs/cppreference.rst
```diff
@@ -0,0 +1,49 @@
...
+
+   .. cpp:function:: template <typename T> \
+                     std::enable_if_t<std::is_same_v<T, char> || \
+                                      std::is_same_v<T, uint8_t>, \
+                                      std::string> \
+                     ConvertVectorToString(std::vector<T> x);
+
+      Converts an :cpp:expr:`std::vector` to an :cpp:expr:`std::string`.
+
+.. confval:: cppreference_xml_files
```
2bndy5, line 27: This option seems like it would be required if using the ext in another theme that doesn't ship with the XML files used as a default here. Does the extension somehow use this theme's xml files (even for other themes)?

docs/conf.py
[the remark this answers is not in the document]
2bndy5: Resolving this for now. It should be clear to anyone that inspects the conf.py file for our docs, but I still think a mention is appropriate for those migrating from mkdocs.

docs/cppreference.rst
jbms: The files are just data files of the package, and they are just accessed using the normal `importlib.resources` machinery, so they will work fine even if not using this theme.  They are only used during the documentation generation process, but they don't get included in the output as static files.

docs/format_signatures.rst
2bndy5: On second thought, its probably better to break this up into 2 sentences. Sorry, I'm being "that guy", but grammar mistakes always
1. throw me when I encounter them in other people's docs (flask docs simply don't use commas unless it is followed by a conjunction 😠 )
2. may cause difficulty when input to google translate (or similar)

It looks like "as" is used here _as_ a preposition and a conjunction (which is followed by another conjunction "and" => run-on sentence).

jbms: Fixed.

jbms:
Thanks for your review @2bndy5  --- I believe I have now addressed all of your comments.

I also added some new functionality --- synopses of C/C++ domain objects displayed as tooltips and in search results.

2bndy5:
Wow! The description fields feature is something I've dreamed of for years. And, you just dished it out like a side entree 🤣 . I'm reviewing those changes now - everything else looks good.

sphinx_immaterial/cpp_domain_fixes.py
```diff
@@ -0,0 +1,975 @@
...
+
+
+def _monkey_patch_cpp_add_precise_template_parameter_object_types():
+    """Adds more precise template{Type,NonType,Template}Param object types."""
+    sphinx.domains.cpp.CPPDomain.object_types[
+        "templateTypeParam"
+    ] = sphinx.domains.ObjType(
+        "type template parameter",
+        "identifier",
+        "identifier",
```
2bndy5, line 472: Is this duplicated on purpose?

2bndy5:
I've noticed that the tooltips for `confval` references no longer show up.

2bndy5:
Would it be possible to move the hidden permalink symbol to outside the `code` element?

It looks a bit weird
image image

sphinx_immaterial/cpp_domain_fixes.py
jbms: Fixed.

jbms:
> Would it be possible to move the hidden permalink symbol to outside the `code` element?
> 
> It looks a bit weird !

Fixed.

jbms:
> I've noticed that the tooltips for `confval` references no longer show up.

jbms:
> I've noticed that the tooltips for `confval` references no longer show up.

I wasn't aware that they had tooltips --- can you explain a bit more what you mean?

jbms:
> Wow! The description fields feature is something I've dreamed of for years. And, you just dished it out like a side entree rofl . I'm reviewing those changes now - everything else looks good.

To be clear, `include_object_description_fields_in_toc` just controls whether "Parameters", "Returns", etc. end up in the TOC.

Whether the individual parameters themselves end up in the TOC is independent of that currently, and requires domain-specific support --- this PR adds support for that to the C++ domain (in the cross-link parameters commit).  And in the tensorstore docs I also do that for Python and JSON, but those changes aren't part of this theme yet.

However, it looks a bit weird to have the individual parameters in the TOC without a "Parameters" heading.

2bndy5:
> However, it looks a bit weird to have the individual parameters in the TOC without a "Parameters" heading.

I am seeing a section for Parameters and such
in the python domain
image
and in the C++ domain
image

docs/api.rst
```diff
@@ -0,0 +1,8 @@
+General API customization
+=========================
+
+.. confval:: include_object_description_fields_in_toc
+
+   :python:`bool` indicating whether to include domain object description
+   fields, like "Parameters", "Returns", "Raises", etc. in the table of
+   contents.  Defaults to :python:`True`.
```
2bndy5, line 8: Different people learn differently. While I'm ok with reading this paragraph and understanding this option's function, some might benefit from a simple example.

Instead of demonstrating all forms of the fields effected, a `seealso` link (or list/table of links) to a demonstration of this would be a good idea. However, due to #58 , it might be better to have a localized example, so the right toc isn't confusing when following a link.

docs/cpp.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. confval:: cpp_generate_synopses
+
+   :python:`bool` specifying whether to generate a *synopsis* for C++ domain
+   objects based on the first paragraph of their content (first sentence for
+   parameters).  The synopsis is shown as a tooltip when hovering over a
+   cross-reference link, and is also shown in the search results list.
+
+   Defaults to :python:`True`.
+
```
2bndy5, line 12: It would be harder to add an example here because the user needs to be instructed where to hover their mouse cursor. Again, maybe a link to an example description that is worded like
> This sentence is part of the tooltip in the right side ToC. Hover your mouse cursor over this member's entry in the right side ToC to see it.

docs/cpp.rst
```diff
@@ -0,0 +1,68 @@
...
+.. confval:: cpp_qualify_parameter_ids
+
+   :python:`bool` specifying whether function, template, and macro parameters
+   should be assigned fully-qualified ids (for cross-linking purposes) of the
+   form ``<parent-id>-p-<param-name>`` based on the id of the parent
+   declaration.  Defaults to :python:`True`.
+
+   If set to :python:`False`, instead the shorter unqualified id
+   ``p-<param-name>`` is used.  This option should only be set to
+   :python:`False` if each C++ declaration is on a separate page.
```
2bndy5, line 53: An example of how to cross-reference a parameter here would be greatly helpful (even for me).

2bndy5:
> > I've noticed that the tooltips for `confval` references no longer show up.
> 
> I wasn't aware that they had tooltips --- can you explain a bit more what you mean?

I must have been mis-remembering this. I thought that hovering over a confval ToC entry or inline xref would yield a tooltip like "option_name (confval)". But, I can't seem to find an example of this anywhere (not even in sphinx-doc.org). Please disregard.

2bndy5:
The param direction for C++ domain looks a bit sequestered.
image
Especially for wide viewports.

I couldn't devise a CSS way to replace the `floatr: right` property. My idea was either append to the end of a param name or use a nested `dl` in which the `dt` is the param direction and the `dd` is the param name(s).

docs/api.rst
jbms: I added some clarification and a link to an example.

docs/cpp.rst
jbms: I added an example.

docs/cpp.rst
jbms: I added an example to the cpp_generate_synopses description.

jbms:
> The param direction for C++ domain looks a bit sequestered. image Especially for wide viewports.
> 
> I couldn't devise a CSS way to replace the `floatr: right` property. My idea was either append to the end of a param name or use a nested `dl` in which the `dt` is the param direction and the `dd` is the param name(s).

I am also not too satisfied with the appearance of this, but I couldn't figure out a better alternative.  Perhaps could be revised in the future.

2bndy5 (approved):
As always these changes go above and beyond my expectations! Thank you for this!

2bndy5:
I'm going to tag the master branch head, but I wanted to ask if you think this is a minor version bump or a patch bump. I've been using minor version bumps for merges from upstream, so I'm leaning toward a patch version bump.

jbms:
In general patch -> bug fixes only, while minor -> backwards compatible feature addition.  I think if we synchronized our version numbers with mkdocs-material then a patch bump could make sense, but so far we have not done that, and as we add more independent sphinx-related functionality to this theme (such as in this PR), the upstream changes from mkdocs-material may not necessarily be the most significant changes.  So I think minor version bump would make sense as well here, and maybe we won't bother to use a patch bump.

2bndy5:
Well I pushed a v0.5.0 tag, and setuptools_scm interpreted it as `0.5.0.post1.dev0` which got treated as a pre-release on pypi. So, fresh installs will still be getting v0.4.1. Pushing a tag v0.5 (after v0.5.0 tag was pushed) wasn't uploaded to pypi (only test-pypi).

I'm guessing the `post1.dev0` is indicative of pre-release status on pypi, but I don't know what I could've done to prevent that from being appended.

jbms:
I have a fix for the version number issue.  I removed the v0.5.0 tag and the post release from pypi --- we can push a new v0.5.0 tag once the PR is merged.

PR #71, opened 2022-04-24
Fix package version

jbms:
This fixes the version number issue https://github.com/jbms/sphinx-immaterial/pull/64#issuecomment-1107765536.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,10 +1,19 @@
-# We need to make sure that the remote and workspace line endings all use LF due to
-# linting and other reasons.
+# Prevent line ending conversions by default to avoid problems with binary files.
+* -text
 
-# Set core.autocrlf to false in case people have this set differently.
-# If this is set to 'true' or 'input', then it overrides the eol option.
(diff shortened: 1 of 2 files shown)


2bndy5:
I think those pictures were my mistake from when I first revised the docs...

.gitattributes
```diff
@@ -1,10 +1,17 @@
...
+*.md text eol=lf
+*.toml text eol=lf
+*.txt text eol=lf
+*.scss text eol=lf
+*.ts text eol=lf
+*.json text eol=lf
+*.html text eol=lf
+*.css text eol=lf
+*.tsx text eol=lf
+*.svg text eol=lf
```
2bndy5, line 16: *.sh and *.xml from for files added in #64

jbms: Done.

2bndy5 (approved):

2bndy5:
I found this cmd helpful when analyzing what files have which line ending
```
git ls-files --eol
```

PR #72, opened 2022-04-24
Fix C++ parameter cross-linking for object descriptions with nested object descriptions

### LINKED ISSUE #67: getting WARNING: Invalid parameter name
I'm now getting warnings saying
```
GitHub\RF24\docs\sphinx\classRF24.rst:6: WARNING: Invalid parameter name: '_cepin'
GitHub\RF24\docs\sphinx\classRF24.rst:6: WARNING: Invalid parameter name: '_cspin'
GitHub\RF24\docs\sphinx\classRF24.rst:6: WARNING: Invalid parameter name: '_spi_speed'
GitHub\RF24\docs\sphinx\classRF24.rst:7: WARNING: Invalid parameter name: '_spi_speed'
```
The only commonality that these parameters have is that they're used ~as verbatim parameter names in multiple functions (~ in overloaded c'tor ~and Arduino style `begin()`)~, and each name begins with a `_`. I might be misunderstanding the problem here, but I doubt that the `_` is the culprit (removing the `_` didn't prevent the warning).

EDITED: better diagnosis in [first comment below](https://github.com/jbms/sphinx-immaterial/issues/67#issuecomment-1107844387)

~I'm guessing this warning should be saying "Non-unique parameter name" given that~ it is generated from
https://github.com/jbms/sphinx-immaterial/blob/163fe76d74cfa2fc75b0d51223417544f844e534/sphinx_immaterial/cpp_domain_fixes.py#L728-L740

~However, I would argue that re-using parameter names for multiple functions is an accepted practice. So, we shouldn't raise warnings about it (which would cause RTD builds to fail fast).~

jbms:
Fixes #67.

### DIFF
--- a/sphinx_immaterial/cpp_domain_fixes.py
+++ b/sphinx_immaterial/cpp_domain_fixes.py
@@ -719,13 +719,17 @@ def _add_parameter_documentation_ids(
 
     qualify_parameter_ids = env.config.cpp_qualify_parameter_ids
 
-    for param_node in list(obj_content.traverse(condition=docutils.nodes.term)):
-        param_name = param_node.get("paramname")
-        if not param_name:
-            continue
+    def cross_link_single_parameter(param_node: docutils.nodes.term) -> None:
         kind = param_node.get("param_kind")
 
-        # Determine the number of unique declarations of this parameter
+        # Determine the number of unique declarations of this parameter.
+        #
+        # If this single object description has multiple signatures, the same
+        # parameter name may be declared in more than one of those signatures.
+        # In the parameter description, we will replace the bare parameter name
+        # with the list of all of the distinct declarations of the parameter.
+        # Identical declarations in more than one signature will only be
+        # included once.
         unique_decls: Dict[str, Tuple[int, docutils.nodes.Element]] = {}
         for i, sig_param_nodes in enumerate(sig_param_nodes_for_signature):
             desc_param_node = sig_param_nodes.get(param_name)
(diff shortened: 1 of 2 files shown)


jbms:
I also added some extra comments here.

.pylintrc
```diff
@@ -107,7 +107,8 @@ disable=raw-checker-failed,
         too-many-ancestors,
         too-many-nested-blocks,
         global-statement,
-        too-many-lines
+        too-many-lines,
+        line-too-long
```
2bndy5, line 111: Disabling this might make formatting with black even more important. Furthermore, there'd be no visual indicator in the IDE that black will need to be run. What line(s) is this needed for?

sphinx_immaterial/cpp_domain_fixes.py
```diff
@@ -734,10 +738,17 @@ def _add_parameter_documentation_ids(
             decl_text = desc_param_node.astext().strip()
             unique_decls.setdefault(decl_text, (i, desc_param_node))
         if not unique_decls:
+            all_params = {}
+            for sig_param_nodes in sig_param_nodes_for_signature:
+                all_params.update(sig_param_nodes)
             logger.warning(
-                "Invalid parameter name: %r", param_name, location=param_node
+                "Parameter name %r does not match any of the parameters "
+                "defined in the signature: %r",
```
2bndy5, line 746: >sphinx_immaterial/cpp_domain_fixes.py:745:0: C0301: Line too long (102/90) (line-too-long)

It looks like this line was split to avoid the line-length mandated in black. You could remove that disabling flag in pylintrc file.

2bndy5:
I can confirm that this works to solve #67 with pure RST and breathe+Doxygen. Thanks for the quick response.

sphinx_immaterial/cpp_domain_fixes.py
jbms: Done.

.pylintrc
jbms: It seemed redundant since black already handles that, and black doesn't split long string literals.  But anyway I removed this change since as you noted I fixed it by splitting the string manually.

jbms:
Thanks for your review!

Issue #69, opened 2022-04-24
synopsis is inconsistent with doxygen+breathe


2bndy5:
This isn't really a problem with this theme's synopsis feature.

I noticed that [`sphinx_utils.summarize_element_text()`](https://github.com/jbms/sphinx-immaterial/blob/163fe76d74cfa2fc75b0d51223417544f844e534/sphinx_immaterial/sphinx_utils.py#L6) may yeild the wrong text because doxygen's XML tends to use different paragraph structure, and breathe simply attempts to output a node tree based on the doxygen XML output (1:1). 

So, I'm suggesting a note about synopsis compatibility with other sphinx extensions that don't use `nodes.paragraph` exactly like sphinx/docutils does.

### expected results
useing the RST code
```rst
    .. cpp:function:: RF24::RF24(uint32_t _spi_speed=RF24_SPI_SPEED)

        :param _spi_speed: The SPI speed in Hz ie: 1000000 == 1Mhz

            - Users can specify default SPI speed by modifying :c:macro:`RF24_SPI_SPEED` in :file:`RF24_config.h`
        
              - For Arduino, the default SPI speed will only be properly configured this way on devices supporting SPI TRANSACTIONS
              - Older/Unsupported Arduino devices will use a default clock divider & settings configuration
              - For Linux: The old way of setting SPI speeds using BCM2835 driver enums has been removed as of v1.3.7
```

<details><summary>parameter's pformat() output</summary>

```xml
<definition>
    <paragraph>
        <inline translatable="True">
            The SPI speed in Hz ie: 1000000 == 1Mhz
            <bullet_list>
                <list_item>
                    <paragraph>
                        Users can specify default SPI speed by modifying
                        <pending_xref refdomain="std" refexplicit="True" refid="RF24__config_8h_1a63ef6fd21b0b83db10e95c0da126d1e2" reftarget="RF24__config_8h_1a63ef6fd21b0b83db10e95c0da126d1e2" reftype="ref">
                            RF24_SPI_SPEED
                         in
                        <pending_xref refdomain="std" refexplicit="True" refid="RF24__config_8h" reftarget="RF24__config_8h" reftype="ref">
                            RF24_config.h
                        <bullet_list>
                            <list_item>
                                <paragraph>
                                    For Arduino, the default SPI speed will only be properly configured this way on devices supporting SPI TRANSACTIONS
                            <list_item>
                                <paragraph>
[...]

PR #75, opened 2022-04-26
Add object description options mechanism

### LINKED ISSUE #70: add option to disable cross-linking C++ parameters
There are 2 reasons why I think this would be useful.
1. If a documented API uses a lot of parameters, then the ToC gets very long and visually complex (despite the `include_object_description_fields_in_toc` config value).
2. When writing new documentation for API, it can be annoying to see a warning related to missing parameter descriptions.
   - I wouldn't do this myself, but some might consider a parameter's name/type as "self-descriptive enough." Thus, they would deliberately neglect adding parameter description(s).

The only objection I can think of is: If disabling cross-linked parameters, **would it affect other C++ domain customization** options, namely synopses?

I imagine that the option's name would be `cpp_cross_link_parameters` (defaults to True). This way it can be used independently from similar implementations for other domains (like Java or JS) - thinking optimistically about future additions 😃 .

jbms:
This replaces manny config options with a more general "object
description options" mechanism that allows various options to be
configured on a per domain/object type basis.

This commit also adds a number of features built on top of that mechanism:

- Generating object description synopses for the std domain, including
  custom object types added using `app.add_object_type`.

- Excluding certain object types from the TOC

- Excluding the object type from the cross-reference tooltip.

Fixes #70.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,11 +1,112 @@
 General API customization
 =========================
 
-.. confval:: include_object_description_fields_in_toc
+This theme supports a number of options that can be customized for each
+domain/object type pair.
 
-   :python:`bool` indicating whether to include domain object description
-   fields, like "Parameters", "Returns", "Raises", etc. in the table of
-   contents.  Defaults to :python:`True`.
+.. confval:: object_description_options
+
+   :python:`list` of :python:`(pattern, options)` pairs, where :python:`pattern`
+   is a regular expression matching strings of the form
+   :python:`"domain:objtype"` and :python:`options` is a dictionary of supported
+   `object-description-options`.
+
+   We need something `object_description_options`.
+
+   The actual options for a given object type are determined by first
+   initializing each option to its default value, and then applying as overrides
+   the options associated with each matching pattern in order.
+
+.. _object-description-options:
+
+Object description options
+--------------------------
+
+The following options can be customized for each object type using
+:confval:`object_description_options`.
+
+.. objconf:: include_in_toc
+
+   Indicates whether to include the object description in the table of contents.
+
+   .. admonition:: Example
+      :class: example
+
+      To prevent C++ parameter descriptions from appearing in the TOC, add the
+      following to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options = [
+             ("cpp:.*Param": dict(include_in_toc=False)),
+         ]
+
+.. objconf:: generate_synopses
+
+   Indicates whether to generate a *synopsis* from the object description.  The
+   synopsis is shown as a tooltip when hovering over a cross-reference link, and
+   is also shown in the search results list.  Supported values are:
+
+   :python:`None`
+     Disables synopsis generation.
+
+   :python:`"first_paragraph"`
+     Uses the first paragraph of the description as the synopsis.
+
+   :python:`"first_sentence"`
+     Uses the first sentence of the first paragraph of the description as the synopsis.
+
+   The default is :python:`"first_paragraph"` except for :regexp:`c(pp)?:.*Param`
+   where the default is :python:`"first_sentence"`.
+
+   .. note::
+
+      Synopsis generation is currently supported only for the following domains:
+
+      - ``std`` (including object types added using :py:obj:`sphinx.application.Sphinx.add_object_type`)
+      - ``c`` and ``cpp``
+
+   .. admonition:: Example
+      :class: example
+
+      To use the first sentence rather than the first paragraph as the synopsis
+      for C++ class descriptions, add the following to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options = [
+             ("cpp:class", dict(generate_synopses="first_sentence")),
+         ]
+
+.. objconf:: include_object_type_in_xref_tooltip
+
+   Indicates whether to include the object type in cross-reference and TOC
+   tooltips.
+
+   .. note::
+
+      For TOC entries, this is supported for all domains.  For regular cross
+      references, this is supported only for the following domains:
+
+      - ``std`` (including object types added using :py:obj:`sphinx.application.Sphinx.add_object_type`)
+      - ``c`` and ``cpp``
+
+   .. admonition:: Example
+      :class: example
+
+      To exclude the object type from all ``py`` domain xrefs, add the following
+      to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options = [
+             ("py:.*", dict(include_object_type_in_xref_tooltip=False)),
+         ]
+
+.. objconf:: include_fields_in_toc
+
+   Indicates whether to include fields, like "Parameters", "Returns", "Raises",
+   etc. in the table of contents.
 
    For an example, see: :cpp:expr:`synopses_ex::Foo` and note the ``Template
    Parameters``, ``Parameters``, and ``Returns`` headings shown in the
@@ -13,11 +114,73 @@ General API customization
 
    .. note::
 
-      This option does not control whether there are separate TOC entries for
-      individual parameters, such as for :cpp:expr:`synopses_ex::Foo::T`,
+      To control whether there are separate TOC entries for individual
+      parameters, such as for :cpp:expr:`synopses_ex::Foo::T`,
       :cpp:expr:`synopses_ex::Foo::N`, :cpp:expr:`synopses_ex::Foo::param`, and
-      :cpp:expr:`synopses_ex::Foo::arr`.  Currently, for the C and C++ domains,
-      any parameter documented by a :rst:``:param x:`` field will always result
-      in a TOC entry, regardless of the value of
-      :confval:`include_object_description_fields_in_toc`.  Other domains are
-      not yet supported.
+      :cpp:expr:`synopses_ex::Foo::arr`, use the :objconf:`include_in_toc`
+      option.
+
+
+   .. admonition:: Example
+      :class: example
+
+      To exclude object description fields from the table of contents for all
+      ``py`` domain objects, add the following to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options = [
+             ("py:.*", dict(include_fields_in_toc=False)),
+         ]
+
+Other options described elsewhere include:
+
+- :objconf:`wrap_signatures_with_css`
+- :objconf:`wrap_signatures_column_limit`
+- :objconf:`clang_format_style`
+
+Table of contents icons
+^^^^^^^^^^^^^^^^^^^^^^^
+
+For object descriptions included in the table of contents (when
+:objconf:`include_in_toc` is :python:`True`), a text-based "icon" can optionally
+be included to indicate the object type.
+
+Default icons are specified for a number of object types, but they can be
+overridden using the following options:
+
+.. objconf:: toc_icon_class
+
+   Indicates the icon class, or :python:`None` to disable the icon.  The class
+   must be one of:
+
+   - :python:`"alias"`
+   - :python:`"procedure"`
+   - :python:`"data"`
+   - :python:`"sub-data"`
+
+.. objconf:: toc_icon_text
+
+   Indicates the text content of the icon, or :python:`None` to disable the
+   icon.  This should normally be a single character, such as :python:`"C"` to
+   indicate a class or :python:`"F"` to indicate a function.
+
+.. admonition:: Example
+   :class: example
+
+   To define a custom object type and specify an icon for it, add the following to
+   :file:`conf.py`:
+
+   .. code-block:: python
+
+      object_description_options = [
+          ("std:confval", dict(toc_icon_class="data", toc_icon_text="C")),
+      ]
+
+      def setup(app):
+          app.add_object_type(
+              "confval",
+              "confval",
+              objname="configuration value",
+              indextemplate="pair: %s; configuration value",
+          )
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,6 +12,17 @@ import sys
 
 sys.path.insert(0, os.path.abspath("."))
 
+import docutils
+import sphinx
+import sphinx.domains.python
+import sphinx.environment
+import sphinx.util.logging
+import sphinx.util.typing
+
+from sphinx_immaterial import apidoc_formatting
+
+logger = sphinx.util.logging.getLogger(__name__)
+
 # -- Project information -----------------------------------------------------
 
 project = "Sphinx-Immaterial"
@@ -161,16 +172,15 @@ extlinks = {
     "dudir": ("http://docutils.sourceforge.net/docs/ref/rst/directives.html#%s", ""),
 }
 
+object_description_options = []
+
 # BEGIN: sphinx_immaterial.format_signatures extension options
-clang_format_signatures_domain_styles = {
-    "cpp": """{
-    BasedOnStyle: LLVM,
-    ColumnLimit: 68,
-    }""",
-}
+object_description_options.append(
+    ("cpp:.*", dict(clang_format_style={"BasedOnStyle": "LLVM"}))
+)
 # END: sphinx_immaterial.format_signatures extension options
 
-html_wrap_signatures_with_css = ["py"]
+object_description_options.append(("py:.*", dict(wrap_signatures_with_css=True)))
 
 # BEGIN: sphinx_immaterial.external_cpp_references extension options
 external_cpp_references = {
@@ -210,6 +220,50 @@ rst_prolog = """
 """
 
 
+object_description_options.append(
+    (
+        "std:confval",
+        dict(
+            toc_icon_class="data", toc_icon_text="C", generate_synopses="first_sentence"
+        ),
+    )
+)
+
+object_description_options.append(
+    (
+        "std:objconf",
+        dict(
+            toc_icon_class="data", toc_icon_text="O", generate_synopses=None,
+        ),
+    )
+)
+
+
+def _parse_object_description_signature(
+    env: sphinx.environment.BuildEnvironment, signature: str, node: docutils.nodes.Node
+) -> str:
+    registry = apidoc_formatting.get_object_description_option_registry(env.app)
+    registry_option = registry.get(signature)
+    node += sphinx.addnodes.desc_name(signature, signature)
+    if registry_option is None:
+        logger.error("Invalid object description option: %r", signature)
+    else:
+        node += sphinx.addnodes.desc_sig_punctuation(" : ", " : ")
+        annotations = sphinx.domains.python._parse_annotation(
+            sphinx.util.typing.stringify(registry_option.type_constraint), env
+        )
+        node += sphinx.addnodes.desc_type("", "", *annotations)
+        node += sphinx.addnodes.desc_sig_punctuation(" = ", " = ")
+        default_repr = repr(registry_option.default)
+        node += docutils.nodes.literal(
+            default_repr,
+            default_repr,
+            language="python",
+            classes=["python", "code", "highlight"],
+        )
+    return signature
+
+
 def setup(app):
     app.add_object_type(
         "confval",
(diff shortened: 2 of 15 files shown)


docs/conf.py
```diff
@@ -161,16 +161,16 @@
     "dudir": ("http://docutils.sourceforge.net/docs/ref/rst/directives.html#%s", ""),
 }
 
+object_description_options_map = {}
+object_description_options_patterns = []
+
 # BEGIN: sphinx_immaterial.format_signatures extension options
```
2bndy5, line 167: should probably move this comment (and the `# END: ...` comment) so the config option changes show up in the doc's `literalblock`.

docs/conf.py
```diff
@@ -161,16 +161,16 @@
...
-    ColumnLimit: 68,
-    }""",
-}
+object_description_options_patterns.append(("cpp:.*", {"BasedOnStyle": "LLVM"}))
 # END: sphinx_immaterial.format_signatures extension options
 
-html_wrap_signatures_with_css = ["py"]
+object_description_options_patterns.append(
+    ("py:.*", dict(wrap_signatures_with_css=True))
+)
```
2bndy5, line 173: I just pushed the v0.5.0 tag without knowing your work on this was changing the config option names. These config option changes seem like a breaking change, so I'd use a major version bump for the next tag.

docs/conf.py
2bndy5: Oh, I see what you've done to the docs now. Please disregard this suggestion.

However, using `append()` in an example snippet implies you can append to the default value (without declaring it like you have done properly here).

docs/conf.py
jbms: It is true that it is a breaking change, but on the other hand if we get this in soon I doubt too many people will have used the new options yet.

Also v0 generally means no stability guarantees.

But yeah it could make sense to increase to v1.

docs/conf.py
jbms: Yeah unfortunately that is not possible.  But we have done the same with appending to extensions.  Not sure what the best approach is.

jbms: (The reason it is not possible is that the config file has to be evaluated in order to determine which extensions to load, but extensions can define config options, so we can't know what the defaults are until after loading the config)

docs/api.rst
```diff
@@ -1,23 +1,207 @@
...
+   where the default is :python:`"first_sentence"`.
+
+   .. note::
+
+      Synopsis generation is currently supported only for the following domains:
+
+      - ``std`` (including object types added using
+        :py:obj:`sphinx.application.Application.add_object_type`);
+
+      - ``c`` and ``cpp``
```
2bndy5, line 85: This list looks funny
image

1. Not sure why the `:py:obj:`sphinx.application.Application.add_object_type`);` gets put on a separate line.
2. don't need a blank line between list items unless the items are supposed to be separate lists.
```suggestion
      - ``std`` (including object types added using :py:obj:`sphinx.application.Application.add_object_type`);
      - ``c`` and ``cpp``
```

docs/conf.py
2bndy5: I'm ok with bumping minor version because I'd also consider v0 as "beta" state.

docs/format_signatures.rst
```diff
@@ -65,35 +69,23 @@ set the :confval:`clang_format_signatures_domain_styles` configuration option.
...
-   <https://clang.llvm.org/docs/ClangFormatStyleOptions.html>`__ for each
-   domain.  Formatting is enabled *only* for domains that are listed.
+   Specifies the `clang-format style options
+   <https://clang.llvm.org/docs/ClangFormatStyleOptions.html>`__ as a
+   :python:`dict` (JSON object), or :python:`None` to disable clang-format.
 
    .. literalinclude:: conf.py
       :language: python
       :start-after: # BEGIN: sphinx_immaterial.format_signatures extension options
       :end-before: # END: sphinx_immaterial.format_signatures extension options
```
2bndy5, line 85: Here is the snippet that has the misleading `append()` call.
```suggestion
   .. code-block:: py

      object_description_options_patterns = [("cpp:.*", {"BasedOnStyle": "LLVM"})]
```

docs/conf.py
2bndy5: I'm resolving this as my main concern here is addressed by [this review comment](https://github.com/jbms/sphinx-immaterial/pull/75/files#r858192878).

sphinx_immaterial/apidoc_formatting.py
```diff
@@ -11,6 +15,25 @@
...
+        include_object_description_fields_in_toc: bool
+        toc_icon_text: Optional[str]
+        toc_icon_class: Optional[str]
+
+
+else:
+    ObjectDescriptionOptions = dict
+
+
 if TYPE_CHECKING:
```
2bndy5, line 37: We're not using mypy for type checking in CI. This feels specific to using mypy after looking into it. Pylint doesn't do type checking to the extent that mypy does - mypy is the org that introduced type checking in python and provided the mypy tool to make proper use of `typing` module.

docs/format_signatures.rst
jbms: Okay, I removed the literalinclude since it is misleading.

sphinx_immaterial/apidoc_formatting.py
jbms: We might want to use mypy in the future, but for now I've removed this.

docs/api.rst
jbms: Fixed.

docs/format_signatures.rst
jbms: And actually there was a bug in conf.py (now fixed) that meant clang-format wasn't being used at all.

sphinx_immaterial/apidoc_formatting.py
2bndy5: Good idea for a separate PR. I tend to avoid mypy out of laziness, but it makes sense... I primarily use type checking to augment VSCode's intellisense tool (& "go to definition" option).

docs/conf.py
```diff
@@ -210,10 +212,33 @@
...
+object_description_options_map["std:objconf"] = dict(
+    toc_icon_class="data", toc_icon_text="O", generate_synopses="first_sentence"
+)
+
+
+def _validate_parallel_build(app):
+    # Verifies that all of the extensions defined by this theme support parallel
+    # building.
+    assert app.is_parallel_allowed("read")
+    assert app.is_parallel_allowed("write")
```
2bndy5, line 228: This feels like a "slap on the wrist". 🤣 My bad.

jbms: It is easy to miss since we had no test, and I also missed it when reviewing.  I only notice when building the TensorStore docs since they are large and slow to build without parallel building.

sphinx_immaterial/apidoc_formatting.py
```diff
@@ -200,17 +334,16 @@ def setup(app: sphinx.application.Sphinx):
     # to apply.
     sphinx.addnodes.desc_signature.classes.append("highlight")
 
-    app.add_config_value("html_wrap_signatures_with_css", default=None, rebuild="env")
+    app.connect("object-description-transform", _wrap_signatures)
+    _monkey_patch_object_description_to_include_fields_in_toc()
+
     app.add_config_value(
-        "html_wrap_signatures_with_css_column_limit", default=68, rebuild="env"
+        "object_description_options_map", default=[], rebuild="env", types=(dict,)
```
2bndy5, line 341: Would it make sense to ease usability if we create a class(es) that can be used as user input to avoid erroneous/unsupported values? The revised config options seem a bit more complex (in terms of user input) than they were.

2bndy5: I know this is an idea for a different PR, and I'm still trying wrap my head around the config changes. It looks like `object_description_options_patterns` can span several domains, and the user won't get notified if they typo'd an option in the build log. Sphinx intuitively lets the user know about an unrecognized/unused option in `html_theme_options`, but this intuition won't be extended to `object_description_options_patterns` or `object_description_options_map`.

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+      .. code-block:: python
+
+         object_description_options_map = [
+             ("py:.*", dict(include_object_description_fields_in_toc=False)),
+         ]
+
+Other options described elsewhere include:
+
+- :objconf:`html_wrap_signatures_with_css`
+- :objconf:`html_wrap_signatures_with_css_column_limit`
```
2bndy5, line 156:
```suggestion
- :objconf:`wrap_signatures_with_css`
- :objconf:`wrap_signatures_with_css_column_limit`
```
should fix the resolution of the pending refs
image

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+   :python:`None`
+     Disables synopsis generation.
+
+   :python:`"first_paragraph"`
+     Uses the first paragraph of the description as the synopsis.
+
+   :python:`"first_sentence"`
+     Uses the first sentence of the first paragraph of the description as the synopsis.
+
+   The default is :python:`"first_paragraph"` except for ``c(pp)?:.*Param``
```
2bndy5, line 75: I think you can use the `:regexp:` role here.
```suggestion
   The default is :python:`"first_paragraph"` except for :regexp:`c(pp)?:.*Param`
```

sphinx_immaterial/apidoc_formatting.py
jbms: That is a good point.   Right now there is no registry of object description options, but we could add e.g. a `add_object_description_option` function to register them, and then could validate them like sphinx does for the config options.

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+      .. code-block:: python
+
+         object_description_options_map = [
+             ("py:.*", dict(include_object_description_fields_in_toc=False)),
+         ]
+
+Other options described elsewhere include:
+
+- :objconf:`wrap_signatures_with_css`
+- :objconf:`wrap_signatures_with_css_column_limit`
```
2bndy5, line 156: Damn! I had a typo that last suggestion. The xref is still unresolved.
```suggestion
- :objconf:`wrap_signatures_column_limit`
```

docs/conf.py
```diff
@@ -210,10 +212,33 @@
 """
 
 
+object_description_options_map["std:confval"] = dict(
+    toc_icon_class="data", toc_icon_text="C", generate_synopses="first_sentence"
+)
```
2bndy5, line 217: Just an observation: It looks like this will only generate a synopsis when the `confval` role is used.

2bndy5 (approved):

2bndy5:
Please hold. I'm having trouble getting the solution to #70 working. I'm trying
```py
object_description_pattern = [
    ("cpp:functionParam", dict(include_in_toc=False, generate_synopses=None))
]
```
But it still shows param in the toc and synopses are the entire paragraph (albeit still incorrect because of #69 )

2bndy5:
Nevermind. I see the typo in the config option now. That also explains why the docs weren't rebuilding when I changed the typo'd option's setting.

All is good to go when you are.

2bndy5:
According to the docs you added, the `object_description_options_map` is a dict and `object_description_options_patterns` is a list. But all example snippets that use a list are assigned to `object_description_options_map`.

I'm starting to notice that this more generic approach (making the user specify a specific object type) is too generic as it requires intimate knowledge of sphinx inner workings (most of which aren't really documented like functionParam). Add regex into the mix, and you've got a pretty confusing user interface.

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+
+   .. admonition:: Example
+      :class: example
+
+      To use the first sentence rather than the first paragraph as the synopsis
+      for C++ class descriptions, add the following to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options_map = [
```
2bndy5, line 93: Should be a dict or ``object_description_options_patterns`

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+
+   .. admonition:: Example
+      :class: example
+
+      To exclude object description fields from the table of contents for all
+      ``py`` domain objects, add the following to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options_map = [
```
2bndy5, line 149: Should be a dict or `object_description_options_patterns`

jbms:
I will have to fix the doc issues.  It could be nice to have a more powerful version of rst-example that lets us use  modified conf.py but that might be tricky to implement.

Most of the time the object type matches the directive name so it should be fairly easy to understand.  However it is true that the param types are hidden --- we would need to document them.

docs/api.rst
2bndy5, line 149: Should be a dict or `object_description_options_patterns`

2bndy5 (changes requested):

2bndy5:
Originally, I had two rst-results type directives. Only the simpler form made it to PR. The other more complicated form was meant to take the first literal block designated as RST (via the node's `language` property) and output that rendered. What I wanted was to be able to feed it tabbed content and show the result below the tabbed content (despite which tab was active). But, it was harder to implement (trying to compensate for user error) and buggy (absent a closing html div tag).

Wouldn't using a modified conf.py for an example snippet require a independent (or possible divergent from primary) sphinx env object? That sounds a bit more than just "tricky". It would be great for tests/validation though.

2bndy5:
I could revisit the tabbed results idea and grab the first literal block with a caption title ending in "conf.py" (probably would use `os.path` for that). Then any code in the conf.py snippet would be temporarily augmented to the `app.env.config` (restoring the env would be the tricky part). This is all subject to using `code-block` directive properly (via `language` and `caption` properties). I would still need to solve the missing `</div>` problem though (felt like I was going bald trying to resolve that).

jbms:
Yeah, using a modified conf.py would require running a separate sphinx build entirely (but presumably just for a single page) and then somehow incorporating the result into the main documentation.  Probably would work best as a link to a separate page, or embedded as an iframe that is zoomed out perhaps.

This would be particularly nice for demonstrating the mkdocs-material `features` options.

2bndy5:
`iframe` idea might be more complicated since it uses a isolated DOM.

2bndy5:
I also can't seem to get `include_object_description_fields_in_toc=False` to work.
```py
object_description_options_patterns = [
    ("cpp:.*", dict(include_object_description_fields_in_toc=False)),
]
```

2bndy5:
It seems that option is actually named `include_fields_in_toc` now. `("cpp:.*", dict(include_fields_in_toc=False))` works.

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+      To exclude the object type from all ``py`` domain xrefs, add the following
+      to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options_map = [
+             ("py:.*", dict(include_object_type_in_xref_tooltip=False)),
+         ]
+
+.. objconf:: include_object_description_fields_in_toc
```
2bndy5, line 122:
```suggestion
.. objconf:: include_fields_in_toc
```

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+   .. admonition:: Example
+      :class: example
+
+      To exclude object description fields from the table of contents for all
+      ``py`` domain objects, add the following to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options_map = [
+             ("py:.*", dict(include_object_description_fields_in_toc=False)),
```
2bndy5, line 150:
```suggestion
             ("py:.*", dict(include_fields_in_toc=False)),
```

2bndy5 (changes requested):

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
 
-   :python:`bool` indicating whether to include domain object description
-   fields, like "Parameters", "Returns", "Raises", etc. in the table of
-   contents.  Defaults to :python:`True`.
+.. confval:: object_description_options_map
+
+   :py:obj:`dict` mapping strings of the form :python:`"domain:objtype"` to a
+   dictionary of supported `object-description-options`.
+
+   This option may be used to specify options for a single option type.  To
```
2bndy5, line 12: I'm partial to just removing this config option entirely in favor of only using `object_description_options_patterns`. This would make our lives a bit simpler, yes?

What is stopping
```py
object_description_options_patterns = [
    ("cpp:functionParam", dict(include_in_toc=False, generate_synopses=None)),
    ("cpp:function", dict(include_fields_in_toc=False, generate_synopses=None)),
]
```
from being equivalent to 
```py
object_description_options_map = {
    "cpp:functionParam": dict(include_in_toc=False, generate_synopses=None),
    "cpp:function": dict(include_fields_in_toc=False, generate_synopses=None),
}
```

It looks like regex is only applied to the `object_description_options_patterns` in which case this needs to be altered.
```suggestion
   This option may be used to specify options for a single object type.  To
```

sphinx_immaterial/apidoc_formatting.py
jbms: I added a registry and added type validation using pydantic.

docs/conf.py
jbms: The default value is "first_paragraph", so it will generate a synopsis by default, this just overrides it to use first_sentence.

docs/api.rst
jbms: Fixed.

docs/api.rst
```diff
@@ -1,23 +1,203 @@
...
+
+   .. admonition:: Example
+      :class: example
+
+      To exclude the object type from all ``py`` domain xrefs, add the following
+      to :file:`conf.py`:
+
+      .. code-block:: python
+
+         object_description_options_map = [
```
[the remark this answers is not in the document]
jbms: Fixed.

jbms:
> It seems that option is actually named `include_fields_in_toc` now. `("cpp:.*", dict(include_fields_in_toc=False))` works.

Fixed.

docs/api.rst
jbms: Done.  Originally I added this for efficiency, to avoid having to scan through all patterns for all object types, given that there are a lot of rules for `toc_icon_class`.  The options are cached per object type, so it wouldn't be too bad, but I am still wary of adding any quadratic complexity to the build.

However I realized I could still apply the same optimization even without needing a separate map.

docs/conf.py
2bndy5: I understands the code here. I'm just noting that
```rst
.. confval:: html_theme_options

   Some synopsis statement.

   - This gets no synopsis: `html_theme_options`, but the xref target it resolved.
   - This gets a synopsis: :confval:`html_theme_options`.
```

sphinx_immaterial/apidoc_formatting.py
2bndy5: I'm now getting an error (repeated for each time I set `generate_synopses=None`):
```
ERROR: Invalid value None for object description option 'generate_synopses' specified for pattern re.compile('cpp:functionParam'): 1 validation error for ParsingModel[Literal[None, 'first_paragraph', 'first_sentence']]  
__root__
  none is not an allowed value (type=type_error.none.not_allowed)
```
 when using the config:
```py
    ("cpp:functionParam", dict(include_in_toc=False, generate_synopses=None)),
    ("cpp:function", dict(include_fields_in_toc=False, generate_synopses=None)),
]
```

docs/conf.py
jbms: Ah, sorry --- I misunderstood.  Yes, I had forgotten to also monkey patch `resolve_any_xref`.  Now fixed.

sphinx_immaterial/apidoc_formatting.py
jbms: It seems like there may be a bug in pydantic.  In any case I changed the type to `Optional[Literal["first_sentence", "first_paragraph"]]` as a workaround, and that seems to solve the issue.

2bndy5 (approved):
Testing looks good now. I can't find any other objections.

jbms:
Thanks for the very thorough review!

PR #77, opened 2022-04-28
Fix styling of punctuation in signatures

### LINKED ISSUE #76: API signature punctuation has modified CSS
A small observation, but...

I noticed that after the changes in #60 through #73 the API signature punctuation color has changed. It used to be `var(--md-code-hl-punctuation-color)`, but now is overridden in some way. 

This can be reproduced by adding any function to your docs that has a `List[int]`, or some equivalent type hint. The `[` are considered punctuation. Notice they were gray before but are pink now.

### 8c96b11dd87e008e0e827c49c211efc401a68afd

image

### 0e0dcc8c9cc88f2d160ec80a38410547c8a904d3

image

jbms:
Fixes #76.

### DIFF
--- a/src/assets/stylesheets/main/_api.scss
+++ b/src/assets/stylesheets/main/_api.scss
@@ -78,7 +78,7 @@ $objinfo-icon-size: 16px;
       > a.reference {
         color: $api-color-type;
         // Prevent pygments highlight rules from taking precedence.
-        * {
+        .n {
           color: inherit;
         }
       }


jbms:
I think this is the simplest fix --- we just need to make sure it doesn't inadvertently break something else.

2bndy5 (approved):
It looks to me like #76 was already fixed in #75 , but I'm in favor of this change since it is less generic and doesn't regress the current solution (as far as I can tell).

2bndy5:
I don't think that adding mypy to intersphinx config settings will allow hyperlinking abstracted `types` (like `Tuple`, `Dict`, etc). It looks like the [mypy docs don't actually declare the abstract types using the `py` domain](https://mypy.readthedocs.io/en/stable/builtin_types.html), so we can't hyperlink the type hints. Currently, the python docs (already in our intersphinx config) do provide the abstractions found in [`collections.abc` module](https://docs.python.org/3/library/collections.abc.html#module-collections.abc).

mhostetter:
I can confirm this PR fixes #76 for me. I see no side effects in my docs. Thanks!

Also, the `object_description_options` config is pretty neat! 🎉

jbms:
The issue with hyperlinking Tuple, Dict and other types defined by the typing module is simply that they are defined in the typing module.  In the TensorStore docs I have monkey patched the Python domain xref resolution to handle these as a special case.

mhostetter:
@2bndy5 I think that's a bug in Sphinx. Notice, `Union` links correctly [here](https://jbms.github.io/sphinx-immaterial//demo_api.html#test_py_module.test.func), but `Optional` does not [here](https://jbms.github.io/sphinx-immaterial//api.html#objconf-toc_icon_class).

I noticed something similar in my docs and added the `autodoc-process-signature` callback to resolve it. (I copied things I saw @jbms do in tensorstore)

```python
def autodoc_process_signature(app, what, name, obj, options, signature, return_annotation):
    signature = modify_type_hints(signature)
    return_annotation = modify_type_hints(return_annotation)

    return signature, return_annotation

def modify_type_hints(signature):
    if signature:
        # Ensure types from the typing module are properly hyperlinked
        for type_name in ["Tuple", "List", "Sequence", "Dict", "Optional", "Union", "Iterator", "Type", "Literal"]:
            signature = signature.replace(f"{type_name}[", f"~typing.{type_name}[")
            signature = signature.replace("~typing.~typing", "~typing")

    return signature

def setup(app):
    app.connect("autodoc-process-signature", autodoc_process_signature)
```

mhostetter:
Jinx... lol

mhostetter:
I found that in functions and methods, `typing.List` was provided in the signature. However in properties, only `List` was provided, which didn't cross reference. A modified version of the above code was meant to add the `typing.` prefix to instances of `List` (etc) in signatures that didn't have it. 🤷

PR #79, opened 2022-04-28
Add Python type annotation transformation options

jbms:


docs/conf.py
```diff
@@ -271,12 +297,51 @@ def _parse_object_description_signature(
     return signature
 
 
+def _parse_confval_signature(
```
2bndy5, line 300: I'm guessing this is another test to make sure an xref to a confval is resolved?

docs/python.rst
```diff
@@ -0,0 +1,78 @@
...
+.. confval:: python_transform_type_annotations_pep585
+
+   Specifies whether to add the following :pep:`585` mappings to
+   :confval:`python_type_aliases`:
+
+   .. jinja:: sys
+
+      {%- for k, v in sys.modules["sphinx_immaterial.python_type_annotation_transforms"].PEP585_ALIASES.items() %}
+      - :py:obj:`{{ k }}` -> :py:obj:`{{ v }}`
+      {% endfor %}
```
2bndy5, line 47: image
Its not too important to me because users can type hint directly from the `re` namespace. However, I'm see unresolved references here

2bndy5 (commented):
Now that we're using `sphinx-jinja`, we could probably throw in some docs about domain-specific parameter object types from https://github.com/jbms/sphinx-immaterial/blob/7cb46822fed2571606bf3aa7511efc5bd469e6a7/sphinx_immaterial/apidoc_formatting.py#L231

docs/python.rst
jbms: Yeah, I think that is just because the Python docs don't include those symbols.

docs/conf.py
jbms: Yeah it checks for that, but it also pulls in the type and default value, so that we no longer have to manually include that information in the description text.

Using a type like `Dict[str, str]` in `add_config_value` is a bit questionable since Sphinx doesn't check them properly (in contrast the object_description_options are properly type checked), but still they seem to work.

docs/python.rst
2bndy5: Best to keep it as is then to demonstrate py-docs' shortcomings then.

2bndy5:
> I'm not sure what makes sense as far as defaults for all of these options.

The booleans make sense. In a perfect world, I think that `python_type_aliases` would use the actual imported symbol, but using strings makes sense implementation wise.

I can't think of any objections to any of this. Let me give it a spin on my CirPy_nRF24 project...

2bndy5 (approved):
I forgot that I have avoided using `typing` in my CirPy_nRF24 lib because CircuitPython doesn't have the a ported variant of CPython's `typing` module (due to limited memory resources). The good news is that nothing is broken.

jbms:
What do you think about the `concise_literals` change?  Since that is not valid Python syntax, should we still apply that transformation by default?

2bndy5:
I actually like that change. I think there was a recent PEP that now prefers the `|` syntax. Primary reason that I didn't take to type hinting in python is because I tend to abuse the dynamic typing of python and the rendered type hints in docs tend to be "wordy" (which makes me think it would only add confusion to the reader's experience).

I guess some square brackets would help differentiate from other data in the signature (like default values). Since python uses the `or` instead of `|`, we're not really at risk of ambitiousness between concise literals vs default values

jbms:
An example where concise literals are ambiguous is the following:

```
class Foo:
  pass

def foo(t: Literal[Foo]):
  pass

foo(Foo)
```'

With the concise literals transform, this would be documented as:

```
foo(t: Foo)
```

which is incorrect.

**Comment by mhostetter** [comment]
Is there a case when a `Literal` with only one option is used? Wouldn't the argument be superfluous at that point? I suspect this case wouldn't come up much in practice.

Really like these additions, by the way!! 🔥

**Comment by jbms** [comment]
Probably the 1-argument Literal would only be used as part of a larger type, e.g. `Union[int, Literal[False]]` or something like that.

There is still an ambiguity:

```python
class Foo: pass
class Bar: pass

def foo(t: Literal[Foo, Bar]): pass
```

Here `foo` would be incorrectly documented as:

```python
foo(t: Foo | Bar)
```

jbms:
I added an additional restriction to only use the concise syntax for constants.  That avoids the ambiguity.

PR #83, opened 2022-04-30
Fix concise literals with autodoc type hint shortening

### LINKED ISSUE #82: Literal shortening has issue with autodoc type hint shortening
I love the new features in #79!! Thank you for them. 🙏 

However, I have noticed a couple minor issues. One is that `Union` and `Literal` conversions with `python_transform_type_annotations_pep604 = True` have an issue when type shortening is used with autodoc, namely the `autodoc_typehints_format = "short"` feature (here's the [docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints_format)).

I know that feature converts `typing.Literal` to `~typing.Literal`, which is likely where the extra ~ comes from. I've been looking at the code to try to find a quick fix, but haven't found it yet.

Here's an example ripped from my code. But any function with a literal will work to reproduce.

```python
def irreducible_poly(order: int, degree: int, method: Literal["min", "max", "random"] = "min") -> Poly:
    pass
```

### Without using `autodoc_typehints_format = "short"`

image

### Using `autodoc_typehints_format = "short"`

image

jbms:
Fixes #82.

### DIFF
--- a/docs/python.rst
+++ b/docs/python.rst
@@ -71,6 +71,9 @@ Python domain customization
       .. py:function:: foo(x: typing.Literal[1, 2, "abc"])
          :noindex:
 
+      .. py:function:: foo(x: ~typing.Literal[1, 2, "abc"])
+         :noindex:
+
    .. warning::
 
       The concise syntax is non-standard and not accepted by Python type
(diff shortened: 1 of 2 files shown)


2bndy5 (approved):
I can confiem that this works with the config
```py
autodoc_typehints_format = "short"
```

PR #86, opened 2022-04-30
Use typing_extensions to support Python 3.7

### LINKED ISSUE #84: `Literal` not available in typing for python 3.7
This is a regression from #79. When using python 3.7, I get an error saying
```
Extension error:
Could not import extension sphinx_immaterial (exception: cannot import name 'Literal' from 'typing' 
```

Locally, I use python 3.10, but my CI workflow previously used python 3.7 - I upgraded the CI to use python 3.9, and it runs fine now. I have also verified that this theme works fine with python 3.8

Since 3.7 is the next version to be deprecated, I'm ok with making this theme require 3.8+. Otherwise, we'd have to conditionally support the `python_transform_type_annotations_concise_literal` option based on the version of python used.

jbms:
This also adds a github actions build for Python 3.7, 3.8, and 3.10.

Fixes #84.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,10 @@ jobs:
         os:
           - 'ubuntu-latest'
         python-version:
+          - '3.7'
+          - '3.8'
           - '3.9'
+          - '3.10'
         node-version:
           - '16.x'
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


2bndy5 (approved):
If the dependencies keep growing, then it might help to cache the installed deps in the workflow. So far, I don't think the runner is taking a big performance hit when installing theme deps.

PR #90, opened 2022-05-02
Add mypy checking

jbms:


dev-requirements.txt
```diff
@@ -0,0 +1,6 @@
+wheel
+pylint
+black
+mypy
+types-PyYAML
+docutils-stubs
```
2bndy5, line 6: [typeshed has docutils stubs](https://pypi.org/project/types-docutils/) included. Not sure which is a better requirement though.

2bndy5: In breathe, they had to pin to a version that was consistent with sphinx requirements, but sphinx may have changed their reqs since then
```
types-docutils>=0.14,<0.18
```

sphinx_immaterial/apidoc_formatting.py
```diff
@@ -50,33 +52,29 @@ def visit_desc(self, node: sphinx.addnodes.desc) -> None:
...
         super().visit_desc(node)
 
-    def visit_desc_type(self, node: sphinx.addnodes.desc_type) -> None:
+    def visit_desc_type(self, node: docutils.nodes.Element) -> None:
         self.body.append(
             self.starttag(node, tagname="span", suffix="", CLASS="desctype")
         )
 
-    def depart_desc_type(self, node: sphinx.addnodes.desc_type) -> None:
+    def depart_desc_type(self, node: docutils.nodes.Element) -> None:
```
2bndy5, line 60: Its a shame we can't use the more specific type for specialized functions like these.

dev-requirements.txt
jbms: docutils-stubs is by the maintainer of sphinx and I think is more up to date than typeshed, but not sure

sphinx_immaterial/apidoc_formatting.py
jbms: Yeah.  I guess # type: ignore is the alternative but I'm not sure that is better

2bndy5: Could we do a `Union[docutils.nodes.Element, <specific node type>]` instead of `type: ignore`?

sphinx_immaterial/content_tabs.py
```diff
@@ -123,23 +122,26 @@ def depart_tab_label(self, node):
 
 def visit_tab_set(self: HTMLTranslator, node: content_tab_set):
     # increment tab set counter
-    self.tab_set_count = getattr(self, "tab_set_count", 0) + 1
+    self.tab_set_count = getattr(self, "tab_set_count", 0) + 1  # type: ignore
```
2bndy5, line 125: This will always be an `int`.
```suggestion
    self.tab_set_count: int = getattr(self, "tab_set_count", 0) + 1  # type: ignore
```

sphinx_immaterial/apidoc_formatting.py
jbms: I suppose, but technically that union type should be equivalent to just `docutils.nodes.Element` so it doesn't add anything.

sphinx_immaterial/content_tabs.py
```diff
@@ -123,23 +122,26 @@ def depart_tab_label(self, node):
 
 def visit_tab_set(self: HTMLTranslator, node: content_tab_set):
     # increment tab set counter
-    self.tab_set_count = getattr(self, "tab_set_count", 0) + 1
+    self.tab_set_count: int = getattr(self, "tab_set_count", 0) + 1  # type: ignore
```
2bndy5, line 125:
```suggestion
    self.tab_set_count: int = getattr(self, "tab_set_count", 0) + 1
```

sphinx_immaterial/content_tabs.py
```diff
@@ -123,23 +122,26 @@ def depart_tab_label(self, node):
...
 
     # configure tab set's div attributes
-    tab_set_identity = f"__tabbed_{self.tab_set_count}"
-    attributes = {"data-tabs": f"{self.tab_set_count}:{len(node.children)}"}
+    tab_set_identity = f"__tabbed_{self.tab_set_count}"  # type: ignore
+    attributes = {"data-tabs": f"{self.tab_set_count}:{len(node.children)}"}  # type: ignore
```
2bndy5, line 129: If it knows that self.tab_set_count is always an `int`, then we shouldn't need the `type: ignore` here. Unless `len(nodes.children)` is causing confusion in mypy.

sphinx_immaterial/content_tabs.py
2bndy5: damn you're fast. I noticed and fixed my suggestion after you commit it. 😆

sphinx_immaterial/content_tabs.py
jbms: The `type: ignore` was in regards to the attribute `tab_set_count`.  mypy can't handle adding properties to a class outside of its definition, I believe.

jbms: `# type: ignore` just suppresses all type checking errors on that line.  So in this case it isn't about the type of `tab_set_identity`, but about the access to `self.tab_set_count`.

jbms: I think we could narrow the scope of `# type: ignore` by moving the access of `self.tab_set_count` to a separate line.

sphinx_immaterial/apidoc_formatting.py
2bndy5: I think I understand the need for this change. running `mypy --show-error-codes` on main branch yielded
```
sphinx_immaterial\apidoc_formatting.py:58: error: Argument 1 of "depart_desc_type" is incompatible with supertype "HTML5Translator"; supertype defines the argument type as "Element"  [override]
sphinx_immaterial\apidoc_formatting.py:58: note: This violates the Liskov substitution principle
sphinx_immaterial\apidoc_formatting.py:58: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```
The error is misleading because it says argument 1 (self) should be what the super class' argument 2 (node) type is. Errors like these are why I tend to avoid using mypy. Remember that mypy is still in "beta" (v0.950 is latest), so we'll have to take the advice from mypy output with a inquisitive mindset.

We could start a mypy config in pyproject.toml that ignores certain errors for certain files.
```
[mypy]

[mypy-sphinx_immaterial.apidoc_formatting]
disable-error-code = "override"
```

2bndy5: It seems that mypy doesn't support ignoring certain errors per module. It can only be done globally, and I think it can only be done to 1 error code in the config file. It is starting to look like `type: ignore` is the only mechanism we can use to ignore an error, and I was hoping to avoid having that comment littered all over the src code.

The problem with `type: ignore` is that it is distracting and can cause violation of line length restrictions.

I still prefer to have the specific type in these specialized functions because it better describes some of the attributes implemented by a derivative of `docutils.nodes.Element`.

jbms: As far as the argument numbers, I think it is just excluding `self`.

I agree that mypy isn't great (the typescript type system is much, much better).  In putting together this PR I discovered a few incorrect type annotations but no actual bugs.  Nonetheless the type annotations themselves certainly are useful for documentation, and mypy is needed to ensure they bear some resemblance to reality.

I'm reluctant to disable error codes file-wide, but I'll go ahead and change the `# type: ignore` to `#type: ignore[code]` so that it is more selective and documents what is being ignored.

2bndy5: I'm looking into alternatives to mypy. I don't want to be tied down to a beta tool while this repo is still also in beta (too many bleeding edges).

I see that [microsoft has a tool called pyright](https://github.com/microsoft/pyright) that is available from npm. Since the type checking I've seen in VS Code (which uses pyright) is much more tolerant than mypy, I'm seriously considering the switch.

2bndy5: I'm sorry you did all this work already. I don't like shooting down PRs, but this is something I would've liked to discuss in an issue first.

jbms: There is `pytype`, which we use internally at Google.  However, it isn't really well supported for external use.  From what I've seen, mypy is pretty universal, and while it may be in "beta" it has been around for many years at this point.  However I haven't looked at `pywright` -- maybe it is a better alternative.

jbms: I'm happy to put this on hold.

2bndy5: I really like pyright in VS Code. There's few instances where I have to explicitly declare an object's type. I'm curious what IDE you use.

My main concern is we'd be losing specificity (in this instance) just to please the mypy tool. other errors like dynamic attribute definitions shouldn't require a `type: ignore` comment since it basically ignores an advantage to using python in general.

jbms: I use Emacs --- it supports the language server protocol, though for Python I don't use one.

Have you tried running pyright on this package?

Dynamic attributes are tricky, because if you just misspell an attribute you would want to receive an error.

The issue with the `visit_xxx` members is that technically the error is correct, it is just that the type annotations in Sphinx aren't as precise as we would like so we are trying to override them.  Still we could use a `# type: ignore` there rather than change the types.

2bndy5: > it is just that the type annotations in Sphinx aren't as precise as we would like

Agreed. I still don't want to extend that seemingly bad practice though (I'm sure sphinx devs have their reasons).

> Have you tried running pyright on this package?

I just ran it using `npx pyright .\sphinx_immaterial\*.py` and it found 25 source files for which it summarized:
```
Found 25 source files
... much better errors/warnings compared to mypy
242 errors, 1 warning, 0 informations
Completed in 3.353sec
```
It also made notice of the lack of configuration (in pyproject.toml) before file discovery.

jbms:
I removed `object_pairs_hook` parameter because it wasn't needed.  The other refactoring was related to mypy --- mypy does not support the same variable name being assigned different types, e.g. in one block it being used with one type, then later the same name being used independently with a different type.  Therefore I split up one of the functions into a number of smaller functions.

2bndy5:
oh I see; I just type ignored that param because I'm not sure how to tell pyright that it is a class (not an instance of a class).

I did some type aliasing to cpp_domain_fixes.py and found that pyright doesn't like using a module's namespace (`sphinx.domains.c`/`sphinx.domains.cpp`) as an arg to a function. I haven't checked with the PEPs that pyright adheres to though.

jbms:
> oh I see; I just type ignored that param because I'm not sure how to tell pyright that it is a class (not an instance of a class).

I'm not sure which parameter you're referring to, but `typing.Type[X]` is how you specify a parameter that is a type itself that inherits from `X`, rather than an instance of `X`.

2bndy5:
https://github.com/jbms/sphinx-immaterial/blob/bd1e3595e6e114a3bb6c20369112e1533542bce1/sphinx_immaterial/cpp_domain_fixes.py#L715

I've been going through and adding types to all function params. I have pyright to emit this violation as a warning so builds still pass, but it helps me understand the focus of a function without having to backtrack what was passed to it (in trying to understand what the function is doing with the param).

jbms:
> https://github.com/jbms/sphinx-immaterial/blob/bd1e3595e6e114a3bb6c20369112e1533542bce1/sphinx_immaterial/cpp_domain_fixes.py#L715
> 
> I've been going through and adding types to all function params. I have pyright to emit this violation as a warning so builds still pass, but it helps me understand the focus of a function without having to backtrack what was passed to it (in trying to understand what the function is doing with the param).

I think the correct type for that would be: `typing_extensions.Literal[sphinx.domains.c, sphinx.domains.cpp]`

Not clear if it will handle that, though.

jbms:
Nevermind, I see that that `Literal` is only supposed to contain a limited set of possible values:
https://peps.python.org/pep-0586/#illegal-parameters-for-literal-at-type-check-time

Maybe `types.ModuleType` (which I don't think can be used like `ModuleType[sphinx.domains.c]`) is the best we can do.

2bndy5:
tried that. pyright says that 
>Type arguments for "Literal" must be None, a literal value (int, bool, str, or bytes), or an enum value

also tried `Union[sphinx.domains.c, sphinx.domains.cpp]` which lead to 
>Module cannot be used as a type

In fact, the CI log errored out when I combined `Union[sphinx.domains.c, sphinx.domains.cpp], # type: ignore`; something about the incorrect args given to `Union`...

There really isn't many `type: ignore` comments added, so I'm ok with just using
https://github.com/jbms/sphinx-immaterial/blob/55d187c4388f248405f5c6d2f051d0883eaf2954/sphinx_immaterial/cpp_domain_fixes.py#L731
because the parameter is well named enough

2bndy5:
Hey! `types.ModuleType` worked. Pyright is ok with using
```py
    domain_module: ModuleType,  # Union[sphinx.domains.c, sphinx.domains.cpp]
```
I put the `Union` of the specific modules for precision.

dev-requirements.txt
2bndy5: After a closer look at typeshed repo, I doubt we can expect any updates to `types-docutils>=0.14,<0.18` stubs. It probably better to use the one maintained for/by sphinx devs

sphinx_immaterial/content_tabs.py
2bndy5: I'm working locally on this branch, and I found that using the following doesn't require the type ignore comments
```py
    tab_set_count: int = getattr(self, "tab_set_count", 0) + 1
    setattr(self, "tab_set_count", tab_set_count)

    # configure tab set's div attributes
    tab_set_identity = f"__tabbed_{tab_set_count}"
    attributes = {"data-tabs": f"{tab_set_count}:{len(node.children)}"}
```

---
I'm also going through the other changes on this branch... I'll push when ready.

2bndy5: I went forward with the `Union[base_t, derived_t]` because the type **hinting** isn't just programmatic for me (it helps during code review).

sphinx_immaterial/apidoc_formatting.py
2bndy5: >> Could we do a `Union[docutils.nodes.Element, <specific node type>]` instead of type: ignore?
>
> I suppose, but technically that union type should be equivalent to just docutils.nodes.Element so it doesn't add anything.

I went forward with this idea because type **hinting** isn't just programmatic when reviewing (since the functions names are subject to authors' thought process).

resolving this long thread.

2bndy5:
I made some changes:
1. use custom stubs for `pymdownx.keymap_db` and `sphinxcontrib.details.directive` modules (they now exist in _typings_py_ folder
   - turns out typeshed has stubs for `jsonschema` pkg 👍🏼 
2. add mypy config to pyproject.toml (use pretty/verbose error ouput and point to custom stubs)
4. removed some `type: ignore` comments by using `getattr()` and `setattr()`

2bndy5 (approved):
ready to merge when you are

PR #107, opened 2022-05-28
Fix hamburger menu and fix local/global toc split

jbms:
This fixes the hamburger menu issue identified by @2bndy5  in #99.

Unfortunately, this fix does not work for the case where a Sphinx page has multiple top-level sections (rather than a single "title").  The design of the hamburger menu in the mkdocs-material theme does not seem to be able to accommodate that possibility.


2bndy5 (approved):
I didn't realize my concern was theme specific (thought it was specific to the custom python auto-gen ext).

Issue #112, opened 2022-06-21
C++ domain parameter names not conventionally required


2bndy5:
Its rather common to document a header that has no parameter names (only datatypes) when the parameter names only exist in the implementation file.

```cpp
// in demo.h
/**
 * @param a info about `a`
 * @param b info about `b`
 */
void func(int, char);

// in demo.cpp
void func(int a, char b) { /* do something with params */ }
```

However, this theme expects the declaration signature to have the parameter names.

Furthermore, the warning message for this situation isn't very helpful:
```
/path/to/breathe/documentation/source/doxygen.rst::Parameter name 'b' does not match any of the parameters defined in the signature: []
```

## Ideas
1. It would be nice to have a workaround that allowed bypassing this warning because it is theme related, rather than solely an author problem.
2. It would be nicer to show the originating signature for which the parameter name is being sought. I was able to hack this by changing https://github.com/jbms/sphinx-immaterial/blob/246a5c983708a24a0247cf5d6747c13c15331246/sphinx_immaterial/cpp_domain_fixes.py#L750-L756 to
   ```py
               full_sig = (
                   param_node.parent.parent.parent.parent.parent.parent.parent.astext()
               )
               full_sig = full_sig[: full_sig.find("\n\n")]
               logger.warning(
                   "Parameter name %r does not match any of the parameters "
                   "defined in the signature:\n%s",
                   param_name,
                   full_sig,
                   location=param_node,
               )
   ```
   resulting in a warning that reads:
   ```
   /path/to/breathe/documentation/source/doxygen.rst::Parameter name 'b' does not match any of the parameters defined in the signature:
   void func(int, char);
   ```
   I'm sure there's a cleaner way of doing this.

jbms:
The lack of line numbers is due to https://github.com/sphinx-doc/sphinx/pull/10249

With line numbers the user would at least have an easy route to figure out which signature the warning relates to.

To me it seems like the warning is kind of useful in this case, though --- wouldn't it be better to include the parameter names, so that they can be cross-linked?  And also then the user would more easily know which parameter is being described by the `@param` comments.

In any case, sphinx does have the `suppress_warnings` option:
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings

I think we would need to do something to give this warning a recognizable warning type so that it could be targeted by that option.

We could then provide an example in the documentation of how to suppress this warning.

2bndy5:
[...]

PR #114, opened 2022-06-30
add cross_link_parameter_description option

2bndy5:
solves #112 

I also added a collapsed list of the supported `object_description_options` patterns in the _api.rst_ for completeness (helpful for undocumented object types).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -135,6 +135,26 @@ The following options can be customized for each object type using
              ("py:.*", dict(include_fields_in_toc=False)),
          ]
 
+.. objconf:: cross_link_parameter_descriptions
+
+   This option can be used to prevent the cross linking of parameter descriptions to their
+   declarations in the a function signature. This was specifically provided for the C/C++
+   domains because omitting parameter names in C/C++ prototypes is conventionally allowed.
+   However, it can also be disabled in the python domain when using non-pythonic optional
+   syntax as seen in our `format_exception()` example.
+
+   .. admonition:: Example
+      :class: example
+
+      By default, this option is enabled (`True`), but it can be disabled by setting this
+      option to `False`.
+
+      .. code-block:: python
+
+         object_description_options = [
+            ("cpp:function", dict(cross_link_parameter_descriptions=False)),
+         ]
+
 Other options described elsewhere include:
 
 - :objconf:`wrap_signatures_with_css`
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -146,7 +146,7 @@ The following options can be customized for each object type using
 .. objconf:: cross_link_parameter_descriptions
 
    This option can be used to prevent the cross linking of parameter descriptions to their
-   declarations in the a function signature. This was specifically provided for the C/C++
+   declarations in the a signature. This was specifically provided for the C/C++
    domains because omitting parameter names in C/C++ prototypes is conventionally allowed.
    However, it can also be disabled in the python domain when using non-pythonic optional
    syntax as seen in our `format_exception()` example.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -15,6 +15,14 @@ domain/object type pair.
    initializing each option to its default value, and then applying as overrides
    the options associated with each matching pattern in order.
 
+   .. details:: Here's the list of supported patterns (and their default configuration)
+      :class: faq
+
+      .. literalinclude:: ../sphinx_immaterial/apidoc_formatting.py
+         :language: python
+         :start-at: DEFAULT_OBJECT_DESCRIPTION_OPTIONS
+         :end-before: # end DEFAULT_OBJECT_DESCRIPTION_OPTIONS
+
 .. _object-description-options:
 
 Object description options
--- a/sphinx_immaterial/cpp_domain_fixes.py
+++ b/sphinx_immaterial/cpp_domain_fixes.py
@@ -749,9 +749,9 @@ def _add_parameter_documentation_ids(
                 all_params.update(sig_param_nodes)
             logger.warning(
                 "Parameter name %r does not match any of the parameters "
-                "defined in the signature: %r",
+                "defined in the signature(s): %r",
                 param_name,
-                obj_content.parent.children[0].astext(),
+                [n.astext() for n in obj_content.parent.children[:-1]],
                 location=param_node,
             )
             return
--- a/sphinx_immaterial/cpp_domain_fixes.py
+++ b/sphinx_immaterial/cpp_domain_fixes.py
@@ -751,7 +751,7 @@ def _add_parameter_documentation_ids(
                 "Parameter name %r does not match any of the parameters "
                 "defined in the signature: %r",
                 param_name,
-                list(all_params.keys()),
+                obj_content.parent.children[0].astext(),
                 location=param_node,
             )
             return
@@ -975,16 +975,19 @@ def _monkey_patch_domain_to_cross_link_parameters_and_add_synopses(
         while symbols[-1].siblingAbove:
             symbols.append(symbols[-1].siblingAbove)
         symbols.reverse()
-        _cross_link_parameters(
-            app=self.env.app,
-            domain=domain,
-            content=getattr(self, "contentnode"),
-            symbols=symbols,
-        )
 
         options = apidoc_formatting.get_object_description_options(
             self.env, self.domain, self.objtype
         )
+
+        if options["cross_link_parameter_descriptions"]:
+            _cross_link_parameters(
+                app=self.env.app,
+                domain=domain,
+                content=getattr(self, "contentnode"),
+                symbols=symbols,
+            )
+
         generate_synopses = options["generate_synopses"]
 
         if generate_synopses is not None:
--- a/sphinx_immaterial/cpp_domain_fixes.py
+++ b/sphinx_immaterial/cpp_domain_fixes.py
@@ -749,9 +749,10 @@ def _add_parameter_documentation_ids(
                 all_params.update(sig_param_nodes)
             logger.warning(
                 "Parameter name %r does not match any of the parameters "
-                "defined in the signature(s): %r",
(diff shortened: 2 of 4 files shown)


sphinx_immaterial/cpp_domain_fixes.py
```diff
@@ -751,7 +751,7 @@ def cross_link_single_parameter(
                 "Parameter name %r does not match any of the parameters "
                 "defined in the signature: %r",
                 param_name,
-                list(all_params.keys()),
+                obj_content.parent.children[0].astext(),
```
jbms, line 754: There can be more than one signature, so it would be better to include all of them.

The full list of signatures can be obtained from `obj_content.parent.children[:-1]`  --- the last child is always `obj_content`.

2bndy5: I did that and it returned the documentation nodes for the node as well. Also, `astext()` is specific to a node, not a list of nodes.

docs/api.rst
```diff
@@ -135,6 +143,26 @@ The following options can be customized for each object type using
              ("py:.*", dict(include_fields_in_toc=False)),
          ]
 
+.. objconf:: cross_link_parameter_descriptions
+
+   This option can be used to prevent the cross linking of parameter descriptions to their
+   declarations in the a function signature. This was specifically provided for the C/C++
```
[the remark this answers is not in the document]
2bndy5: good catch!

sphinx_immaterial/cpp_domain_fixes.py
jbms: You are saying that even with excluding the last child, of `obj_content.parent`, you still got the documentation?

In the case that more than one signature is specified, the valid parameter name set is the union of parameter names from every signature, so just displaying the first signature may be confusing.

I would suggest changing to:

```python
            logger.warning(
                "Parameter name %r does not match any of the parameters %r "
                "defined in the signatures %r",
                param_name,
                list(all_params.keys()),
                [signode.astext() for signode in signodes],
                location=param_node,
            )
```

For simplicity, you could just add `signodes` as an additional parameter to `_add_parameter_documentation_ids`.

2bndy5: using
```py
[n.astext() for n in obj_content.parent.children[-1:]]
```
The warning we've been getting (about the optional syntax) looks like
```
sphinx-immaterial\docs\additional_samples.rst:234: WARNING: Parameter name 'limit' does not match any of the parameters defined in the signature: ['Format the exception with a traceback.\n\nParameters\n\netype\n\nexception type\n\nvalue\n\nexception value\n\ntb\n\ntraceback object\n\nlimit : integer or None\n\nmaximum number of stack frames to show\n\nReturn type\n\nlist of strings']
```

docs/api.rst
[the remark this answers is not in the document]
jbms: Typo still present: "in the a signature" should be "in the signature"

sphinx_immaterial/cpp_domain_fixes.py
jbms: `obj_content.parent.children[-1:]` means the list containing only the last child --- we want `obj_content.parent.children[:-1]` which means the list containing all but the last child.

2bndy5: opps. I didn't realize I put the colon in the wrong place. Uding
```py
            logger.warning(
                "Parameter name %r does not match any of the parameters "
                "defined in the signature(s): %r",
                param_name,
                [n.astext() for n in obj_content.parent.children[:-1]],
                location=param_node,
            )
```
produced the desired output:
```
sphinx-immaterial\docs\additional_samples.rst:234: WARNING: Parameter name 'limit' does not match any of the parameters defined in the signature: ['format_exception(etype, value, tb, [limit=None])']
```

I'm going to test this with the C++ domain and commit if the results are consistent.

2bndy5: resolved

jbms: I still think it would be helpful to include the list of allowed parameters as well --- for example, in the above case, it may not be obvious that we cannot properly parse `limit`.

sphinx_immaterial/python_domain_fixes.py
```diff
@@ -545,6 +545,15 @@ def cross_link_single_parameter(
...
 
+    func_options = apidoc_formatting.get_object_description_options(
+        env, "py", "function"
+    )
+    if (
+        not func_options["cross_link_parameter_descriptions"]
+        or not param_options["cross_link_parameter_descriptions"]
+    ):
+        return
+
```
jbms, line 556: It might make more sense to just skip the call to `_cross_link_parameters` altogether.  I think the only difference is that as you have it now, even if cross linking is disabled, we still call `_add_parameter_links_to_signature`.

Also, I would say that this should just be checking the object description options for the actual object type being defined here (which may be e.g. `py:method` rather than `py:function`), and should not check the object description options for `py:parameter`.

2bndy5: 1. I wanted to minimize the changes while #99 is open.
2. I think the synopses are closely tied to the cross-linking mechanism. This made it hard for me to discern where would a good place for this.

At first, I had used `self.objtype` to get the options in  https://github.com/jbms/sphinx-immaterial/blob/1062bc88e22bbb731b7cd437e358ef63fd4b22d0/sphinx_immaterial/python_domain_fixes.py#L659-L660 But that didn't cover the `py:parameter` obj type (which for our docs is solely triggered by the autodoc demo).

jbms: To me it seems more natural that the relevant object type would be the function/method object type (or in the case of C++, any object type).  `py:parameter` does not itself have parameters.  If the user wants to disable cross linking for all python function/method types, they can just use `"py:.*"` as the pattern.

I don't think there should be significant conflicts with #99, since that doesn't make too many changes to `python_domain_fixes.py`.  If there are any conflicts, I can fix them in my PR after this is submitted.

2bndy5: If I do insert an condition that prevents `_cross_link_parameter()` being called, then would the docs about the `python_qualify_parameter_ids` config need to mention this option as an additional circumstance?

I'm ok with excluding `py:parameter` objects, but it does feel like the option name implies that they are affected.

sphinx_immaterial/cpp_domain_fixes.py
2bndy5: With these requested changes, the parameters are listed as provided to the signature. The whole reason for changing this is because the C/C++ domain can have an unhelpful empty list of parameters.

2bndy5: Ok, I've added a list of "known parameters" to the warnings, which uses the original approach.

jbms: Thanks

sphinx_immaterial/python_domain_fixes.py
jbms: I actually removed `python_qualify_parameter_ids` in #99, and replaced it with a different mechanism, so I think we can ignore its documentation.

I assumed the intent of the `cross_link_parameter_descriptions` option is to basically disable the special handling of parameters by this theme --- we no longer create `py:parameter` "objects" corresponding to parameters, they don't show up in search results or in the TOC.  But maybe that isn't the behavior you are looking for --- maybe you want to still create the `py:parameter` objects, but just don't link them to the signature.

2bndy5: > we no longer create py:parameter "objects" corresponding to parameters

Is this related to #99 changes also? 

>  But maybe that isn't the behavior you are looking for 

I only want to allow the prevention of linking of parameters to descriptions. I honestly don't care if we create a `py:parameter` node/object or not (its an implementation detail - I'm focused on user experience).

As long as any other config options aren't negatively affected by this option, then I have no problem preventing the `_cross_link_parameter()` being called.

jbms: The main question is --- do you want the parameters to show up in the TOC and search results (as their own entity, rather than just text within the object description), or not?

2bndy5: I like the ability to cross-reference a specific parameter, but I'm not concerned with search results or ToC (especially if the description fields are going into the ToC). I could forego the cross-ref-ing ability since that has been the norm for so long.

To be honest, I've been disabling the parameters inclusion in the ToC for any of my projects that use this theme.

jbms: Perhaps there should be some finer-grained options rather than just this one option to disable the generation of `parameter` objects entirely.  E.g.:

object_description_options:
- `generate_parameter_objects: bool = True`  --- enables generation of parameter objects from `:param xx:` fields in the object description; these parameter objects can be cross-referenced (with synopsis shown depending on `generate_synopsis`), appear in search results, and can be shown in the TOC (depending on `include_in_toc`).
- `cross_link_signature_parameters: bool = True` --- cross-links parameters defined in signatures to the parameter objects defined by `generate_parameter_objects`.  Has no effect if `generate_parameter_objects` is `False`.
- `nitpick_parameter_names: bool = True` --- emits a warning if a parameter documented with `:param:` can't be matched to a parameter defined by a signature.  Has no effect if either `cross_link_signature_parameters` or `generate_parameter_objects` is `False`.

2bndy5: I'm just working what is currently there. I'm ok with closing this PR and continue waiting for #99 to be merged, then re-addressing #112 with the improved code-base.

For now, I have to use this patch for developing the clang-autodoc approach because some of my test code (which was taken from breathe's example cases) will trigger the problem stated in #112 .

2bndy5:
Closing this as the repo structure is changing with #99 and the recommended way to satisfy #112 is with the addition of more than a single option.

PR #117, opened 2022-07-07
Add C++ autosummary extension

### LINKED ISSUE #92: Incorporation of TensorStore C++ autosummary extension
I have a TensorStore C++ autosummary also mostly implemented, although not yet released in the TensorStore repository.  I would like to potentially incorporate it into this theme, but wanted to find out if that might be useful to anyone else.  (If not, I'll just maintain it separately within TensorStore.)

It works similarly to the TensorStore Python autosummary extension, in that every entity is documented on a separate page.

At the top level, entities are organized into groups (like doxygen groups) and e.g. a `.. cpp-apigen:: group` directive is used to insert the summary of the members of a group.

jbms:
Fixes #92.

This is based on #116.

TODO:

- [ ] Document how to use separate build step to preprocess C++ input header using normal compiler
- [ ] Document how to extract compiler flags (especially include paths) from normal build process
- [ ] Add unit tests of api_parser
- [ ] Add unit tests of apigen
- [ ] Add example of `explicit(bool)` auto-conversion
- [ ] Support `:order:` option as in Python apigen
- [ ] Support equivalent of `python_apigen_default_groups` / `python_apigen_default_order`
- [ ] Support rST-style `:group:` and `:relates:` in addition to doxygen-style commands
- [ ] Support `@`-prefixed doxygen commands in addition to `\`-prefixed ones (e.g. `@param` in addition to `\param`)
- [ ] Add example of documenting multiple signatures together (by leaving no space between them)
- [ ] Add tests of source location (e.g. in signature, and in docstring)
- [ ] Add tests and documentation of NONITPICK


jbms:
This is somewhat working, but perhaps we can work together to refine it.

2bndy5:
> documenting multiple signatures together (by leaving no space between them)

Is this related to overloads sharing the same docstring?

jbms:
> > documenting multiple signatures together (by leaving no space between them)
> 
> Is this related to overloads sharing the same docstring?

Yes --- the result is a single `cpp:function` with multiple signatures.

I added an example of this: `Array::operator[]`

jbms:
This is now updated to use `types-clang` rather than ones I generated myself.  There are a few bugs in `types-clang` that required a `cast` as a workaround: https://github.com/tgockel/types-clang/pull/2

2bndy5:
I also noticed that `CursorKind.value` isn't typed at all in that stub lib...

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -0,0 +1,2217 @@
...
+    `.disallow_macros`, are documented.
+    """
+
+    disallow_macros: List[Pattern] = dataclasses.field(default_factory=list)
+    """List of regular expressions matching *disallowed* macro names.
+
+    Macros matching any of these patterns are undocumented.
+    """
+
+    ignore_diagnostics: List[Pattern] = dataclasses.field(default_factory=list)
```
2bndy5, line 209: I think this could be made an arbitrary bool. In my research, I had a conf.py option to output these diags to the build log (defaulting to false) using the severity of the diag as a logging level. If users really want to use static analysis, then they can use clang-tidy separately.

2bndy5 (commented):
First impressions.

docs/apidoc/cpp/apigen.rst
```diff
@@ -0,0 +1,148 @@
...
+      of the current page/section.  Specifying this flag disables that behavior.
+
+   .. rst-example:: Example usage
+
+      .. cpp-apigen-group:: indexing
+         :notoc:
+
+.. rst:directive:: .. cpp-apigen-entity-summary:: entity-name
+
+   Generates a summary of a single Python entity.
```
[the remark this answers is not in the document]
jbms: Yeah this example still needs to be fixed.

Also at the moment the entity is specified by the clang usr-derived id, but it should probably be something else friendlier.

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -0,0 +1,2217 @@
...
+
+        self.input_source = input_content
+
+        self.index = clang.cindex.Index.create()
+        start_time = time.time()
+        self.tu = self.index.parse(
+            input_path,
+            unsaved_files=[(input_path, input_content)],
+            args=tuple(config.compiler_flags) + ("-ferror-limit=0",),
+            options=(  # TranslationUnit.PARSE_SKIP_FUNCTION_BODIES +
```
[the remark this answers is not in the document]
jbms: Should add a comment.  If function bodies are skipped, the extent does not include the body.  That interferes with detecting if there are multiple entities defined with no space between (so that they will all be documented together).

docs/apidoc/cpp/apigen.rst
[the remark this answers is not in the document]
2bndy5: I couldn't find any documentation about the USR generation, so I often look to the [src code](https://github.com/llvm/llvm-project/blob/main/clang/lib/Index/USRGeneration.cpp). Some USR ids use the line number, but IIRC that is only applied to locally scoped members (like param decls and vars).

sphinx_immaterial/apidoc/cpp/api_parser.py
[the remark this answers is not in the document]
2bndy5: Oh ok. Yeah, a comment would've satisfied that inquiry.

sphinx_immaterial/apidoc/cpp/api_parser.py
jbms: I have found that any compiler errors are usually an indication that the result of parsing will be wrong, sometimes in subtle ways, and should normally not be ignored.  However, for tensorstore I ran into some errors related to an undefined __builtin in a standard library header that I was not able to avoid, and did not seem to cause any problems, so I added this mechanism to suppress them.

2bndy5: I know that clang will fail fast if it can't find an included lib, thus no parsing at all. That's why I mentioned the compilation database path (`-p` I think) can be used when third party libs are involved.

I haven't yet looked at how you're normalizing the path relative to the conf.py path, but that might also be a factor when passing certain args to clang.

2bndy5:
Going forward, can we change NONITPICK to NO_NIT_PICK? I keep reading it as NON_IT_PICK 🤣

sphinx_immaterial/apidoc/cpp/api_parser.py
jbms: Paths are normalized using `include_directory_map`.  We should probably try to support compilation databases, but one tricky thing about a compilation database is that it doesn't include the compiler built-in include paths.   An alternative is to have the user preprocess the desired headers first using their normal compiler.  The original file/line information is preserved by the added `#line` directives.

2bndy5: I think users can specify [`-working-directory`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-working-directory-arg) to clang. Or, maybe we should provide an option to do it.

jbms:
> Going forward, can we change NONITPICK to NO_NIT_PICK? I keep reading it as NON_IT_PICK rofl

Makes sense --- or could be: nowarnxref or something --- nitpick is a sphinx-specific term that may be confusing to readers of the C++ source code.

2bndy5:
ok, I rebased this branch on latest main, and the CI is failing because it now tests for Windows and the fix for that is in my other branch (based on this one).

jbms:
I wonder if we should merge this as an experimental feature, since this PR also contains the default-literal-role and highlight changes that also apply to Python and JSON.

2bndy5:
We could do that. I've been think on ways to refactor this ext. My priorities are 
1. better parsing mechanism
2. specify an entity via user friendly implementation

The database and diagnostics are a lesser priority for me. And I assume unit testing will just develop organically.

jbms:
To be clear, by better parsing mechanism, is that in regards to the doxygen syntax, or in regards to the C++ source itself?

For specifying an entity, we could allow it to be specified by its "object name", e.g. `foo::bar::Baz` or `foo::bar::Baz[overload_id]` if overloaded.  Much more reasonable than the clang USR.  Allowing it to be specified by signature, e.g. as supported by the C++ domain for cross references, would be significantly more challenging I think.

2bndy5:
> To be clear, by better parsing mechanism, is that in regards to the doxygen syntax

Yes. I'm not sure how the C++ parsing would need improvement. I still want to provide a config option to turn off the doxygen syntax parsing, so the docstrings can be written in pure RST.

> we could allow it to be specified by its "object name", e.g. foo::bar::Baz or foo::bar::Baz[overload_id] if overloaded

This was my thinking as well, but it would be more natural to use the parameter datatypes types for overloads (instead of a docstring specific ID). I find the ability to document multiple overloads with a single docstring very appealing - not even doxygen is capable of that.

jbms:
> > To be clear, by better parsing mechanism, is that in regards to the doxygen syntax
> 
> Yes. I'm not sure how the C++ parsing would need improvement. I still want to provide a config option to turn off the doxygen syntax parsing, so the docstrings can be written in pure RST.

Sounds good.

> 
> > we could allow it to be specified by its "object name", e.g. foo::bar::Baz or foo::bar::Baz[overload_id] if overloaded
> 
> This was my thinking as well, but it would be more natural to use the parameter datatypes types for overloads (instead of a docstring specific ID). I find the ability to document multiple overloads with a single docstring very appealing - not even doxygen is capable of that.

That could be done using the C++ domain symbol resolution used for cross-references, but is a bit tricky because currently the symbol tree only gets populated as the c++ domain directives are actually run, which is too late.  We would need to somehow create a separate one and pre-populate it.  I also think it is often inconvenient to specify the full signature if you have a lot of template or function parameters.

2bndy5:
> currently the symbol tree only gets populated as the c++ domain directives are actually run

I noticed that. Is there a reason you don't just save the parsed API to a builder env var?

jbms:
There wasn't a need for the symbol tree except for normal xrefs so I just tried to keep it working closely to how it normally does.

Note that the c++ domain symbol lookup is slow if you have a lot of symbols in the same namespace because it just does a linear search over every name --- it would be better if it used a dict.

2bndy5:
Well I'm ok with this getting merged but there should be a warning on both config and demo doc pages. Possibly even a warning in the build log like you did for graphviz, but that seems a little excessive to me -- you went above and beyond my expectations there (again).

2bndy5:
Instead of a task list in the thread, you could use a [github project](https://docs.github.com/en/issues/planning-and-tracking-with-projects/learning-about-projects/about-projects) to track the progress of this extension.

jbms:
I added a warning that it is experimental.

2bndy5 (dismissed):

2bndy5:
Did I mess up the rebase?

jbms:
> Did I mess up the rebase?

Not sure what happened with `types-clang`

2bndy5:
I thought it was weird when the rebased push didn't trigger the CI (only RTD's CI ran).

setup.py
```diff
@@ -222,6 +222,7 @@ def run(self):
         "jsonschema_validation": ["jsonschema"],
         "clang-format": ["clang-format"],
         "keys": ["pymdown-extensions"],
+        "libclang": ["libclang"],
```
2bndy5, line 225: shouldn't this be specific to the domain (like `pyyaml` is to `[json]`)?

jbms: Done.

2bndy5 (approved):

PR #119, opened 2022-07-10
Some cpp comment stripping parsing

2bndy5:
Includes
- fix for Windows paths
- fix for unresolved cross-refs in the new docs
- support for `\` and `@` prefixed Doxygen commands
- support for `@retval` command
- support for param direction(s)
- add new `strip_comment()` to accommodate for all forms of C++ comment syntax

I also enabled verbosity in the cpp.apigen demo and added a line that shows the number of declarations that will be parsed.

### DIFF
--- a/docs/cpp_apigen_demo/index_interval.h
+++ b/docs/cpp_apigen_demo/index_interval.h
@@ -21,7 +21,7 @@ class IndexInterval {
 
   /// Returns the lower bound.
   ///
-  //! @retval int The lower bound.
+  //! @returns The lower bound.
   int lower() const;
 
   /// Returns the upper bound.
--- a/docs/cpp_apigen_demo/index_interval.h
+++ b/docs/cpp_apigen_demo/index_interval.h
@@ -20,6 +20,8 @@ class IndexInterval {
   friend std::ostream& operator<<(std::ostream& os, IndexInterval x);
 
   /// Returns the lower bound.
+  ///
+  //! @retval int The lower bound.
   int lower() const;
 
   /// Returns the upper bound.
--- a/sphinx_immaterial/apidoc/cpp/api_parser.py
+++ b/sphinx_immaterial/apidoc/cpp/api_parser.py
@@ -54,6 +54,7 @@ from typing import (
     ClassVar,
     Pattern,
 )
+from textwrap import dedent
 
 import ctypes
 
@@ -109,7 +110,7 @@ TEMPLATE_PARAMETER_ENABLE_IF_NON_TYPE_PATTERN = re.compile(
 )
 
 SPECIAL_GROUP_COMMAND_PATTERN = re.compile(
-    r"^\\(ingroup|relates|membergroup|id)\s+(.+[^\s])\s*$", re.MULTILINE
+    r"^(?:\\|@)(ingroup|relates|membergroup|id)\s+(.+[^\s])\s*$", re.MULTILINE
 )
 
 
@@ -430,6 +431,38 @@ def _get_all_decls(config: Config, cursor: Cursor, allow_file):
         if kind in CLASS_KINDS:
             yield from _get_all_decls(config, child, None)
 
+def strip_comment(cmt: str = None) -> Optional[List[str]]:
+    """Strip the raw string of an object's comment into lines.
+    :param cmt: the comment to parse.
+    :returns: A list of the lines without the surrounding C++ comment syntax.
+    """
+    if cmt is None:
+        return None
+    # the first line is never indented (in clang-parsed form)
+    # so dedent all subsequent lines only
+    first_line_end = cmt.find("\n")
+    cmt = cmt[:first_line_end] + dedent(cmt[first_line_end:])
+    # split into a list of lines & account for CRLF and LF line endings
+    body = [line.rstrip("\r") for line in cmt.splitlines()]
+
+    # strip all the comment syntax out
+    if body[0].startswith("//"):
+        body = [line.lstrip("/").lstrip("!").lstrip("<") for line in body]
+    elif body[0].startswith("/*"):
+        body[0] = body[0].lstrip("/").lstrip("*").lstrip("!")
+        multi_lined_asterisk = True  # works also for single-line comments blocks
+        if len(body) > 1:
+            line_has_asterisk = [line.startswith("*") for line in body[1:]]
+            multi_lined_asterisk = line_has_asterisk.count(True) == len(body) - 1
+        body = [
+            (line.lstrip("*").lstrip("<") if multi_lined_asterisk else line)
+            for line in body
+        ]
+        body[-1] = body[-1].rstrip("*/").rstrip()
+    body = dedent("\n".join(body)).splitlines()
+    return body
+
+DOC_COMMENT_PREFIX = re.compile(r"/(//|/\!|\*\*|\*\!)\<?")
 
 def get_doc_comment(config: Config, cursor: Cursor):
     translation_unit = cursor.translation_unit
@@ -463,7 +496,7 @@ def get_doc_comment(config: Config, cursor: Cursor):
                 stop = True
                 break
             spelling = token.spelling
-            if not spelling.startswith("///"):
+            if DOC_COMMENT_PREFIX.match(spelling) is None:
                 stop = True
                 break
             (
@@ -478,10 +511,8 @@ def get_doc_comment(config: Config, cursor: Cursor):
                 stop = True
                 break
             presumed_line = new_presumed_line
-            if spelling.startswith("/// "):
-                comment_lines.append(spelling[4:])
-            else:
-                comment_lines.append(spelling[3:])
+            if DOC_COMMENT_PREFIX.match(spelling) is not None:
+                comment_lines.extend(strip_comment(spelling))
             got_line = True
             break
         if stop or not got_line:
@@ -1661,19 +1692,19 @@ def _get_default_member_group(entity: CppApiEntity) -> str:
 
 
 def _normalize_doc_text(text: str) -> str:
-    text = re.sub(r"^\\brief\s", "", text)
+    text = re.sub(r"^((?:\\|@)(?:brief|details)\s+)", "", text)
     text = re.sub(
-        r"^\\(t?param) ([a-zA-Z_][^ ]*)", ":\\1 \\2:", text, flags=re.MULTILINE
+        r"^(?:\\|@)(t?param)(\[(?:in|out|in,\sout)\])?\s+([a-zA-Z_][^ ]*)", ":\\1 \\3\\2:", text, flags=re.MULTILINE
     )
-    text = re.sub(r"^\\(error)\s+`([^`]+)`", ":\\1 \\2:", text, flags=re.MULTILINE)
+    text = re.sub(r"^(?:\\|@)(error)\s+`([^`]+)`", ":\\1 \\2:", text, flags=re.MULTILINE)
     text = re.sub(
-        r"^\\(returns?|pre|post|checks|dchecks|schecks|invariant|requires)(?: |\n )",
+        r"^(?:\\|@)(returns?|pre|post|[ds]?checks|invariant|requires)(?: |\n )",
         ":\\1: ",
         text,
         flags=re.MULTILINE,
     )
+    text = re.sub(r"^(?:\\|@)(retval)\s+(\S+)", ":\\1 \\2:", text, flags=re.MULTILINE)
     text = SPECIAL_GROUP_COMMAND_PATTERN.sub("", text)
-    text = text.lstrip(" ")
     return text
 
 
--- a/sphinx_immaterial/apidoc/cpp/api_parser.py
+++ b/sphinx_immaterial/apidoc/cpp/api_parser.py
@@ -124,16 +124,16 @@ class Config:
     input_path: str = "__input.cpp"
     """Path to the input file to parse.
 
-    This may either be a path to an existing file, or :py:attr:`input_content` may
+    This may either be a path to an existing file, or `.input_content` may
     specify its content, in which case the filesystem is not accessed.
 
-    If :py:attr:`input_content` is specified and merely contains :cpp:`#include`
+    If `.input_content` is specified and merely contains :cpp:`#include`
     directives, then the actual path does not matter and may be left as the
     default value.
     """
 
     input_content: Optional[bytes] = None
-    """Specifies the content of :py:attr:`input_path`.
+    """Specifies the content of `.input_path`.
 
     If unspecified, the content is read from filesystem.
     """
@@ -165,7 +165,7 @@ class Config:
     defined in the standard library and third-party libraries.
 
     .. important::
-        When building on Windows, all path separators are normalized to :python:`"/"`. 
+        When building on Windows, all path separators are normalized to :python:`"/"`.
         Therefore, in the specified regular expressions, always use :python:`"/"` to
         match a path separator.
     """
@@ -176,7 +176,7 @@ class Config:
     Entities defined in files matching any of these patterns are not documented.
 
     .. important::
-        When building on Windows, all path separators are normalized to :python:`"/"`. 
+        When building on Windows, all path separators are normalized to :python:`"/"`.
         Therefore, in the specified regular expressions, always use :python:`"/"` to
         match a path separator.
     """
@@ -193,8 +193,8 @@ class Config:
     )
     """List of regular expressions matching *allowed* symbols.
 
-    Only symbols matching one of the :py:attr:`allow_symbols` patterns, and not matching
-    :py:attr:`disallow_symbols`, are documented.  By default, all symbols are allowed.
+    Only symbols matching one of the `.allow_symbols` patterns, and not matching
+    `.disallow_symbols`, are documented.  By default, all symbols are allowed.
     """
 
     disallow_symbols: List[Pattern] = dataclasses.field(default_factory=list)
@@ -208,8 +208,8 @@ class Config:
     )
     """List of regular expressions matching *allowed* macros.
 
-    Only macros names matching :py:attr:`allow_macros`, and not matching
-    :py:attr:`disallow_macros`, are documented.
+    Only macros names matching `.allow_macros`, and not matching
+    `.disallow_macros`, are documented.
     """
 
     disallow_macros: List[Pattern] = dataclasses.field(default_factory=list)
@@ -441,6 +441,11 @@ def _get_all_decls(config: Config, cursor: Cursor, allow_file):
             yield from _get_all_decls(config, child, None)
 
 
+NON_DOC_COMMENT = re.compile(
+    r"(^//[^/\!].*$)|(^/\*[^\*\!](?:.|\n)*?\*/$)", re.MULTILINE
+)
+
+
 def split_doc_comment_into_lines(cmt: str) -> List[str]:
     """Strip the raw string of an object's comment into lines.
     :param cmt: the comment to parse.
@@ -450,6 +455,10 @@ def split_doc_comment_into_lines(cmt: str) -> List[str]:
     # so dedent all subsequent lines only
     first_line_end = cmt.find("\n")
     cmt = cmt[:first_line_end] + dedent(cmt[first_line_end:])
+
+    # remove any non-docstring comments
+    cmt = NON_DOC_COMMENT.sub("", cmt)
+
     # split into a list of lines & account for CRLF and LF line endings
     body = [line.rstrip("\r") for line in cmt.splitlines()]
 
@@ -467,12 +476,8 @@ def split_doc_comment_into_lines(cmt: str) -> List[str]:
             for line in body
         ]
         body[-1] = body[-1].rstrip("*/").rstrip()
-    if len(body) > 1:
-        body = dedent("\n".join(body)).splitlines()
+    body = dedent("\n".join(body)).splitlines()
     return [""] if not body else body
-
-
-DOC_COMMENT_PREFIX = re.compile(r"/(//|/\!|\*\*|\*\!)\<?")
 
 
 def get_doc_comment(config: Config, cursor: Cursor):
@@ -485,61 +490,11 @@ def get_doc_comment(config: Config, cursor: Cursor):
     f = location.file
     line = location.line
     end_location = SourceLocation.from_position(translation_unit, f, line, 1)
-    if cursor.raw_comment and DOC_COMMENT_PREFIX.match(cursor.raw_comment):
-        return {
-            "text": "\n".join(split_doc_comment_into_lines(cursor.raw_comment)),
-            "location": _get_location_json(config, end_location),
-        }
-    initial_location = location
-    presumed_file, presumed_line, presumed_column = get_presumed_location(location)
-    comment_lines = []
-    COMMENT = TokenKind.COMMENT
-    while True:
-        prev_location = SourceLocation.from_position(translation_unit, f, line - 1, 1)
-        # `cast` below is workaround for: https://github.com/tgockel/types-clang/pull/2
-        tokens = translation_unit.get_tokens(
-            extent=SourceRange.from_locations(
-                cast(int, prev_location), cast(int, end_location)
-            )
-        )
-        stop = False
-        got_line = False
-        for token in tokens:
-            if token.location == initial_location:
-                stop = True
-                break
-            if token.kind != COMMENT:
-                stop = True
-                break
-            spelling = token.spelling
-            if DOC_COMMENT_PREFIX.match(spelling) is None:
-                stop = True
-                break
-            (
-                new_presumed_file,
-                new_presumed_line,
-                new_presumed_column,
-            ) = get_presumed_location(token.location)
-            if (
-                new_presumed_file != presumed_file
-                or new_presumed_line != presumed_line - 1
-            ):
-                stop = True
-                break
-            presumed_line = new_presumed_line
-            if DOC_COMMENT_PREFIX.match(spelling) is not None:
-                comment_lines.extend(split_doc_comment_into_lines(spelling))
-            got_line = True
-            break
-        if stop or not got_line:
-            break
-        line = line - 1
-        end_location = prev_location
-    if not comment_lines:
-        return None
-    comment_lines.reverse()
+    comment = ""
+    if cursor.raw_comment:
+        comment = "\n".join(split_doc_comment_into_lines(cursor.raw_comment))
     return {
-        "text": "\n".join(comment_lines),
+        "text": comment,
         "location": _get_location_json(config, end_location),
     }
 
--- a/sphinx_immaterial/apidoc/cpp/api_parser.py
+++ b/sphinx_immaterial/apidoc/cpp/api_parser.py
@@ -165,9 +165,9 @@ class Config:
     defined in the standard library and third-party libraries.
 
     .. important::
-        Because this is used as a Regular Expression patterns, this should end in
-        (and exclusively use) the Linux path separator :python:`"/"` (even if
-        building on Windows).
+        When building on Windows, all path separators are normalized to :python:`"/"`. 
+        Therefore, in the specified regular expressions, always use :python:`"/"` to
+        match a path separator.
     """
 
     disallow_paths: List[Pattern] = dataclasses.field(default_factory=list)
@@ -176,9 +176,9 @@ class Config:
     Entities defined in files matching any of these patterns are not documented.
 
     .. important::
-        Because this is used as a Regular Expression patterns, this should end in
-        (and exclusively use) the Linux path separator :python:`"/"` (even if
-        building on Windows).
+        When building on Windows, all path separators are normalized to :python:`"/"`. 
+        Therefore, in the specified regular expressions, always use :python:`"/"` to
+        match a path separator.
     """
 
     disallow_namespaces: List[Pattern] = dataclasses.field(default_factory=list)
--- a/sphinx_immaterial/apidoc/cpp/api_parser.py
+++ b/sphinx_immaterial/apidoc/cpp/api_parser.py
@@ -164,12 +164,21 @@ class Config:
     but this default is not normally usable, because it will include entities
     defined in the standard library and third-party libraries.
 
+    .. important::
+        Because this is used as a Regular Expression patterns, this should end in
+        (and exclusively use) the Linux path separator :python:`"/"` (even if
+        building on Windows).
     """
 
     disallow_paths: List[Pattern] = dataclasses.field(default_factory=list)
     """List of regular expressions matching *disallowed* paths.
 
     Entities defined in files matching any of these patterns are not documented.
+
+    .. important::
+        Because this is used as a Regular Expression patterns, this should end in
+        (and exclusively use) the Linux path separator :python:`"/"` (even if
+        building on Windows).
     """
 
     disallow_namespaces: List[Pattern] = dataclasses.field(default_factory=list)
--- a/sphinx_immaterial/apidoc/cpp/api_parser.py
+++ b/sphinx_immaterial/apidoc/cpp/api_parser.py
@@ -124,16 +124,16 @@ class Config:
     input_path: str = "__input.cpp"
     """Path to the input file to parse.
 
-    This may either be a path to an existing file, or `input_content` may
+    This may either be a path to an existing file, or :py:attr:`input_content` may
     specify its content, in which case the filesystem is not accessed.
 
-    If `input_content` is specified and merely contains :cpp:`#include`
+    If :py:attr:`input_content` is specified and merely contains :cpp:`#include`
     directives, then the actual path does not matter and may be left as the
     default value.
     """
 
     input_content: Optional[bytes] = None
-    """Specifies the content of `input_path`.
+    """Specifies the content of :py:attr:`input_path`.
 
     If unspecified, the content is read from filesystem.
     """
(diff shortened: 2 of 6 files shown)


2bndy5:
The fix for cross-references didn't work. I had to resort to using an explicit role for the cross-refs to actually get resolved. Otherwise they all just show as italics (using a `<cite>` tag in HTML).

I'm not sure what is going on here, but I don't like having to explicitly use a role name for simple cross-refs (especially when there aren't duplicates and no warnings).

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -164,12 +164,21 @@ class Config:
     but this default is not normally usable, because it will include entities
     defined in the standard library and third-party libraries.
 
+    .. important::
+        Because this is used as a Regular Expression patterns, this should end in
```
jbms, line 168: Perhaps change to:

```
When building on Windows, all path separators are normalized to :python:`"/"`.  Therefore, in the specified regular expressions, always use :python:`"/"` to match a path separator.
```

jbms:
> The fix for cross-references didn't work. I had to resort to using an explicit role for the cross-refs to actually get resolved. Otherwise they all just show as italics (using a `<cite>` tag in HTML).
> 
> I'm not sure what is going on here, but I don't like having to explicitly use a role name for simple cross-refs (especially when there aren't duplicates and no warnings).

Turns out the issue is that the C++ apigen extension inserts `default-role` directives in order to ensure the default role is interpreted as `cpp:expr` within doc comments, but then reset it improperly so that it was the docutils default (`title-reference` I believe), rather than the default we set in our `conf.py`, `any`.  I think `title-reference` does not result in nitpick warnings since it is implemented by docutils and not sphinx.  The issue occurred on this page because of the prior apigen example on the same page (the roles are reset with each document).

I'm working on a fix for this and will send out a PR.

Separately, though, the `.member` syntax only works for `py:obj` references, not for `any` references.  So regardless we should probably insert `.. default-role:: py:obj` before those autodoc directives.

jbms:
Thanks!

How about adding some unit tests of the new comment parsing, and other syntax improvements?

jbms:
Regarding `retval`, it sounds like the first argument it is intended for documenting specific values (e.g. error codes), rather than the return type itself (which is already known from the signature):

https://stackoverflow.com/questions/60120282/is-there-a-retval-equivalent-for-out-parameters-in-doxygen-c-c

sphinx_immaterial/apidoc/cpp/apigen.py
```diff
@@ -43,19 +43,22 @@ class ApigenConfig:
     document_prefix: str
     """Sphinx document path prefix for the per-entity documentation pages.
 
-    This should end in :python:`"/"`.
+    .. important::
+        Because this is used as a Regular Expression pattern, this should end in
+        (and exclusively use) the Linux path separator :python:`"/"` (even if
+        building on Windows).
```
2bndy5, line 49: I'm not sure how relevant this is here. I don't think this is used as a regexp, but it does seem that Unix path separators are preferred.

sphinx_immaterial/apidoc/cpp/api_parser.py
2bndy5: I also changed this for the duplicate admonition (in both `dis/allow_paths`).

2bndy5:
> How about adding some unit tests of the new comment parsing, and other syntax improvements?

I'm working on this now... I think I'll use several tests: 1 for each supported Doxygen cmd and another to test various comment syntaxes.

sphinx_immaterial/apidoc/cpp/apigen.py
jbms: Yes, that is true --- I think this was an accidental change.

2bndy5: Yep. It was meant to go in `dis/allow_paths`, but I forgot to remove it after realizing I added it to the wrong file.

Should I alter the admonition? Or should I remove the admonition and make sure the path separators use the Unix form?

jbms: Probably alter it.

2bndy5: altered it.

2bndy5:
I added a couple unit tests. These can be refined as we go, but I just wanted to test the additions in this PR.

BTW, I had to alter array.h (see https://github.com/jbms/sphinx-immaterial/pull/119/commits/ebfe2ba829ea8b0f4d54acaf09083777ea9381a5) because the subsequent line's indentation was getting misinterpreted within the `:param x:` field. This is mostly due to how the `strip_comment()` function was designed. I'd be more comfortable addressing that after we start fetching entire comment blocks (instead of 1 line at a time) because the function is meant to accept an entire comment (be it a block or just 1 line) from `Cursor.raw_comment`; meaning the first line's indenting whitespace is not stripped.

The `strip_comment()` also doesn't take into account non-doctring comments like so:
```cpp
/** A short description. */
// This is not a docstring, but might get lumped into `raw_comment`'s value
auto func();

// This non-docstring comment will cause the regex check for comment prefix to
// skip the following docstr (when using `raw_comment`)
/*! This is a docstring. */
auto func(int);
```

It would be easy to adjust the new `DOC_COMMENT_PREFIX` pattern to capture the docstring comments and pass the matches to `strip_comment()`

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -121,16 +124,16 @@ class Config:
     input_path: str = "__input.cpp"
     """Path to the input file to parse.
 
-    This may either be a path to an existing file, or `input_content` may
+    This may either be a path to an existing file, or :py:attr:`input_content` may
```
jbms, line 127: After my PR to fix the default role issue, these can just be:

```
`.input_content`
```

tests/cpp_api_parser_test.py
```diff
@@ -41,11 +42,87 @@ def test_enable_if_transform():
...
+/** This is a docstring. */
+bool is_used = false;
+
+/*! This is a docstring. */
+bool is_useful = false;
+""",
+    )
+
+    output = api_parser.generate_output(config)
+    doc_strings = [
```
jbms, line 79: I think it would be clearer to test each of these individually, so that there is just a single entity and you can just exactly check it.

jbms:
> I added a couple unit tests. These can be refined as we go, but I just wanted to test the additions in this PR.

Thanks

> 
> BTW, I had to alter array.h (see [ebfe2ba](https://github.com/jbms/sphinx-immaterial/commit/ebfe2ba829ea8b0f4d54acaf09083777ea9381a5)) because the subsequent line's indentation was getting misinterpreted within the `:param x:` field. This is mostly due to how the `strip_comment()` function was designed. I'd be more comfortable addressing that after we start fetching entire comment blocks (instead of 1 line at a time) because the function is meant to accept an entire comment (be it a block or just 1 line) from `Cursor.raw_comment`; meaning the first line's indenting whitespace is not stripped.

Is the change to `array.h` just due to a present bug in the code, or is the issue that doxygen does not allow a continutation line of `\param` to be indented (or rather, interprets the indentation as a blockquote in that case)?

> 
> The `strip_comment()` also doesn't take into account non-doctring comments like so:
> 
> ```c++
> /** A short description. */
> // This is not a docstring, but might get lumped into `raw_comment`'s value
> auto func();
> 
> // This non-docstring comment will cause the regex check for comment prefix to
> // skip the following docstr (when using `raw_comment`)
> /*! This is a docstring. */
> auto func(int);
> ```

Does doxygen allow non-doc comments to be interspersed with doc comments?

I guess it could be reasonable to allow, not sure how common it would be.

> 
> It would be easy to adjust the new `DOC_COMMENT_PREFIX` pattern to capture the docstring comments and pass the matches to `strip_comment()`

It sounds like the current approach, of processing each comment token individually, works fine.  What do you see as the advantage to using `raw_comment`?  It sounds like that doesn't capture all comments, and would need more parsing, compared to the current approach of extracting each comment token separately.

I suppose one issue with the current approach is how to handle stripping of a single leading space on single-line doc comments:

```cpp
///Is this a doc comment?
/// And is this indented relative to it?
```

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -429,6 +441,41 @@ def _get_all_decls(config: Config, cursor: Cursor, allow_file):
             yield from _get_all_decls(config, child, None)
 
 
+def strip_comment(cmt: str = None) -> List[str]:
```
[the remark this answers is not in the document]
2bndy5: Its an artifact from experimenting. It can go.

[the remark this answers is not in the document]
2bndy5: No objections.

2bndy5:
>Is the change to array.h just due to a present bug in the code, or is the issue that doxygen does not allow a continutation line of \param to be indented (or rather, interprets the indentation as a blockquote in that case)?

It mostly because Doxygen does not allow the continuation of a `@param` to be indented. Think of the Doxygen paragraphs like Markdown syntax, not rST. There's also a slight misconception in the function that splits the comment into lines. For example, `raw_comment` of a class member function is written like
```cpp
class DummyClass
{
    /**
     * Some docstring.
     * @param arg This is 1 sentence of a paragraph.
     * This is the second sentence of the same paragraph.
     * @note This is a new paragraph because Doxygen doesn't allow
     * nesting certain commands that expect a single paragraph
     * @par
     * This is a second paragraph in the note because we explicitly use the `par` command
     * @warning This is a new paragraph
     * @parblock
     * This is a second paragraph because we use the `parblock` command.
     * 
     * This is a third paragraph in the warning.
     * @endparblock
     * This is a stray paragraph not in the warning.
     */
    void fun(bool arg);
}
```
The `raw_comment` value will be:
```py
"""/**
     * Some docstring.
     * @param arg This is 1 sentence of a paragraph.
     * This is the second sentence of the same paragraph.
     * @note This is a new paragraph because Doxygen doesn't allow
     * nesting certain commands that expect a single paragraph
     * @par
     * This is a second paragraph in the note because we explicitly use the `par` command
     * @warning This is a new paragraph
     * @parblock
     * This is a second paragraph because we use the `parblock` command.
     * 
     * This is a third paragraph in the warning.
     * @endparblock
     * This is a stray paragraph not in the warning.
     */"""
```
with the line endings obviously.

---

> Does doxygen allow non-doc comments to be interspersed with doc comments?
>
> I guess it could be reasonable to allow, not sure how common it would be.

Doxygen does allow this, and I agree that it isn't all that common. I'm just trying to think of ways that might break our approach. To compensate for this wouldn't be hard (using regex). It just something on the radar, not a really high priority.

---
> I suppose one issue with the current approach is how to handle stripping of a single leading space on single-line doc comments:
>```
>///Is this a doc comment?
>/// And is this indented relative to it?
>```

This is why I join the stripped comment, dedent it, and split it back up again. Generally, Doxygen doesn't have to worry if the whitespace after the comment syntax isn't uniform because Markdown indents are at least 3 spaces or more.

I should note that Doxygen does mandate that comments using the `///` prefix requires that they begin and end with a blank line. So, really that would be written like 
```cpp
///
///This is a doc comment
/// This is another doc comment within the same paragraph.
///
```

---
> It sounds like the current approach, of processing each comment token individually, works fine.

We need to improve the traversal better. Currently, it only returns a comment token if the comment is a single line. It is much more common to use `/** ... multi-lined block ... */`.  The token approach also doesn't capture trailing comments used as docstrings
```cpp
struct dummy
{
   int var; ///< This is the docstring.
}
```
However, `raw_comment` will easily capture the comment (albeit a leading block or a trailing single line). I also did a quick test, and `raw_comment` does work for macros -- I'm not sure what that other extension's src comment was talking about anymore.

I just think its over complicated to ignore clang's functionality and parse the source in python (even though clang has already done it for us).

jbms:
> > Is the change to array.h just due to a present bug in the code, or is the issue that doxygen does not allow a continutation line of \param to be indented (or rather, interprets the indentation as a blockquote in that case)?
> 
> It mostly because Doxygen does not allow the continuation of a `@param` to be indented. Think of the Doxygen paragraphs like Markdown syntax, not rST.

I was attempting to support both rST syntax and a subset of doxygen syntax at the same time.  That also would make it relatively easy to migrate incrementally.  Do you think that is still feasible, or do we have to have a configuration option to interpret as either doxygen syntax or rST syntax, which would make it significantly more difficult to migrate?

Maybe we can just convert the continuation lines after these doxygen commands appropriately.
 
> > Does doxygen allow non-doc comments to be interspersed with doc comments?
> > I guess it could be reasonable to allow, not sure how common it would be.
> 
> Doxygen does allow this, and I agree that it isn't all that common. I'm just trying to think of ways that might break our approach. To compensate for this wouldn't be hard (using regex). It just something on the radar, not a really high priority.
> 
> > I suppose one issue with the current approach is how to handle stripping of a single leading space on single-line doc comments:
> > ```
> > ///Is this a doc comment?
> > /// And is this indented relative to it?
> > ```
> 
> This is why I join the stripped comment, dedent it, and split it back up again. Generally, Doxygen doesn't have to worry if the whitespace after the comment syntax isn't uniform because Markdown indents are at least 3 spaces or more.

Okay, I guess in any case when using `raw_comment` it won't be an issue.
> 
> I should note that Doxygen does mandate that comments using the `///` prefix requires that they begin and end with a blank line. So, really that would be written like
> 
> ```c++
> ///
> ///This is a doc comment
> /// This is another doc comment within the same paragraph.
> ///
> ```

I would prefer to relax this blank line restriction --- unless it serves some important purpose.

> 
> > It sounds like the current approach, of processing each comment token individually, works fine.
> 
> We need to improve the traversal better. Currently, it only returns a comment token if the comment is a single line. It is much more common to use `/** ... multi-lined block ... */`. The token approach also doesn't capture trailing comments used as docstrings
> 
> ```c++
> struct dummy
> {
>    int var; ///< This is the docstring.
> }
> ```
> 
> However, `raw_comment` will easily capture the comment (albeit a leading block or a trailing single line). I also did a quick test, and `raw_comment` does work for macros -- I'm not sure what that other extension's src comment was talking about anymore.
> 
> I just think its over complicated to ignore clang's functionality and parse the source in python (even though clang has already done it for us).

If `raw_comment` handles all of the cases that we care about, then it sounds like that is the way to go.  I wasn't aware of it when I wrote this code.  It looks like it does handle multiple consecutive comments, and also looks like we can get the source range via `clang_Cursor_getCommentRange` C function (that we can call from Python).

2bndy5:
> Do you think that is still feasible, or do we have to have a configuration option to interpret as either doxygen syntax or rST syntax

We can provide a config option; that was my original intention. However, people using breathe may already have their custom doxygen aliases setup to inject actual RST in their doc strings. This often looks like
```cpp
@rst
.. note:: 
    This is an admonition.

    This is a second paragraph in the note.
@endrst
```
Where `@rst` and`@endrst` commands are `ALIASES` created in their doxygen config to pass raw RST as `verbatim` blocks in XML.

2bndy5:
There's a huge difference between Doxygen syntax (which is mostly markdown) and RST syntax. I don't think expecting to parse a mix of the 2 is sustainable.

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -429,6 +441,41 @@ def _get_all_decls(config: Config, cursor: Cursor, allow_file):
...
+    """Strip the raw string of an object's comment into lines.
+    :param cmt: the comment to parse.
+    :returns: A list of the lines without the surrounding C++ comment syntax.
+    """
+    if cmt is None:
```
2bndy5, line 449: I'll remove this when I'm sure that the `raw_comment` is not None.

tests/cpp_api_parser_test.py
2bndy5: If I hadn't done it this way, then I never would've guessed that the `re.sub(brief|details, "", txt)` call needed the `re.MULTILINE` flag.

Are you ok with testing each regex pattern per test, instead of testing each supported command per test?

jbms: I'm not sure I understand why this particular test structure helped spot that issue.  But I just meant that it would be clearer to have the test just generate one entity at a time, so that each test case is more isolated, at least for these comment syntax tests.

2bndy5: I noticed that the re.sub() call was only replacing the first `\\brief` on the first line but not `\\details` on the third line of the same comment... The multiline flag fixed that.

I could split up this test into per comment syntax form. Initially, I thought this comment was for the other `test_function_fields()` test. 

Looking at the regex patterns, I don't know why you're looking for some fields like `checks`, `dchecks`, `schecks` (not really sure what they are supposed to represent). I get the `pre`, `post`, and `invariant` commands because those actually exist for Doxygen. The `error` (which isn't treated like an admonition) and `requires` (which seems equivalent to [`@concept`](https://doxygen.nl/manual/commands.html#cmdconcept)) commands seem to be something you added (they don't exist for Doxygen).

jbms: Yes, checks, error, and requires are "custom" commands I added for tensorstore, along with corresponding custom sphinx docfields that I left out.  I agree they shouldn't be provided by default --- it would be good instead to have a mechanism for custom doxygen commands though.

2bndy5: We can look into implementing something similar to [Doxygen's `ALIASES` setting](https://doxygen.nl/manual/config.html#cfg_aliases). That is [how breathe supports raw RST in Doxygen's XML output](https://breathe.readthedocs.io/en/latest/markups.html#aliases). Theoretically it can be used to add custom commands like the ones you're adding.

jbms:
> There's a huge difference between Doxygen syntax (which is mostly markdown) and RST syntax. I don't think expecting to parse a mix of the 2 is sustainable.

What would the migration path be, then?

My own opinion is that rST syntax is better than doxygen syntax, and I imagine that may naturally be a common opinion among those using Sphinx for documentation --- when using doxygen rST syntax is not an option, but here it is.

But if it has to be toggled via a global option, then it would be difficult for users to gradually make use of more rST functionality.  Instead they would have to convert all of their comments at once.

I suppose the user could have two entries in `cpp_apigen_configs`, one that filters by filename or symbol to include only entities documented using doxygen syntax, and another entry for entities documented using rST syntax.  But not sure how convenient that will be.  And currently with the way separate apigen configs are handled, you can't make an entity from one config a related entity of one defined in the other config.  Perhaps that could be fixed, though.

Alternatively we could have a separate regular expression to indicate which symbols are documented using doxygen syntax.

2bndy5:
If they're migrating from Doxygen, then there will inevitably need to be changes made to the comments. If they're migrating from breathe, then it should be less painless because they're likely already writing their comments in rst (albeit encapsulated in
`@rst ... @endrst` blocks). I'm not against supporting an `@rst`/`@endrst` command blocks out-of-the-box because its a very preferred approach for documenting C++ with Doxygen+breathe[+exhale]. Inline RST has a slightly different beginning command
(`@rst-inline ... @endrst`)

Am I understanding your use of "migrating" correctly?

I agree in that RST is the superior documenting syntax. I often look at motivations for using mkdocs or MyST parser as _seriously_ flawed opinions. My grievances against Doxygen are endless...

jbms:
Are you going to change this to use `raw_comment` ?

2bndy5:
> Does breathe somehow define these automatically, or is it up to the user to define these as aliases in their doxygen config, and these are just the recommended names? 

Breathe recommends these names but does not automatically define them. Some people use exhale which automatically define these commands (based on the breathe recommendations). FYI, exhale is just a wrapper for breathe that aims to be more automated and a quicker onboarding of breathe into projects; it also includes some patches for breathe...

> Are you going to change this to use `raw_comment`?

yes, I've already got a stash for this. Currently, it will default to the line-by-line search if `raw_comment` is null.

tests/cpp_api_parser_test.py
2bndy5: I split up the comment stripping tests into a `class` of tests. With using `raw_comment`, the test for trailing docstring comment is passing. 🎉 

I could split up the test about function fields, but I think that can be done after the `_normalize_doc_comment()` function matures more (it is currently pretty limited).

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -429,6 +441,41 @@ def _get_all_decls(config: Config, cursor: Cursor, allow_file):
...
+        multi_lined_asterisk = True  # works also for single-line comments blocks
+        if len(body) > 1:
+            line_has_asterisk = [line.startswith("*") for line in body[1:]]
+            multi_lined_asterisk = line_has_asterisk.count(True) == len(body) - 1
+        body = [
+            (line.lstrip("*").lstrip("<") if multi_lined_asterisk else line)
+            for line in body
+        ]
+        body[-1] = body[-1].rstrip("*/").rstrip()
+    body = dedent("\n".join(body)).splitlines()
```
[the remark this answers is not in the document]
2bndy5: >  But I think we should only dedent multiline comments.

I added a condition to skip dedenting single line comments, but the doc builds fail because the leading whitespace is now being interpreted as content to the ext's generated `.. highlight:: cpp` directive.

> ```
> <cpp_apigen_rst_prolog>:6: ERROR: Error in "highlight" directive:
> no content permitted.
> .. highlight:: cpp
>  Returns the data order.
> ```

Kinda glad I did the rebase on main now (this would've gone unnoticed until much later).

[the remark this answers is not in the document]
2bndy5: To be clear, A single line comment (with respect to `raw_comment`) is a block that has 1 line

```cpp
/// A single line comment

/// Not a single line comment because
/// it has multiple lines in sequence
```
So, I'm not sure what the advantage is for not dedenting a single line comment. Maybe we could delegate this to regex with `re.sub(r"^(\s{0, 3})\S", "", text)` (notice there's no multiline flag). Any other concerns I can think of all require multiple lines (eg. lists)

jbms:
I think this currently doesn't handle the case where you have multiple consecutive comments of different types, e.g. a `/*` comment followed by a `//` comment.

2bndy5:
> I think this currently doesn't handle the case where you have multiple consecutive comments of different types, e.g. a `/*` comment followed by a `//` comment.

Correct. I think that's a rare scenario but I could do some extra handling in `split_comment_into_line()`.

tests/cpp_api_parser_test.py
```diff
@@ -41,11 +42,134 @@ def test_enable_if_transform():
...
+
+    def assert_output(self, expected: str):
+        output = api_parser.generate_output(self.config)
+        doc_strings = [v["doc"]["text"] for v in output.get("entities", {}).values()]
+        assert expected in doc_strings
+
+    def test1(self):
+        self.config.input_content = b"\n".join(
+            [
+                b"//! This is a docstring.",
```
jbms, line 69: Sorry for the confusion regarding the dedenting --- if we have e.g.


```cpp
//! This is a docstring.
//! This is the next line of the docstring.
```

Then we want the resultant doc text to be:

```python
"This is a docstring\nThis is the next line of the docstring."
```

If we don't dedent I think that will cause problems.

Regarding the highlight issue, we just need to add extra new lines I guess.

tests/cpp_api_parser_test.py
```diff
@@ -41,11 +42,134 @@ def test_enable_if_transform():
...
+    config = api_parser.Config(
+        input_path="a.cpp",
+    )
+
+    def assert_output(self, expected: str):
+        output = api_parser.generate_output(self.config)
+        doc_strings = [v["doc"]["text"] for v in output.get("entities", {}).values()]
+        assert expected in doc_strings
+
+    def test1(self):
```
jbms, line 66: Potentially these test cases could be made into a single parameterized test case using `pytest.mark.parametrize`, where one parameter is the input source code and the other parameter is the expected docstring.

2bndy5: I don't have a lot of experience with pytest. I'm actually writing unit tests with pytest right now for a different project (needed my overdue attention). So, I will look into the `mark.parameterize` idea.

tests/cpp_api_parser_test.py
```diff
@@ -41,11 +42,134 @@ def test_enable_if_transform():
...
+        input_path="a.cpp",
+        input_content="""
+/// @brief A dummy function for tests.
+///
+/// \\details A more detailed paragraph.
+///
+/// @param arg1 An arg passed by value.
+/// \\param[in] arg2 An unaltered arg passed by reference.
+/// @param[in, out] arg3 An arg passed by reference that gets altered.
+/// @retval int A negative number indicating a specific error, otherwise 0.
```
jbms, line 144: I think to avoid confusion this example should be changed to use `retval` in the way it seems to be intended:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-c-function

Specifically, both Sphinx and doxygen seem to intend that the first argument to `retval` is the value being returned, not the type.

tests/cpp_api_parser_test.py
2bndy5: That's the way I originally had it in `strip_comment_into_lines()`. I'll change it back... My other concern is that an oddly formed block (with slightly non-uniform prefixed whitespace) might screw up the expected result when using `dedent()`:
```cpp
//!This is a docstring.
//! This is the next line of the docstring.
```
which will end up being
```py
"This is a docstring\n This is the next line of the docstring."
```
notice the space on at start of 2nd line

jbms: Yes, that seems like expected behavior, though, and up to the user to fix.

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -429,18 +441,57 @@ def _get_all_decls(config: Config, cursor: Cursor, allow_file):
...
         location = token.location
         break
     else:
         location = cursor.location
-    initial_location = location
     f = location.file
     line = location.line
-    presumed_file, presumed_line, presumed_column = get_presumed_location(location)
     end_location = SourceLocation.from_position(translation_unit, f, line, 1)
+    if cursor.raw_comment and DOC_COMMENT_PREFIX.match(cursor.raw_comment):
```
jbms, line 488: Is there any case where we need the old logic?  Otherwise we may as well remove it.

2bndy5: Not that I know of right now. I can remove it, and if we need it in the future, then we can fetch it from the git history.

2bndy5: removed old logic.

I also removed the regex `DOC_COMMENT_PREFIX` because the new `NON_DOC_COMMENT` regex is used to remove any non-docstring comments from within the given `raw_comment`.

tests/cpp_api_parser_test.py
2bndy5: changed this to
```
/// @retval NULL If unsuccessful.
```

I don't often use this command in doxygen because it doesn't group them into the `@returns` paragraph(s).

tests/cpp_api_parser_test.py
2bndy5: Took me a while, but I got the hang of using `pytest.mark.parametrize` ("parametrize" is kind of a weird spelling IMO). The only down side to this is that it makes failed tests harder to read (especially when parametrizing numerous args).

I also added a couple tests to check that non-docstring comments are removed.

2bndy5:
## https://github.com/jbms/sphinx-immaterial/pull/119/commits/3a153c6b98b5d0a9e88ae442246902e69af3dc3d passed the unit tests, but failed to build the docs
Not sure what happened, but I think its trying to document some header gaurds:
```
/home/runner/work/sphinx-immaterial/sphinx-immaterial/docs/cpp_apigen_demo/index_interval.h.rst:1: WARNING: No group or relates specified for entity CPP_APIGEN_DEMO_INDEX_INTERVAL_H
/home/runner/work/sphinx-immaterial/sphinx-immaterial/docs/cpp_apigen_demo/fake_stdlib.h.rst:1: WARNING: No group or relates specified for entity CPP_APIGEN_DEMO_FAKE_STDLIB_H
/home/runner/work/sphinx-immaterial/sphinx-immaterial/docs/cpp_apigen_demo/array.h.rst:1: WARNING: No group or relates specified for entity CPP_APIGEN_DEMO_ARRAY_H
```
Maybe this is related to removing the old line-by-line algorithm.

Other warning reads
```
WARNING: Disambiguating 2 overloads of operator-subscript using numerical ids.  Definitions at cpp_apigen_demo/array.h:58, cpp_apigen_demo/array.h:59
```
But I'm not sure what "operator-subscript" means.

docs/cpp_apigen_demo/index_interval.h
```diff
@@ -20,6 +20,8 @@ class IndexInterval {
   friend std::ostream& operator<<(std::ostream& os, IndexInterval x);
 
   /// Returns the lower bound.
+  ///
+  //! @retval int The lower bound.
```
jbms, line 24: Change to @returns or something.

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -429,68 +441,60 @@ def _get_all_decls(config: Config, cursor: Cursor, allow_file):
             yield from _get_all_decls(config, child, None)
 
 
+NON_DOC_COMMENT = re.compile(
+    r"(^//[^/\!].*$)|(^/\*[^\*\!](?:.|\n)*?\*/$)", re.MULTILINE
```
jbms, line 445: I think if we have `/*` inside of a `///` doc comment, it will also get stripped out.  Also, if we just delete lines, then the line numbers won't be correct.  We could either change to specifying separate location info for each line, or could replace each removed line with an empty line.

Additionally, the logic below for stripping out the comment syntax assumes just a single comment type.

I think to properly handle multiple comment types a different strategy may be required here:
1. Parse `cmt` into a sequence of comments and whitespace (can be done with a regular expression).
2. Coalesce consecutive doc comment lines that start with `///` or `//!`.  I think you could actually design the regular expression so that this can be done as part of step 1.
3. Remove any non-doc comments, including the comment begin/end marker, but preserve any line breaks to avoid messing up the line numbers.
4. For each coalesced doc comment, strip the comment begin/end marker and dedent the text.

jbms:
Thanks!

> ## [3a153c6](https://github.com/jbms/sphinx-immaterial/commit/3a153c6b98b5d0a9e88ae442246902e69af3dc3d) passed the unit tests, but failed to build the docs
> Not sure what happened, but I think its trying to document some header gaurds:
> 
> ```
> /home/runner/work/sphinx-immaterial/sphinx-immaterial/docs/cpp_apigen_demo/index_interval.h.rst:1: WARNING: No group or relates specified for entity CPP_APIGEN_DEMO_INDEX_INTERVAL_H
> /home/runner/work/sphinx-immaterial/sphinx-immaterial/docs/cpp_apigen_demo/fake_stdlib.h.rst:1: WARNING: No group or relates specified for entity CPP_APIGEN_DEMO_FAKE_STDLIB_H
> /home/runner/work/sphinx-immaterial/sphinx-immaterial/docs/cpp_apigen_demo/array.h.rst:1: WARNING: No group or relates specified for entity CPP_APIGEN_DEMO_ARRAY_H
> ```
> 
> Maybe this is related to removing the old line-by-line algorithm.

If there is no doc comment for an entity, `get_doc_comment` should return `None` (as it did previously) --- that will exclude it from the documentation.

> 
> Other warning reads
> 
> ```
> WARNING: Disambiguating 2 overloads of operator-subscript using numerical ids.  Definitions at cpp_apigen_demo/array.h:58, cpp_apigen_demo/array.h:59
> ```
> 
> But I'm not sure what "operator-subscript" means.

That means `operator[]`.

sphinx_immaterial/apidoc/cpp/api_parser.py
2bndy5: For `//` comments that get stripped, the line break should still remain, although `dedent()` may be stripping empty lines (I can't remember what the docs said about that).

For `/* ... */` comments that get stripped, only the final line break is preserved. It would take multiple regex patterns to preserve the line breaks for comments that span multiple lines (see the last test in `test_comment_styles()`).

I'd like to say it is easier to add empty lines to preserve the src's line number, but maybe not. Adjusting the lines' number for the src seems harder, but I haven't looked into that yet.

The steps you propose might require an approach that doesn't use `decent()`, which I opted for out of convenience.

In my research, I translated the stripped `raw_comment` line-by-line while keeping track of the indentation depth -- it wasn't perfect for `@code\n ... \n ... \n@endcode` blocks, but it worked. I was hoping to refactor the comment stripping/parsing out of api_parse.py before taking a line-by-line approach.

jbms: I think it will be possible to use `dedent`, we just need to apply it after splitting by comment type.

Once you split by comment type it is also easy to remove a comment while preserving newlines.  Just remove all characters except `\n`.

tests/cpp_api_parser_test.py
2bndy5: I reverted dedenting behavior. `raw_comment`s are first dedented (all lines but the first because clang does it for the first line), then they are dedented again after stripping the comment syntax out.

I also added some IDs to this test so the pytest output is more understandable.

sphinx_immaterial/apidoc/cpp/api_parser.py
2bndy5: I split the raw_comment into chunks based on the `re.search()` result's `start/end()`. Now it strips the sections that are docstrings (1 at a time), replaces the non-doc comments with blank lines (determined by `match.group().count("\n")`), and finally dedents the striped docstrings.

I think this should satisfy the concern of keeping the src's line numbers consistent in any sphinx warnings.

jbms:
Thanks!

This looks ready to merge!

2bndy5:
Sorry about the messy guy history, I was kinda hoping this would get squashed.

Also sorry for the delay, that other project that needed my attention has turned into a "whack-a-mole" carnival game. Ironically, it also uses the clang project... I'm going to need a vacation from clang when we get this feature finished. 😂

PR #124, opened 2022-07-12
Support dirhtml builder

### LINKED ISSUE #110: Add support for dirhtml builder
It would be great to get support for the `dirhtml` builder, which I use to replicate the default `mkdocs` behvaiour (turns `doc.rst` into `doc/index.html`, etc).

Naively, it looks like the following patch might be all that is required:

```patch
$ git diff
diff --git a/sphinx_immaterial/__init__.py b/sphinx_immaterial/__init__.py
index d3005c8..eca458a 100644
--- a/sphinx_immaterial/__init__.py
+++ b/sphinx_immaterial/__init__.py
@@ -321,6 +321,7 @@ def setup(app):
     )

     app.add_builder(_get_html_builder(app.registry.builders["html"]), override=True)
+    app.add_builder(_get_html_builder(app.registry.builders["dirhtml"]), override=True)
     app.add_html_theme("sphinx_immaterial", os.path.abspath(os.path.dirname(__file__)))

     # register our custom adminition directive that are tied to the theme's CSS
```

I am happy to post a PR with that if there are no objections.

jbms:
Fixes #110

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,7 +69,7 @@ jobs:
       run: sphinx-build . _build/html -W --keep-going -T
     - name: Build dirhtml docs
       working-directory: docs
-      run: sphinx-build . _build/dirhtml -W --keep-going -T
+      run: sphinx-build . _build/dirhtml -W --keep-going -T -b dirhtml
     - name: Build latex docs
       working-directory: docs
       run: sphinx-build -b latex . _build/latex -W --keep-going -T
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,8 @@ jobs:
       matrix:
         os:
           - 'ubuntu-latest'
+          - 'windows-latest'
+          - 'macos-latest'
         python-version:
           - '3.7'
           - '3.8'
@@ -38,8 +40,11 @@ jobs:
     - name: Install Python packaging & linting tools
       run: python -m pip install -r dev-requirements.txt -r requirements.txt
     - run: mypy sphinx_immaterial/**/*.py
+      shell: bash
     - run: pylint sphinx_immaterial/**/*.py
+      shell: bash
     - run: black --diff --color sphinx_immaterial/**/*.py
+      shell: bash
     - name: Check for dirty working directory
       run: git diff --exit-code
     - name: Show package version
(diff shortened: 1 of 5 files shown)


2bndy5:
Building this theme on windows is a pain (still need to use WSL). If I can download the artifacts, then it is just as good as doing the entire workflow locally (just to inspect the rendered output of a PR).

2bndy5:
I suppose I could keep downloading the built wheel artifacts and build the docs myself using that, but it is just an extra step to getting a rendered output.

It is also handy to know that any future changes don't break this feature. The step for the latex builder has proven useful for that as well (and those builds are included in the artifacts).

2bndy5:
Well I see you added Windows and Mac OS to the CI, but you can't use the wheels built from windows because the CSS and JS bundles aren't named correctly (see #18 - abridged hash is missing from the minified filenames). Thus the CI artifacts of the doc builds are... um... broken.

The doc build artifacts are being overwritten by the non-ubuntu jobs. Since that is currently the case, I hesitate to even install the wheel that is uploaded as an artifact.

.github/workflows/build.yml
```diff
@@ -8,6 +8,8 @@ jobs:
       matrix:
         os:
           - 'ubuntu-latest'
+          - 'windows-latest'
+          - 'macos-latest'
```
2bndy5, line 12: While I appreciate the inclusiveness here, the windows wheels are actually false-positive results (un-useable).

Personally, I'm rather anit-Apple, so I can't vouch for anything that VM produces. Github CI runners using macos actually have usage limits because they can't be served from Microsoft Azure servers. I doubt you'll hit these limits on this repo, but its yet another way Apple milks it's own enthusiasts for 💰.
```suggestion
          # - 'windows-latest'
          # - 'macos-latest'
```

2bndy5:
The last real problem may actually be related to the html templates. This is the rendered result from a windows wheel:
```html
      <link rel="stylesheet" href="_static/stylesheets/main.css">
      
        
        <link rel="stylesheet" href="_static/stylesheets/palette.css">
        
```
At least the bundles' filenames are named correctly now.

jbms:
I think I have now fixed the issue of artifacts being overwritten, and I think I have also fixed #18 (sounds like maybe it is not quite fixed yet, though). We could file that as an upstream bug as well.

I agree that Apple makes it annoying to develop for macOS in many ways, but a lot of developers use macOS laptops so it is useful to support it if not too difficult.

2bndy5:
Could rebase on latest main?

This is looking great to solve #110 !

.github/workflows/build.yml
```diff
@@ -48,45 +53,80 @@ jobs:
...
       run: pip install dist/*.whl -r docs/requirements.txt
+      shell: bash
     - run: pytest -vv -s tests/
       name: Run Python tests
     - name: Build html docs
       working-directory: docs
       run: sphinx-build . _build/html -W --keep-going -T
+    - name: Build dirhtml docs
+      working-directory: docs
+      run: sphinx-build . _build/dirhtml -W --keep-going -T
```
2bndy5, line 72: need to actually use the dirhtml builder here. I just realized I'm inspecting duplicates of the default html builder.
```suggestion
      run: sphinx-build . _build/dirhtml -W --keep-going -T -b dirhtml
```

2bndy5 (approved):

PR #128, opened 2022-07-14
Fix Python type_annotation_transforms for Python<3.9

### LINKED ISSUE #127: Issue with `python_transform_type_annotations_pep604` not adding `|` correctly
I'm observing some issues with `python_transform_type_annotations_pep604`. There are times when `Union[a, b]` isn't converted to `a | b` but rather `a, b`. The same happens when wrapping in an `Optional[]` type hint.

Here is a reproducible `foo` project. I'm testing with current main (00c681084ef9ec4be3a936d30367f64802bc7854).

[foo-example.zip](https://github.com/jbms/sphinx-immaterial/files/9102329/foo-example.zip)

```python
def multiplication(x: Union[int, float], y: int) -> float:
```

image

```python
def func(x: int, y: Optional[Union[int, float]] = None) -> int:
```

image

jbms:
Fixes #127.


mhostetter:
Thanks @jbms! I can confirm this PR resolves the issues pointed out in #127. 👍

2bndy5 (approved):

PR #136, opened 2022-07-22
Add `include_rubrics_in_toc` object description option

### LINKED ISSUE #126: Add additional fields in `python-apigen`?
I really like the new section headers (Parameters, Returns, Raises) for `python-apigen` functions and methods. Would it be worthwhile to  add section headers for other "rubrics", e.g. Notes, References, Examples? Perhaps for *all* rubrics in a docstring?

Just a thought...

image

mhostetter:
Fixes #126. I believe I correctly implemented this feature. Please let me know your thoughts. I tested this locally on my project and on a standalone `foo` project.

The option is disabled by default. When enabled, all rubrics (that haven't already been converted to fields or admonitions by Napoleon) are added to the TOC. Here's an example in this `foo` module.  [foo-example.zip](https://github.com/jbms/sphinx-immaterial/files/9167987/foo-example.zip)

image

### DIFF
--- a/sphinx_immaterial/apidoc/object_toc.py
+++ b/sphinx_immaterial/apidoc/object_toc.py
@@ -87,6 +87,22 @@ def _monkey_patch_toc_tree_process_doc(app: sphinx.application.Sphinx):
         section += titlenode
         return section
 
+    def _make_section_from_rubric(
+        source: docutils.nodes.rubric,
+    ) -> Optional[docutils.nodes.section]:
+        rubric = cast(docutils.nodes.rubric, source)
+        ids = rubric["ids"]
+        if not ids:
+            # Not indexed
+            return None
+        section = docutils.nodes.section()
+        section["ids"] = ids
+        title = rubric.astext()
+        # Sphinx uses the first child of the section node as the title.
+        titlenode = docutils.nodes.comment(title, title)
+        section += titlenode
+        return section
+
     def _make_section_from_term(
         source: docutils.nodes.term,
     ) -> Optional[docutils.nodes.section]:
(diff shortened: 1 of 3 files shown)


mhostetter:
The only *slightly* non-ideal thing I noticed was that rubrics from the class definition (not the `__init__()` signature) are included in the class TOC on the left-hand side.

@jbms have any thoughts on how that might be avoided?

I noticed that when you click on the class `FLFSR` on the left side, the "Notes", "References", and "Examples" TOC entries are not shown on the left.

image

But when you select an individual method / property, then the "Notes", "References", and "Examples" appear on the left.

image

sphinx_immaterial/apidoc/object_toc.py
```diff
@@ -143,6 +159,12 @@ def _collect(
                 if new_node is not None:
                     target += new_node
                     target = new_node
+            elif isinstance(source, docutils.nodes.rubric):
+                # Rubric.  Try to create synthetic section.
+                new_node = _make_section_from_rubric(source)
+                if new_node is not None:
+                    target += new_node
+                    target = new_node
```
jbms, line 167: No need to assign to target, instead should return as in the `term` case before since rubrics cannot contain children (because they are just headings, not sections).  If you want them to be able to have chosen that would require more work.

jbms:
> The only _slightly_ non-ideal thing I noticed was that rubrics from the class definition (not the `__init__()` signature) are included in the class TOC on the left-hand side.
> 
> @jbms have any thoughts on how that might be avoided?

I agree that this is not good.  Does this issue occur with all sections, or just rubrics?  If it occurs with all sections, the fix will be in nav_adapt.py

mhostetter:
> I agree that this is not good. Does this issue occur with all sections, or just rubrics? If it occurs with all sections, the fix will be in nav_adapt.py

By "all sections" do you mean the fields you convert into TOC headings, like "Parameters" and "Returns"? If so, yes. (I'm not sure how else to add section headings in a docstring.)

I just added "Parameters" and "Returns" into the class docstring and they manifest the same way the rubric TOC entries do.

image

jbms:
Thanks, this looks perfect.  The issue with the global toc is unrelated so it could be fixed separately I suppose.

mhostetter:
@jbms thanks. I'm looking through the `nav_adapt.py` code and trying to figure it out. If I get too stuck, I may ask for help.

mhostetter:
Ok... so... there's a lot going on in `nav_adapt.py`. Perhaps handing the global TOC *should* be a standalone issue. I'll open a new issue to track that. I'm ready to merge this PR whenever you all are ready. Thanks.

jbms:
Thanks

PR #139, opened 2022-07-26
Fix several issues with nav_adapt and add unit tests

### LINKED ISSUE #137: Fields/rubrics from class description also appear in left-side TOC
It was noted in https://github.com/jbms/sphinx-immaterial/pull/136#issuecomment-1192569737 that when `include_fields_in_toc=True` or `include_rubrics_in_toc=True` fields and rubrics from a class docstring (not `__init__()`) appear in the left-side TOC. It is desired that they only appear in the right-side TOC on the class description page.

[foo-example.zip](https://github.com/jbms/sphinx-immaterial/files/9173852/foo-example.zip)

image

image

### LINKED ISSUE #138: local toc on site's root index.html has malformed URLs
I just noticed in v0.8.0, the URLs in the local toc are missing the hashlocation. This is reproduced in this theme's docs, I also enabled builds for v0.7.3 on RTD for any further comparison.
- [v0.8.0 -- The theme's landing page](https://sphinx-immaterial.readthedocs.io/en/latest/) has a local toc item named "Getting Started" that links to https://sphinx-immaterial.readthedocs.io/en/latest/index 
- [v0.7.3 -- The theme's landing page](https://sphinx-immaterial.readthedocs.io/en/v0.7.3/) has a local toc item named "Getting Started" that links to https://sphinx-immaterial.readthedocs.io/en/v0.7.3/#getting-started

jbms:
Fixes #137.
Fixes #138.

Also fixes support for `only` directives that affect sections.


2bndy5:
I can confirm this fixes #138 and allows `.. only::` (tested w/ `html` as arg) applied to `.. toctree::`.

@mhostetter can you confirm #137 is satisfied? I tested it with our autodoc example, but not with the apigen example.

@jbms I'm noticing the first entry in the global toc is missing for sub root pages.
For example "General API customization" page isn't listed in the global toc
image
but I can get to it by clicking "Formatting signatures" and hitting the "Previous" button at the bottom. I first noticed this when the demo_api page wasn't listed in the examples pages.

mhostetter:
>  I'm noticing the first entry in the global toc is missing for sub root pages.

I noticed this too.

> @mhostetter can you confirm https://github.com/jbms/sphinx-immaterial/issues/137 is satisfied? I tested it with our autodoc example, but not with the apigen example.

With `include_rubrics_in_toc=True`, I'm noticing that all the rubrics that were added to the TOC are now removed from both the left- and right-side TOCs (which is equivalent behavior to `include_rubrics_in_toc=False`).

<details>
<summary>Class example</summary>

image

</details>

<details>
<summary>Function example</summary>

image

</details>

2bndy5:
when I added
```rst
.. rubric:: A rubric
```
 to the autodoc'd class' docstr, the rubric showed in the local toc:
image
I'm not sure if that's what was requested.

mhostetter:
> I'm not sure if that's what was requested.

That is the desired behavior. Not sure why I'm seeing something different. Perhaps it's the difference between `autodoc` and `python-apigen`. I'll have to do more testing.

jbms:
Thanks for the testing.  The issue with `include_rubrics_in_toc` was just that I had accidentally based my branch on an older commit before `include_rubrics_in_toc` was merged in.  I have now rebased it.

Additionally, I have fixed the issue whereby the first child of a `toctree` with a `:caption:` was hidden.

Separately, I refactored the python domain fixes to split them into separate files, as was done previously for C++.  As part of that refactor, I ensured that all monkey patching happens at module init time rather than during the `setup` function, to avoid issues if `setup` is called more than once (e.g. during unit tests where we build more than once per process).

jbms:
> I can confirm this fixes #138 and allows `.. only::` (tested w/ `html` as arg) applied to `.. toctree::`.

To be clear, while it will build with `only` applied to a `toctree`, due to the linked Sphinx bug the `only` directive has no effect on the global toc --- the toctree is always included in the global toc regardless of the condition.

mhostetter:
> Thanks for the testing. The issue with include_rubrics_in_toc was just that I had accidentally based my branch on an older commit before include_rubrics_in_toc was merged in. I have now rebased it.
 
Thanks. It does appear that the rubrics are working as intended. See below that the "Examples" rubric is suppressed from the global TOC, which was the goal of #137. 👍 

I did notice, though, that with `python-apigen` object pages now seem to have an extra heading. See below that the `galois.BCH.generator_poly` page with the yellow "P" marker appears as the first heading in the local TOC and `generator_poly` without the yellow "P" is added to the global TOC. The same is true for the extra heading in the `galois.BCH` page.

<details>
<summary>With `main`</summary>

image

</details>

<details>
<summary>With 488e35f</summary>

image

</details>

jbms:
> I did notice, though, that with `python-apigen` object pages now seem to have an extra heading. See below that the `galois.BCH.generator_poly` page with the yellow "P" marker appears as the first heading in the local TOC and `generator_poly` without the yellow "P" is added to the global TOC. The same is true for the extra heading in the `galois.BCH` page.

This was due to a bug in my refactoring of the Python domain monkey patches, which broke `nonodeid` (used by Python apigen).  That is now fixed and I added a unit test.

mhostetter:
> This was due to a bug in my refactoring of the Python domain monkey patches, which broke nonodeid (used by Python apigen). That is now fixed and I added a unit test.

Awesome, thanks. I can confirm this works for me now. 👍 

> Separately, I refactored the python domain fixes to split them into separate files, as was done previously for C++.

I like the refactoring -- makes the code easier to follow.

sphinx_immaterial/apidoc/format_signatures.py
```diff
@@ -58,7 +58,7 @@ class CollectSignaturesTransform(sphinx.transforms.SphinxTransform):
 
     def apply(self, **kwargs: Any) -> None:
         collected_signatures = _get_collected_signatures(self.env)
-        for node in self.document.traverse(sphinx.addnodes.desc_signature):
+        for node in self.document.findall(sphinx.addnodes.desc_signature):
```
2bndy5, line 61: Is this to satisfy a deprecation?

sphinx_immaterial/apidoc/python/default.py
```diff
@@ -0,0 +1,27 @@
...
+from . import parameter_objects
+from . import style_default_values_as_code
+from . import subscript_methods
+from . import attribute_style
+from . import napoleon_admonition_classes
+from . import strip_property_prefix
+from . import section_titles
+from . import autodoc_property_type
+from . import type_annotation_transforms
+from . import strip_self_and_return_type_annotations
```
2bndy5, line 15: I assume some of these unused imports invoke the one-off patches without an ext `setup()`

2bndy5 (approved):
I also appreciate the refactor to bite-size modules. It makes reviewing much more palatable.

pytest snapshots looks like a huge time saver!

sphinx_immaterial/apidoc/format_signatures.py
jbms: Yes

sphinx_immaterial/apidoc/python/default.py
jbms: Yes that's correct.

jbms:
Thanks for the review!

sphinx_immaterial/apidoc/format_signatures.py
2bndy5: I suppose this won't impact us since we're pinned to whatever ver of docutils that sphinx installs. IIRC, that deprecation has been around since docutils v0.16, and sphinx v4+ never pinned to an earlier version of docutils.

jbms:
> Think I'll push v0.8.1

Sounds good

PR #143, opened 2022-08-13
Ensure tables don't exceed content area width

### LINKED ISSUE #141: not all tables are confined to article's width
I often use `csv-table` or `list-table` for creating tables, but I've noticed that not all tables are confined to the article section's width. When the window is narrow enough for the local Toc to be shown, sometimes I see a table bleed over:
image

Adding the CSS,
```css
.md-typeset table.data:not(.plain) {
  /* formerly `display: table;` */
  display: inline-block;
}
```
fixes this issue:
image

jbms:
Fixes #141.

### DIFF
--- a/src/assets/stylesheets/main/_typeset.scss
+++ b/src/assets/stylesheets/main/_typeset.scss
@@ -456,7 +456,14 @@ kbd {
 
   // Data table
   table.data:not(.plain) {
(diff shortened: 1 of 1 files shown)


2bndy5 (approved):
this indeed fixes #141

Issue #111, opened 2022-06-15
Google Fonts are loaded from remote server


jbms:
Currently, this theme references Google Fonts from the Google server (inherited from mkdocs-material).

However, that means even a local copy of the built documentation still requires network access to obtain the fonts.

It would be better if the fonts could somehow be fetched/cached when building the documentation so that the built docs have no outside dependencies.

2bndy5:
Google fonts are hosted from [their own dedicated repo](https://github.com/google/fonts). Downloading them all can add an upward of 600 MB to the distribution. And then there's the problem that Google seems to update them whenever they want without notice (and often not tagging the changes).

Luckily, there seems to be a project that provides the fonts via npm called [fontsource](https://github.com/fontsource/fontsource). Maybe that can be used to better solve this issue.

2bndy5:
Have you thought about taking this upstream? I wonder if squidfunk would be keen on the idea. There's still the fetching of  mermaid.js to consider.

jbms:
Yes, we wouldn't want to include all 600MB of fonts as part of the PyPI package.  I think it would be better if as part of the documentation build (as opposed to the sphinx-immaterial package build), the fonts selected by the user were downloaded if not already cached and copied into the output directory.

Possibly mkdocs-material would also be interested in solving this issue, although I think they are less concerned about references external resources.  Also since the solution would happen at documentation build time, it would be specific to sphinx or mkdocs.

squidfunk:
[We already solved this 4 months ago](https://squidfunk.github.io/mkdocs-material/setup/ensuring-data-privacy/#built-in-privacy-plugin)! 😊

2bndy5:
We seem to building a growing list of features available to insiders; #61, #58, #56, and now this -- not a bad sign for us, but a competitive edge vs upstream public releases.

---

Unfortunately, I doubt we can solve this with an existing Sphinx ext because this theme is a rather unorthodox Sphinx theme (to keep faithful to upstream src).

squidfunk:
Linked content tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#linked-content-tabs) have been released with [8.3.0](https://squidfunk.github.io/mkdocs-material/changelog/#8.3.0) when the last funding goal was reached.

2bndy5:
fixed in v0.9.0

PR #144, opened 2022-08-13
Add graphviz integration

jbms:


2bndy5:
need to add graphviz to the readthedocs.yml dependencies:
```diff
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -12,6 +12,9 @@ build:
     python: "3.9"
     nodejs: "16"

+  apt_packages:
+    - graphviz
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
```

jbms:
> need to add graphviz to the readthedocs.yml dependencies:
> 
> ```diff
> --- a/.readthedocs.yaml
> +++ b/.readthedocs.yaml
> @@ -12,6 +12,9 @@ build:
>      python: "3.9"
>      nodejs: "16"
> 
> +  apt_packages:
> +    - graphviz
> +
>  # Build documentation in the docs/ directory with Sphinx
>  sphinx:
>    configuration: docs/conf.py
> ```

Thanks, just added that.

2bndy5:
xrefs in dot syntax?! I'll never use another theme again!

A while ago, I looked into changing the colors of a graph based on CSS vars, but I gave up when I saw the rendered output used a `iframe` element in the DOM.

jbms:
> xrefs in dot syntax?! I'll never use another theme again!
> 
A bit hacky but I think it works well enough.  This part could be upstreamed into the sphinx.ext.graphviz extension itself.

> A while ago, I looked into changing the colors of a graph based on CSS vars, but I gave up when I saw the rendered output used a `iframe` element in the DOM.

Yeah there are many ways to embed the graphviz output, but just inlining the svg directly into the HTML provides the most control.

jbms:
One difficulty with the fonts is that graphviz needs the font to be available at build time so that it can properly compute the size of the labels.  This specifies the same font used by the theme, but it is hard to ensure that font will be available locally.  Per #111 it would be nice to download the font as part of the build anyway, but the additional challenge is that graphviz seems to get the fonts in several different ways depending on how it is built.

2bndy5:
> inlining the svg directly into the HTML provides the most control

I think there was a reason they chose to use an `iframe` in upstream. It was a little more than a while ago since I looked into it.

> Per #111 it would be nice to download the font as part of the build anyway, but the additional challenge is that graphviz seems to get the fonts in several different ways depending on how it is built.

I was thinking the same thing.

docs/test.dot
```diff
@@ -0,0 +1,11 @@
+digraph D {
+
+  A [shape=diamond]
+  B [shape=box]
+  C [shape=circle]
+
+  A -> B [style=dashed]
+  A -> C [color="black:invis:black"]
```
2bndy5, line 8: Would be good to have a comment indicating the color is not dynamic to the color scheme of the theme:
```suggestion
  A -> C [color="black:invis:black"] // edge using a static color with transparent fill
```
At first, I thought the static color was a bug

docs/index.rst
```diff
@@ -74,6 +74,7 @@ or ``theme.conf`` for more details.
     admonitions
     content_tabs
     mermaid_diagrams
+    graphviz
```
2bndy5, line 77: I think it would less cluttered if we group these docs together under a "Graphs" `toctree`.

2bndy5 (changes requested):
First impressions are amazing. I'll try to play with this on Windows to see if there's an easy solution to the font problem (I have graphviz installed via the official downloads - not via npm).

docs/index.rst
jbms: To be clear, I think the font issue also exists on Linux and macOS --- it isn't customary to have the Google fonts available as system fonts.  I think it is just that the fontconfig library always picks some font if the requested font can't be found and doesn't warn about it.  In contrast, on Windows graphviz does warn, and we are treating warnings as errors in this theme's docs.

Note that for svg rendering it is kind of subtle --- at build time the font is only used to determine the size of the label.  The svg file will just specify whichever font name was requested, even if it is not available at build time.  So as long as the font is available to the browser, the labels will be displayed with the correct font.  But the layout may have been computed with the wrong sizes, which in most cases may not be very noticable.

jbms:
> > inlining the svg directly into the HTML provides the most control
> 
> I think there was a reason they chose to use an `iframe` in upstream. It was a little more than a while ago since I looked into it.

`sphinx.ext.graphviz` doesn't use an iframe, but it does embed the svg as an `<object>` which prevents links from linking.  I think it might be for older browser support (https://caniuse.com/?search=svg), or simply that it was implemented before inline svg was supported. In any case it seems inline svg is supported by all common browsers for several years.

src/assets/stylesheets/main/integrations/_graphviz.scss
```diff
@@ -0,0 +1,19 @@
+// All definitions
+:root > * {
+  // Colors
+  --md-graphviz-edge-color:             var(--md-default-fg-color--light);
```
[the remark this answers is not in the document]
jbms: Thanks, Done.

docs/test.dot
jbms: Done.

docs/index.rst
jbms: > I think it would less cluttered if we group these docs together under a "Graphs" `toctree`.

Done.

sphinx_immaterial/graphviz.py
```diff
@@ -0,0 +1,305 @@
...
+        "-Gfontname=" + font,
+    ]
+
+    code = re.sub(r'"((?:var|calc)\s*\(.*?\))"', replace_var_in_code, code)
+
+    graphviz_dot = options.get("graphviz_dot", self.builder.config.graphviz_dot)
+    dot_cmd = [graphviz_dot]
+    dot_cmd.extend(command_line_options)
+    dot_cmd.extend(self.builder.config.graphviz_dot_args)
+    dot_cmd.append("-Tsvg")
```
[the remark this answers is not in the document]
jbms: Done.

jbms:
I have now added support for automatically downloading the fonts, both to fix #111 and also so that they can provided in TTF format to Graphviz.

2bndy5:
This may be off topic, but I figured I'd mention it now that this aims to solve #111. I think the fetching of mermaid.js should be added to `external_resource_cache`. Currently, the `sphinx_immaterial.mermaid_diagrams` ext is on by default because the theme src seems to be fetching the mermaid.js resource regardless.

jbms:
> This may be off topic, but I figured I'd mention it now that this aims to solve #111. I think the fetching of mermaid.js should be added to `external_resource_cache`. Currently, the `sphinx_immaterial.mermaid_diagrams` ext is on by default because the theme src seems to be fetching the mermaid.js resource regardless.

Agreed, and also for MathJax.  But that could be deferred for a separate PR.

2bndy5:
This runs fine locally on my Windows machine. Maybe there's an extra step that we need to do in CI to make it work like an end-user machine (like using a venv to ensure PATH env var is adequate).

I get no warnings in local (fresh and consecutive) builds. Fetching the fonts only happens on fresh doc builds (as expected). Just for completeness, this is what I see when the fonts are fetched:
> ```
> Fetching: https://fonts.gstatic.com/s/roboto/v30/KFOlCnqEu92Fr1MmSU5fBxc4EsA.woff2 with {}
> ```

The commit you're using to install graphviz: ts-graphviz/setup-graphviz@f1590cf4e989dd1429bff4fd6339e098050a1657 seems a bit our of date. That action has seen [a release 2 days ago](https://github.com/ts-graphviz/setup-graphviz/releases/tag/v1.1.0) that would allow specifying a specific version of graphviz for different OS runners, but I'm not sure if that is helpful.

2bndy5:
For the record, that last comment was running graphviz v2.50. I just installed graphviz v5.0, and it still ran fine (from a fresh doc build).

2bndy5:
This error may actually be related to the contents of the dot file. Our .gitattributes doesn't enforce LF for dot files... Maybe the dot exe expects LF, but the Windows runner checks out the dot file with CRLF?

Additionally, I'd like to propose changing the dot file to use a radial layout to distinguish the advantage of using graphviz over mermaid.js (aside from latex support and well-seasoned age).

Can I push a commit to this branch?

jbms:
> This error may actually be related to the contents of the dot file. Our .gitattributes doesn't enforce LF for dot files... Maybe the dot exe expects LF, but the Windows runner checks out the dot file with CRLF?

I'll have to look into it more.  It appears that dot exited with return code of 0, but generated non-empty stderr after stripping whitespace, which is shown as a warning, but the output looks empty in the logs.  Maybe there is a special character getting in somehow.

> 
> Additionally, I'd like to propose changing the dot file to use a radial layout to distinguish the advantage of using graphviz over mermaid.js (aside from latex support and well-seasoned age).

In general I think graphviz is more powerful and much more standard, but mermaid has some nice built-in graph styles.  I wonder if there is a mermaid to dot converter.

> 
> Can I push a commit to this branch?

Feel free to push commits.

sphinx_immaterial/graphviz.py
```diff
@@ -0,0 +1,483 @@
...
+            pathlib.Path(
+                os.path.join(tempdir, os.path.basename(config_info.orig_config_path))
+            ).write_bytes(config_info.new_config)
+
+            env["GVBINDIR"] = tempdir
+            if os.name == "nt":
+                _winapi.CreateJunction(orig_lib_dir, new_lib_dir)  # type: ignore[attr-defined]
+            else:
+                os.symlink(orig_lib_dir, new_lib_dir)
+            cwd = os.path.dirname(config_info.orig_config_path)
```
2bndy5, line 287: There's no reason (that I can see) to create a junction on Windows. My local tests show
```py
            orig_lib_path = pathlib.PurePath(config_info.orig_config_path)
            new_lib_dir = pathlib.Path(tempdir, "plugins")
            pathlib.Path(tempdir, orig_lib_path.name).write_bytes(
                config_info.new_config
            )

            env["GVBINDIR"] = tempdir
            new_lib_dir.symlink_to(orig_lib_path.parent, target_is_directory=True)
            cwd = str(orig_lib_path.parent)
```
working fine on windows (the `target_is_directory` param is ignored on other Posix systems).

I don't think this is the breaking point since it gets through this fine when the dot syntax is embedded. I'm just trying to isolate anything specific to Windows...

2bndy5:
I'm not entirely sure the font is being used correctly (on Windows) to calculate the sizes.

| RTD output | Local (Windows) output |
|-------------|-------------------------|
| image | image |
| image | image |

sphinx_immaterial/graphviz.py
2bndy5: I'm not sure if you're aware, but the dot exe is in the same dir as the plugin DLLs (by default that's `C:\Program Files\Graphviz\bin`). I say this because I noticed they're in separate locations on Linux (or at least for my WSL install of graphviz).

I don't know if that means anything though.

2bndy5: I added the `-v` arg to dot exe calls ,and I'm seeing:
```
dot - graphviz version 5.0.0 (20220707.1540)
libdir = "abs_path\to\AppData\Local\Temp\tmp88gcajyy"
Activated plugin library: plugins\gvplugin_core.dll
Using render: svg:core
Using device: svg:svg:core
Activated plugin library: plugins\gvplugin_dot_layout.dll
Using layout: dot:dot_layout
The plugin configuration file:
        abs_path\to\AppData\Local\Temp\tmp88gcajyy\config6
                was successfully loaded.
    render      :  dot dot_json fig gdiplus json json0 map mp pic pov ps svg tk visio vml xdot xdot_json 
    layout      :  circo dot fdp neato nop nop1 nop2 osage patchwork sfdp twopi
    textlayout  :  textlayout
    device      :  bmp canon cmap cmapx cmapx_np dot dot_json emf emfplus eps fig gif gv imap imap_np ismap jpe jpeg jpg json json0 metafile mp pic plain plain-ext png pov ps ps2 svg tif tiff tk vdx vml xdot xdot1.2 xdot1.4 xdot_json
    loadimage   :  (lib) bmp eps gif jpe jpeg jpg png ps svg
Activated plugin library: plugins\gvplugin_neato_layout.dll
Using layout: twopi:neato_layout
fontname: unable to resolve "abs_path\to\docs\_build\html\_static\fonts\3254c528e2ab56454a9f22191035c5fe.ttf"
root = A max steps to root = 1
Rank separation = 0.000 1.000
overlap: none value 0 scaling 0.0000
Edge separation: add=1 (3.200000,3.200000)
Creating edges using line segments
Using render: svg:core
Using device: svg:svg:core
gvRenderJobs %3: 0.00 secs.
```
Notice it is still unable to resolve the fonts. The path output is correct, but I've shortened it here.
Unsuccessful attempts:
- I tried using posix style path separators
- I tried using `file:///` URI
- I set the `DOTFONTPATH` env var to `ttf_font`'s folder and only passing the font file's name (with and without suffix) to the dot exe args.

jbms:
The reason for using a junction of windows is that my understanding is that symlinks aren't supported on most/all versions of windows by default, they have to be explicitly enabled globally first.  In contrast, junctions are always available.  However, I think there might be some restrictions on junctions across filesystems that don't apply to symlinks, so there could still be some advantage in using symlinks on Windows when supported.

sphinx_immaterial/graphviz.py
jbms: Seems like maybe the libgd-based text layout still isn't being used --- perhaps the relevant plugin to disable on Windows is different?

2bndy5:
> they have to be explicitly enabled globally first

I've never had to do this. It might be true for peculiar editions of Windows (like "enterprise" or whatever they call the build that gets sold to corporate entities).

> I think there might be some restrictions on junctions across filesystems that don't apply to symlinks

This might be what was failing in CI because the github runner's use docker containers.

sphinx_immaterial/graphviz.py
2bndy5: CI passes when using symlinks!

jbms: Great that it works with symlinks --- still unclear what the issue with junctions was.  A bit surprising that there would be no error in that case, although _winapi being an undocumented module may lack proper error handling.

It appears that "developer mode" has to be enabled for unprivileged users to create symlinks: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinka?redirectedfrom=MSDN

2bndy5: >  _winapi being an undocumented module may lack proper error handling.

It definetly doesn't have error checking. In searching for the module's docs, I came across a post from a python core dev that said [creating junctions] wasn't actually used by python std libs and it is only there for tests despite the fact that it doesn't check for things like target existence.

> It appears that "developer mode" has to be enabled for unprivileged users to create symlinks

I must have enabled it when installing Visual Sudio at some point. Not sure how this will impact end-users of this extension. Win8 required development mode for people that wanted to install apps without using the Windows store (win10 or newer has sideloading enabled by default because win8 was the worst _ever_). I'm not too worried about this because people developing docs with graphs are often using it to document an app that they support (thus developer mode should already be enabled).

2bndy5: Should probably add a warning to the docs about symlinks on Windows. IDK about MacOS because Apple doesn't deserve my money.

2bndy5: As a workaround (for this config stuff workaround), we could create 1 temp dir that has copies of the needed DLL files and avoid symlinks all together. Currently, it seems a little repetitive to create a temp dir and symlink to installed DLL for each use of the `graphviz` directive.

Despite what I said in https://github.com/jbms/sphinx-immaterial/pull/144#discussion_r950682756, the DLLs and config file seem to always be located in the same dir - not surprising given what the Graphviz docs say about the `GVBINDIR`  env var.

jbms:
Are you still observing bad font metrics in your local build on Windows?

2bndy5:
Yes. I'm not sure how to get graphviz to resolve the font from the given path. See https://github.com/jbms/sphinx-immaterial/pull/144#discussion_r950749998

jbms:
The problem appears to be that the normal build of Graphviz for Windows excludes the "libgd" plugin, while on Linux builds it is normally excluded.  Instead, on Windows there is a gdiplus plugin that provides text layout based on the Windows GDI+ API.

It looks like there are two possible ways to inject custom fonts:

- We can copy the font to `%HOME%\AppData\Local\Microsoft\Windows\Fonts` but it is a bit unfortunate to touch the user's font directory.
- There is also this function call: https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-addfontresourceexa  Unfortunately, that has to be invoked from within the graphviz process itself; to accomplish that we could create a dll and then while the graphviz process is waiting for the graph input, using a DLL injection technique (https://en.wikipedia.org/wiki/DLL_injection) to inject it, probably the `CreateRemoteThread` one.

2bndy5:
I would first try using the AppData approach. Instead of using `%HOME%`, I think it'd be better to use `%USERPROFILE%` which is the older equivalent to `~` on Linux.

I've never compiled anything with the GDI lib, so that seems like an interesting challenge (assuming you need me for my windows platform). The DLL injection sounds more like an exploit than a hack though.

jbms:
I filed this issue with graphviz: https://gitlab.com/graphviz/graphviz/-/issues/2267

It sounds like the issue is that they are relying on some pre-built binaries for all of the dependencies on Windows, and are missing libgd for x86_64 Windows, so it is excluded.  Ideally they would setup a continuous build for graphviz that builds all of the dependencies from source, at least on Windows where they have to be bundled.  It looks like currently there is no automated process for building the dependencies; the binaries were just manually obtained somehow at some point and rarely updated.

Regarding adding the fonts the the AppData directory, I'm somewhat hesitant to go that way because then we will be persistently installing a font for the user, which is an unexpected outcome of just doing a documentation build.  I believe the DLL injection is sometimes used for exploits, but here I think of it as essentially just monkey patching taken to the next level.

2bndy5:
I was thinking about raising an issue over there too. It shouldn't be this hard to specify a font.

> It looks like currently there is no automated process for building the dependencies; the binaries were just manually obtained somehow at some point and rarely updated.

The [docs say that the dev env for Windows build uses submodules](https://graphviz.org/doc/winbuild.html), but I haven't tried to build it myself.

---
So, I tried manually copying the Roboto.ttf font into my `%USERPROFILE%/local/Microsoft/Windows/Fonts`. But, I'm still getting the "unable to resolve font" in verbose output. I did narrow down the source of that prompt to `-Nfontname` option BTW.

2bndy5:
I'm ok with merging this as long as there's a warning about experimental/developmental support for Windows. The dev cycle for graphviz seems sparse, and I don't see a lot of feature or patch releases over the last few major versions.

As it is now the docs build on Windows and use the correct font in HTML (despite the inaccurate node sizes).

sphinx_immaterial/graphviz.py
```diff
@@ -0,0 +1,509 @@
...
+        config_content,
+    ):
+        if m.start() != prev_index:
+            parse_error()
+            return None
+        prev_index = m.end()
+        plugin_path = m.group(1)
+        plugin_name = m.group(2)
+        if plugin_name == b"cairo":
+            continue
```
2bndy5, line 191: What happened to excluding pango? I thought that was a problematic default as well.

jbms: The path to the plugin includes "pango" but in the config the plugin name is called "cairo".  I think the plugin makes use of both of those libraries.  In any case that line was stale --- instead it now just excludes the `textlayout` functionality from everything except LibGD.  On Windows there is a GDI+ textlayout plugin that we also have to exclude.

jbms:
This still doesn't correctly handle the fonts on Windows, but the option to suppress the warning there works correctly and CI passes.

2bndy5:
Is there anything about external_resource_cache.py that should be doc'd? It seems automatic enough to leave out as long as users specifies a font from the suggested google fonts.

2bndy5 (approved):

2bndy5 (approved):

docs/graphviz.rst
```diff
@@ -0,0 +1,131 @@
...
+   This extension relies on the LibGD graphviz plugin to load the same font used
+   by this theme, in order to correctly determine the size of text labels.
+   While the LibGD plugin is normally included in the Linux and macOS graphviz
+   builds, the official x86_64 Windows build does not include it.  If the plugin
+   is not found, by default this theme logs a warning.  This option may be set
+   to :python:`True` silence that warning.
+
+   .. rst-example:: Add to :file:`conf.py` to silence the warning
+
+      graphviz_ignore_incorrect_font_metrics = True
```
2bndy5, line 51: I missed this. Probably didn't mean to use rst-example
image

PR #149, opened 2022-08-17
Python apigen: Ensure correct `module` is specified

### LINKED ISSUE #147: Incorrect object names for inherited members from undocumented classes
I noticed that documenting undocumented parent methods has an issue with the object name and cross-linking. Here is a self-contained reproducer [foo-147.zip](https://github.com/jbms/sphinx-immaterial/files/9355539/foo-147.zip).

`AbstractBase` is not documented / is not a public class in the `foo` package.

```python
# foo/__init__.py
from ._bar import B

# foo/_bar.py
class AbstractBase:
    @property
    def something(self) -> str:
        """
        The base docstring that is applicable to all children.

        This is the second paragraph.
        """
        pass

class B(AbstractBase):
    pass
```

Notice `something` is white and unlinked.

image

Notice the URL is correct as `api/foo.B.something/`, however the object name is `foo._bar.foo.B.something` not `foo.B.something`.

image

jbms:
Fixes #147.

### DIFF
--- a/sphinx_immaterial/apidoc/python/apigen.py
+++ b/sphinx_immaterial/apidoc/python/apigen.py
@@ -1505,6 +1505,8 @@ class _ApiEntityCollector:
 
         split_result = _split_autodoc_rst_output(rst_strings)
 
+        split_result.options.pop("module", None)
+
         overload_id: Optional[str] = None
         if entry.overload is not None:
             overload_id = entry.overload.overload_id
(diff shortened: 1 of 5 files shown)


mhostetter:
Thank you @jbms. I can confirm this resolves #147 for me. 👍

2bndy5 (approved):

PR #166, opened 2022-09-30
update typing related to the cpp.apigen ext

2bndy5:
Looks like the latest release of Sphinx (v5.2.2) brought with it a few "tweaks" to their type hints.

This addresses just that.


jbms:
Thanks!

PR #167, opened 2022-09-30
add `social` icons theme option to docs

2bndy5:
- add `social` option to theme.conf
- add description of `social` option to customization.rst
- also adds common venv folders to gitignore which makes it easier to use an isolated venv on local dev machine.

Special thanks to @mhostetter for bringing this to my attention!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,10 @@ __pycache__/
 /*.egg-info
 /dist
 
+# python virtual env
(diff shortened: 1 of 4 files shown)


jbms:
Thanks!

2bndy5:
Think I'll push v0.9.1 to fix the warnings we weren't aware of in current release.

Issue #61, opened 2022-04-14
Linked content tabs


2bndy5:
Well, it looks like linked content tabs will remain an insiders feature only. Originally, I thought the feature would get included in the public releases when the upstream funding goal hit $5000+, but the linked tabs feature wasn't made public when that goal was recently met.

So, we have to implement linked tabs ourselves.

### Implementation

I envision this as a `:linked:` option to the `md-tab-item` directive. This could be a flag (option accepts no value) or an option that expects an identifying string.
- Expecting a string value would align with sphinx-design implementation and allow for linked tabs to have different labels.
- A flag might be easier from the user's perspective (relying on commonly shared tabs' labels), but that may be error prone. For example, case sensitivity or non-text elements in the linked tabs' labels.

jbms:
Should the `:linked` option be just on the `md-tab-item` directive or on the `md-tab-set` directive as well?

I can see how it would be useful to specify an identifier for a non-text label (although for accessibility it would probably be good to have a text label as well even if you want to display an icon), or as an override, but it seems like in almost all cases you would just want to mark an entire tab-set as linked (and use the text of the tab label as the key).

In the same case, you link multiple tab sets that all have the same set of labels.  In that case it is pretty obvious what the behavior should be --- their selected tabs should remain the same at all times.

However, it is a bit trickier if some tabs are linked but not others.

For example, suppose the tab labels correspond to programming language, but not all tab sets include all programming languages.  E.g. first tab set has "C", "Java", "Python"  and the second one has "Java", "Python", "JavaScript".  Presumably if we select "Java" or "Python" in one we want it to be selected in the other.  If we select "C" in one --- presumably the other remains unchanged.  So we would need to choose a representation in `localStorage` (https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) that accommodates that.  Perhaps store the last-selected-timestamp for each linked tab key.

2bndy5:
Glad to see you're already thinking about the implementation details. Since I'm not a JS expert, I'd probably look for inspiration from similar implementations (like sphinx-design ext).

Just to throw it out there, [sphinx-design also supports a special directive](https://sphinx-design.readthedocs.io/en/furo-theme/tabs.html#tabbed-code-examples) aimed at linking tabs based on specified syntax.

I think we can support a `linked` option with and without a specified value: 
1. without a specified value relies on the tab's label (`astext()`)
[...]

Issue #169, opened 2022-10-08
back-links in viewcode should float right?


2bndy5:
I've been noticing that the links in the literal blocks generated by the viewcode ext are oddly placed at the beginning of the line with the xref'd signature. For example:
![image](https://user-images.githubusercontent.com/14963867/194731532-1ab38166-832f-4bcc-a0d9-2ceec5bd9ff0.png)
leads to
![image](https://user-images.githubusercontent.com/14963867/194731548-7d97d437-07ee-4f56-a056-da46ca7207d4.png)

Most other themes float the `[docs]` link to the right in a non-obstructive way, so I've been using custom CSS do achieve this:
```css
.viewcode-block .viewcode-back {
  float: right;
}
```
which results in
![image](https://user-images.githubusercontent.com/14963867/194731586-f63898a2-d351-408a-9656-56a4fc704f22.png)

I was wondering if the CSS I'm using should be added to this theme's CSS.

jbms:
Yeah that sounds like a good idea.

2bndy5:
FYI, I also intend to do a merge from upstream soon (which should help resolve #61)

jbms:
> FYI, I also intend to do a merge from upstream soon (which should help resolve #61)

That would be great!

PR #171, opened 2022-10-20
Merge upstream v8.5.6

2bndy5:
[patch_info.zip](https://github.com/jbms/sphinx-immaterial/files/9827216/patch_info.zip)

This wasn't as painful as I had expected (due to the long time since last update from upstream), or maybe I'm just getting better at it (wishful thinking 🤣 )

Aside from the usual JS tweaks after the merge script did it's thing, I also
- documented linked content tabs (addresses #61 )
- documented how to change the "Edit this page" icon

I had to tweak the merge script for files that are explicitly excluded (in `MKDOCS_EXCLUDE_PATTERNS`) and are no longer in the upstream base/target git indexes. There's still room for improvement because I had to manually add the new files from upstream since they weren't in this repo's index. The additions were mostly language translations and a couple features implemented on upstream only (like builtin site feedback and comments). Maybe there's a way to port the new functionality to sphinx builds, but I'm not overly enthusiastic about them (currently content with directing to other sites like github for a more dynamic user interactive experience).

I did discover 1 regression about mermaid graphs and filed an issue upstream (https://github.com/squidfunk/mkdocs-material/issues/4517). We can merge in the simple fix for that later though.


docs/content_tabs.rst
```diff
@@ -100,7 +89,80 @@ the sphinx-immaterial theme provides its own directives to make use of content t
...
+
+.. code-block:: python
+    :caption: in conf.py
+
+    html_theme_options = {
+        "features": [
+            "content.tabs.link",
+        ],
+
+Include :python:`"navigation.instant"` in the list of :themeconf:`features` to have the
```
jbms, line 108: `navigation.instant` shouldn't be needed if this is working properly; if it doesn't, that suggests a problem with the use of localStorage --- also in my experience `navigation.instant` is kind of buggy and doesn't provide any significant benefit so I'm not sure we want to recommend it

src/assets/javascripts/_/index.ts
```diff
@@ -44,6 +47,7 @@ export type Flag =
   | "search.highlight"                 /* Search highlighting */
   | "search.share"                     /* Search sharing */
   | "search.suggest"                   /* Search suggestions */
+  | "toc.follow"                       /* Following table of contents */
```
jbms, line 50: Duplicate of below

jbms: Probably should remove the one below instead to reduce merge conflicts in the future

src/assets/javascripts/_/index.ts
```diff
@@ -93,6 +97,8 @@ export interface Config {
   base: string                         /* Base URL */
   features: Flag[]                     /* Feature flags */
   translations: Translations           /* Translations */
+  search: string                       /* Search worker URL */
```
jbms, line 100: we don't use these bug I guess they could be kept.  On the other hand it might be better to exclude them so that we catch at compile time any attempt to use them

src/assets/javascripts/_/index.ts
```diff
@@ -31,7 +31,10 @@ import { Version } from "~/templates/version"
  * Feature flag
  */
 export type Flag =
+  | "announce.dismiss"                 /* Dismissable announcement bar */
```
jbms, line 34: are any of these other features working / should be documented?

src/assets/javascripts/components/toc/index.ts
```diff
@@ -296,7 +296,7 @@ export function mountTableOfContents(
 
       /* Look backward */
       for (const [index, [anchor]] of prev.entries()) {
-        anchor.setAttribute("data-md-state", "blur")
```
jbms, line 299: Need to change `anchor.removeAttribute("data-md-state")` above to `anchor.classList.remove("md-nav__link--passed")`

src/assets/javascripts/templates/search/index.tsx
```diff
@@ -93,9 +94,21 @@ function renderSearchDocument(
             {truncate(document.text, 320)}
           </p>
         }
-        {document.tags && document.tags.map(tag => (
-          <span class="md-tag">{tag}</span>
-        ))}
+        {document.tags && (
```
jbms, line 97: I guess we could see about supporting tags?

src/assets/javascripts/_/index.ts
2bndy5: I haven't tested them yet. I just figured it would hurt to have them while we figure it out.

src/assets/javascripts/components/toc/index.ts
2bndy5: good catch!

docs/content_tabs.rst
2bndy5: I quick copied and pasted the linked tabs example into a separate page and verified it worked with the feature specified in conf.py. I didn't think to try without it since the mkdocs-material doc mandated it be specified. I'll re-check without it explicitly enabled.

src/assets/stylesheets/main/layout/_nav.scss
```diff
@@ -113,8 +113,8 @@
         0 #{$margin-height} var(--md-default-bg-color);
     }
 
-    // Navigation link in blurred state
-    &[data-md-state="blur"]:not(&--in-viewport):not(&--active) {
+    // Navigation link that was passed
+    &--passed {
```
jbms, line 117: Not sure if this is correct --- will have to test after the other issue I mentioned related to `--passed` is fixed.

src/partials/feedback.html
```diff
@@ -0,0 +1,85 @@
...
+-->
+
+<!-- Determine feedback configuration -->
+{% if config.extra.analytics %}
+  {% set feedback = config.extra.analytics.feedback %}
+{% endif %}
+
+<!-- Determine whether to show feedback -->
+{% if page.meta and page.meta.hide %}
+  {% if "feedback" in page.meta.hide %}
```
jbms, line 30: I guess we need to add a way to set this?

src/partials/footer.html
```diff
@@ -26,9 +26,13 @@
 
   <!-- Link to previous and/or next page -->
   {% if page.previous_page or page.next_page %}
+    {% if page.meta and page.meta.hide %}
+      {% set hidden = "hidden" if "footer" in page.meta.hide %}
```
jbms, line 30: perhaps we need to add a way to set this?

src/partials/icons.html
```diff
@@ -0,0 +1,58 @@
...
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+  IN THE SOFTWARE.
+-->
+
+<!-- Custom admonition icons -->
+{% if config.theme.icon.admonition %}
```
jbms, line 24: kind of unfortunate implementation here since it is duplicated in every page rather than put in the css bundle, but I guess better than nothing...

src/partials/footer.html
2bndy5: Yeah, I think another meta data tag like our other `:hide-*:` mechanisms would suffice.

src/partials/feedback.html
2bndy5: That would take me a while because I'm not well versed in site analytics. Here's hoping it is a simple patch to our current implementation of the analytics field in `html_theme_options`.

src/assets/stylesheets/main/layout/_nav.scss
2bndy5: I allowed this because I noticed the change in [src/assets/javascripts/components/toc](https://github.com/jbms/sphinx-immaterial/pull/171/files#diff-ed6ae506600ba3c7de17fba6b64bc6a7f67a571894f3f73ce06ea04c63cb6b66). Are you specifically concerned about the `:not(&--in-viewport)`?

I didn't allow the change to the `toc.follow` feature (detailed in the patch_info.diff located in OP for this PR).

src/assets/javascripts/_/index.ts
jbms: yeah sounds reasonable.

src/partials/feedback.html
jbms: Sorry, I just meant a way to add it to page.meta.hide, e.g. via a `:hide-feedback:` meta option or something --- note the existing code in nav_adapt.py that modifies `page["meta"]["hide"]`

jbms:
Thanks for doing this upgrade!

src/assets/javascripts/_/index.ts
2bndy5: I'm ok with excluding them. If someone reports a problem, we can advise against using them or re-assess.

src/assets/stylesheets/main/layout/_nav.scss
jbms: Yeah, I'm not sure if the other selectors can be removed or if they are still required --- will need to test once you have pushed out a fix for the other `--passed` change.

2bndy5: pushed that `--passed` change.

jbms (approved):

2bndy5:
If you don't mind I'd like to keep this open for testing and further needed documentation about newer features. I'm about to push changes to docs about customizing the admonition icons. I also intend to patch the `hide-feedback` and `hide-footer` meta tags.

src/assets/stylesheets/main/layout/_nav.scss
2bndy5: If you've approved, then I'm guessing this can be resolved?

jbms:
> If you don't mind I'd like to keep this open for testing and further needed documentation about newer features. I'm about to push changes to docs about customizing the admonition icons. I also intend to patch the `hide-feedback` and `hide-footer` meta tags.

Sure, that's why I didn't merge it myself.

docs/content_tabs.rst
2bndy5: You're correct, `navigation.instant` isn't required for cross-page syncing of linked tabs. I have adjusted the doc here to express this difference from upstream.

Going forward, we should probably be careful about tab names. For example, a tab named "CSS hint" would be shown on page load if another tab of the same name was the last selected tab from a different section/page.

jbms: Agreed.  We also might wish to improve on the functionality provided by mkdocs-material to allow linking tabs more selectively, rather than an all or nothing option.  I also think in some cases it is used as an example for this theme in a way that might not be ideal for actual use.  My thinking is that tabs are best when there are true alternatives, where the user is unlikely to be interested in both options.  E.g. installation instructions for windows vs linux, or usage examples in Python vs C++.  If the user is likely to care about multiple tabs, then it may be better not to use tabs and instead just use sections.

src/assets/stylesheets/main/layout/_nav.scss
jbms: yes, I verified that it works correctly

docs/content_tabs.rst
2bndy5: It is true that I tend to use tabs as a way of demonstrating steps in an example. My intention there was to keep the page length from growing to an overwhelming size -- I have a bad habit of using my mouse wheel more than the toc entries' links.

Maybe we should open an issue/discussion upstream about specifying certain tabs sets as non-sync-able?

src/assets/javascripts/_/index.ts
2bndy5: `announce.dismiss` works well. I added a banner to our docs that declares the latest upstream version used (with a link to the upstream changelog). I also used the new HTML template as an example in the "New Blocks" section of the customization page. The `announce.dismiss` feature is enabled should we need to verify that it behaves consistently.

---
I'm not sure how to test the `content.lazy` feature since it is browser specific. So, I'm just leaving that flag in there in case we want to re-assess later.

src/assets/javascripts/templates/search/index.tsx
2bndy5: I'm leaving this as is for now. If someone asks, then we can discuss it again (or just create a new issue as a reminder).

src/partials/feedback.html
2bndy5: I think I found a bug (probably in Sphinx). The metadata keys don't get interpreted as metadata if the have a `-` in them. Rather, they just get interpreted as a field. I have tried all of our supported metadata tags, and they currently don't work. So, all our metadata keys should be changed to `hide_*` (thank goodnes we're still in beta).

I'm still digging, and I'll start rolling back the sphinx version I'm using (currently v5.3.0) to see where it was broken.

2bndy5: Well the feedback can be displayed as a demo using a dummy analytics provider (like `G-XXXXXXXXXX`), but it would mean an additional warning in the browser's dev console. I've done this locally to verify that the new `:hide_feedback:` metadata tag works.

I started to look into doing this with my Google account, but apparently all traffic to analytics.google.com domain is blocked by my DNS sinkhole (PiHole) 😅 .

For now, I think I'll just put an example snippet in the docs under the analytics option in customization.rst.

2bndy5:
I'm now ready to merge this. I'll let it stew for a day so you have another chance to review the documentation changes (namely in customization.rst and admonitions.rst).

2bndy5:
The `consent` feature does work, but it's designed to not trigger on local files viewed in the browser. 
1. I'm not sure how to document the cookies portion of the feature.
2. I'm not sure the upstream default cookies are enacted (for google analytics and github) though.

Here's the config I used to test it:
```py
html_theme_options = {
    "consent": {
        "title": "Welcome!",
        "description": "We don't use cookies, so your privacy is safe here.",
        "actions": ["accept"],
    }
 }
```
and then manually remove the `hidden` attr in html because I was viewing it from a file.

I think I'll just add the necessary src code to use it and leave the documentation for another time.

src/partials/feedback.html
2bndy5: > I'm still digging, and I'll start rolling back the sphinx version I'm using (currently v5.3.0) to see where it was broken.

Looks like the JSON domain relies on a module in sphinx that wasn't available in v4.3.0; we should probably update our requirements (or the optional deps for the json domain).

I couldn't find where it was broken. I tried rolling back sphinx (every version after v4.3.0) and docutils (as far back as v0.16), and the metadata tags weren't getting picked up with hyphens in the keys' name (using a fresh doc build each time).

jbms: I'm seeing some confusing behavior with regard to the meta fields: I'm not able to reproduce the issue with hyphens in the meta fields, and indeed the Python and C++ apigen modules emit `:hide-edit-link:` which is correctly respected.  However, I am seeing that in some cases more than the first meta field is being ignored, but it will need more investigation to figure out what is going on.

For example, after adding `if docname == "index": print(meta)` to `_html_page_context` in `nav_adapt.py`, if I change the start of `docs/index.rst` to:

```
:foo-bar: abc def
:hero: A responsive Material Design theme for Sphinx sites.
```

Then both are present in meta:

```
{'foo-bar': 'abc def', 'hero': 'A responsive Material Design theme for Sphinx sites.'}
```

but if I reverse the order at the top of `docs/index.rst` to:

```
:hero: A responsive Material Design theme for Sphinx sites.
:foo-bar: abc def
```

The only `hero` is present:

```
{'hero': 'A responsive Material Design theme for Sphinx sites.'}
```

2bndy5: Interesting, I didn't consider the order when diagnosing. I did try putting blank lines between the metadata in case the lack of a value associated with a metadata key may have been a problem.

2bndy5: The only thing I can find in docutils changelogs is that they added a new directive for inserting [`<meta>`](https://docutils.sourceforge.io/docs/ref/doctree.html#meta) elements in v0.18. I also see they stopped stripping multiple `-` from CSS class names (which I remember having to contend with when porting one of upstream features to a theme ext).

jbms: The problem seems to be related to having an `rst_prolog` specified in `conf.py`: the issue goes away when `rst_prolog` is removed.  It looks like the issue is in docutils but I'm still investigating.

jbms: Actually the bug is in Sphinx.  When sphinx prepends the prolog, it attempts to insert it after the docinfo, which it detects using a regular expression:

https://github.com/sphinx-doc/sphinx/blob/93215f0a64a2e27656ff2bfe6958a20b1e6b6e1e/sphinx/util/rst.py#L76

https://github.com/sphinx-doc/sphinx/blob/93215f0a64a2e27656ff2bfe6958a20b1e6b6e1e/sphinx/util/rst.py#L26

This regular expression does not correctly match the full syntax allowed for fields for multiple reasons, but in particular it uses "\w" to match the field name, which excludes hyphens.

I think the intent here is to allow `rst_prolog` to insert content into the document (rather than just directives like `.. role: python(code)`), without preventing the document from specifying metadata fields.

For our `rst_prolog`, since it does not generate any docutils nodes, it makes no difference if it is inserted before or after the metadata fields.  The problem occurs if it is inserted in the middle of the metadata fields: since `:hero:` matches `docinfo_re` but `:foo-bar:` does not, the prolog gets inserted after `:hero:`, which results in the metadata fields being parsed as two separate field lists, and docutils only treats the first one as docinfo fields.

2bndy5: Sounds like we should propose an amendment to regex in docutils.

jbms: I filed this bug:
https://github.com/sphinx-doc/sphinx/issues/10932

As a workaround, leaving an initial blank line at the start of file before the docinfo fields mitigates the issue; since the regex doesn't match blank lines, the prolog will be inserted before it, and not split the docinfo fields.

2bndy5: Ok I will revert the metadata key names back to using hyphens and test the workaround.

jbms:
Thanks, looks good to me.

2bndy5:
Great, I'll merge this and start/continue squashing open issues. I suppose we could tag this merger, but I'm unreasonably excited about code annotations (despite the implementation's current limits), and want to solve #169 before the next PyPI release.

src/assets/stylesheets/main/layout/_nav.scss
```diff
@@ -661,14 +676,18 @@
 
           // Show navigation link as title
           > .md-nav__link {
+            position: sticky;
+            top: 0;
+            z-index: 1;
```
2bndy5, line 681: This `z-index` caused a regression as reported in #189

PR #173, opened 2022-10-24
resolves #169

### LINKED ISSUE #169: back-links in viewcode should float right?

2bndy5:


2bndy5:
I can confirm this does what it was designed to: https://sphinx-immaterial--173.org.readthedocs.build/en/173/_modules/test_py_module/test.html#Foo

jbms:
Looks good to me, thanks.

PR #175, opened 2022-10-26
Minor doc updates

### LINKED ISSUE #148: Documentation on `version_json` is incorrect
The docs mention:

> When serving the docs from a webserver the `version_json` file is resolved relative to the parent directory that contains the sphinx builder’s HTML output.

I understood this as that the default (not setting `version_json`) corresponds to setting it to `"versions.json"`. That doesn't work however. One needs to set it to `"../versions.json"` instead.

2bndy5:
I found in the partials/social.html template that social links can have a customized tooltip using a `name` field. This adds the relevant info and demos its usage in the docs.

Also resolves #148 by removing a confusing adjective (per user feedback).

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -583,7 +583,7 @@ Configuration Options
         - The :python:`"icon"` field can be specifed as `any of the icons bundled with this theme`_
         - The :python:`"link"` field is simply the hyperlink target added to the icon specified.
         
-        Optionally, custom text used for the icon's tooltip by specifying a :python:`"name"`
+        Optionally, custom text may be specified for the icon's tooltip with the :python:`"name"`
         field. If the :python:`"name"` is not specified, then the domain of the :python:`"link"`
         is used automatically.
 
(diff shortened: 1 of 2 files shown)


docs/customization.rst
```diff
@@ -582,6 +582,10 @@ Configuration Options
 
         - The :python:`"icon"` field can be specifed as `any of the icons bundled with this theme`_
         - The :python:`"link"` field is simply the hyperlink target added to the icon specified.
+        
+        Optionally, custom text used for the icon's tooltip by specifying a :python:`"name"`
```
jbms, line 586: Optionally, custom text may be specified for the icon's tooltip with the :python:`"name"` field.

jbms:
Looks good to me other than the one comment.

docs/customization.rst
2bndy5, line 586:
```suggestion
        Optionally, custom text may be specified for the icon's tooltip with the :python:`"name"`
```

PR #176, opened 2022-10-27
Code annotations

### LINKED ISSUE #170: content.code.annotate feature not working
### What happened:

I'm trying to use the `content.code.annotate` feature, but I can't get it to work inside of a yaml code block. 
There is also almost no documentation for getting it to work inside sphinx rst.

I've dug into the issue, and it appears that the element that is being passed into the "findCandidateList" function is the `<div class="highlight" />`. The function accesses the next sibling element which is expected to be an ordered list element, but the div has an extra parent.

https://github.com/jbms/sphinx-immaterial/blob/a944f278cb7f1066482e59b9a17d3311624cd01e/src/assets/javascripts/components/content/code/_/index.ts#L98

If the code were to access the sibling of the **parent** of the container I believe this code would work correctly, but as of now it just returns null and exits.

Here's an example of my rendered code block:

image

### Environment

Sphinx: v5.1.1
sphinx-immaterial: 0.9.1

2bndy5:
resolves #170 

This started working as expected on a fresh install of the branch, so I think it is ready for review.


docs/code_annotations.rst
```diff
@@ -0,0 +1,166 @@
...
+    `code-block` immediately prior to this directive. The given list will not be rendered in the
+    generated output (even if there is no code snippet to annotate). Additionally, all annotations will be
+    excluded when using the "copy to clipboard" button for an annotated `code-block`; the annotated
+    comments' text will still be copied (if not `Hiding the annotated comment`_).
+
+    An error will be shown in the build log if the given list is not recognized as an
+    :duref:`enumerated list <enumerated-lists>`.
+
+    .. note::
+        When using this directive the :python:`"content.code.annotate"` does not need to be
```
jbms, line 20: When using this directive, :python:`"context.code.annotation"` ...

docs/code_annotations.rst
```diff
@@ -0,0 +1,166 @@
...
+            "features" : [
+                "content.code.annotate"  # (1)
+            ],
+        }
+
+    .. code-annotations::
+        1. .. admonition:: Obsolete
+               :class: failure
+
+               This has no special affect because the :rst:dir:`code-annotations` directive already
```
jbms, line 57: "This has no special effect"

jbms: enables the feature automatically

docs/code_annotations.rst
```diff
@@ -0,0 +1,166 @@
...
+               enables the feature accordingly.
+
+:duref:`Enumerated lists' <enumerated-lists>` markers in reStructuredText can be uniformed with the
```
jbms, line 60: "specified" instead of "uniformed"?   Since the markers in the code block need to specify the actual number, I'm not sure how helpful this suggestion is, though...

sphinx_immaterial/code_annotations.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@html_translator_mixin.override
+def visit_literal_block(
+    self: html_translator_mixin.HTMLTranslatorMixin,
+    node: nodes.literal_block,
+    super_func: html_translator_mixin.BaseVisitCallback[nodes.literal_block],
+):
+    #  we need to be sure that the annotated_list is the next node
+    next_element = node.next_node(nodes.Element, siblings=True, descend=False)
```
jbms, line 48: Move `next_element` definition below following `if` block so that it doesn't need to be duplicated

sphinx_immaterial/code_annotations.py
```diff
@@ -0,0 +1,79 @@
...
+            nodes.Element, siblings=True, descend=False
+        )
+    if isinstance(next_element, annotations_list):
+        try:
+            super_func(self, node)
+        except nodes.SkipNode:
+            pass  # we're not done yet
+        # Add annotate class to the highlighted div tag (generated by pygments), and
+        # remove the closing div tag (`[:-7]`) of the highlighted div's parent node.
+        self.body[-1] = self.body[-1][:-7].replace(
```
jbms, line 63: add comment (to clarify to reader) and assert here regarding content of `self.body[-1]` as check against incompatible changes to sphinx

docs/code_annotations.rst
2bndy5: Well it turns out that the annotations' identifying number (in the code block) is only related to the corresponding _ordered_ list item (`<ol><li>...`). The number used in the list markers doesn't seem to matter. I think this is why squidfunk changed the buttons to use a generic `+` instead of the list items' numbers.

For example, this should still work:
```rst
.. code-block:: py

    # (1)!
    # (2)!
.. code-annotations::
    3. The first
    #. The second
```

I think I'll play around and proofread a bit more. I want to avoid this confusion.

sphinx_immaterial/code_annotations.py
2bndy5: I'm not sure I understand this request. If I move this definition below the
```py
if isinstance(parent.children[0], nodes.caption)
```
block then it would break captioned code-blocks because it overwrites the purpose of the if block.

The `next_element` definition isn't exactly duplicated because the if block takes the `next_node()` of the parent (due to the presence of the caption's extra div parent).

2bndy5: I removed the unnecessary `parent` definition (really an artifact from my debugging `print` statements).

sphinx_immaterial/code_annotations.py
2bndy5: see changes in https://github.com/jbms/sphinx-immaterial/pull/176/commits/201f719be816467188da693a56168196260c67ea

I hope this satisfies the request

docs/code_annotations.rst
2bndy5: I have embellished this example into a separate section to talk about the ambiguity. The "Annotations buttons with numbers" section now refers this new section as rationality in why the behavior changed upstream.

2bndy5:
While I was tinkering, I added some functionality that lets the annotated list be rendered if not used (following upstream's behavior).

2bndy5:
I just noticed a CSS problem in upstream: the annotation button's content is off center when used within an admonition.

In our theme:
image

In upstream:
image

2bndy5:
Turns out we can't annotate rST snippets because of how the JS extracts the comments using pygments' CSS classes 💩 . The `.cp` class represents an rST comment which take the form of multiple `<span>` elements across multiple lines... I'd propose something upstream, but I'm not sure how "clean" this would be in JS.

MD comments don't seem to be supported at all in pygments (probably because they're actually embedded HTML syntax). So, they have a similar limitation upstream.

jbms:
It would be better to handle the annotations at build time using Python rather than with JavaScript, but that would require our own implantation of the feature which would be unfortunate.

2bndy5:
I'm reading up on the history of the feature, and I'm finding that it should be used sparingly... 
- https://github.com/squidfunk/mkdocs-material/discussions/3832#discussioncomment-2617205
- [not fixable issues with annotations](https://github.com/squidfunk/mkdocs-material/issues?q=is%3Aissue+annotations+is%3Aclosed+label%3A%22not+fixable%22)

2bndy5:
I came across a comment from squidfunk saying he prefers JS, and the python side only gets attention when needed. Which is kinda bittersweet for us because we get features to use for free, but it limits our desired implementation sometimes.

squidfunk:
> I came across a comment from squidfunk saying he prefers JS, and the python side only gets attention when needed. Which is kinda bittersweet for us because we get features to use for free, but it limits our desired implementation sometimes.

This is a bold claim. I'm not sure where this is coming from. Yes, I prefer JS/TS over Python as a language, but I still try to solve each problem with the optimal technology. The latest flagship feature, the blog plugin, is almost entirely written in Python with little JS added. Code annotations are implemented in JS, as they need JS anyway for positioning and I'm not sure where preprocessing the HTML would be of any help. If you preprocess the markup, you will end up with a broken state if the JS fails to mount due to network or author errors. The current implementation ensures that it works when printing, works without JS and with JS. Progressive enhancement is at the heart of Material for MkDocs.

2bndy5:
I'd be hard pressed to find the comment now, but I'm glad to be wrong about that.

It's true you need JS for the tooltip positioning, but we could parse the literal block in python with pygments to find the annotations and replace them (or the entire comment) with the buttons that tie into the JS - docutils is very flexible in this regard, not sure about mkdocs though. This approach may collide with the clipboard.js functionality as well.

TBH, I'm just glad to be able to offer this feature at all here. I had a lot of fun playing with it to create the [new doc](https://sphinx-immaterial--176.org.readthedocs.build/en/176/code_annotations.html#using-annotations). I used a "have fun and break things" approach, so [users can actually see what a mistake looks like](https://sphinx-immaterial--176.org.readthedocs.build/en/176/code_annotations.html#hiding-the-annotated-comment) before they skip reading all the paragraphs and breaking it themselves.

squidfunk:
I'd be interested to see how you pull this off preprocessing it during build time, keeping non-JS compatibility by following the principle of progressive enhancement. Feel free to create a PR upstream, but please note that I can only merge it if it actually improves the situation without losing any of the functionality we currently have. From my experience, I don't believe there's much room for improvement, but I'd be very happy to be proven wrong.

2bndy5:
It would dependent on using docutils' rST parser, so I'm not sure if it could be extended to mkdocs... I intend to start reading up on pymdownx to find better ways to contribute upstream (really excited about that new block syntax), but my passion strongly sides with Sphinx because it handles API parsing out of the box.

2bndy5:
@squidfunk FYI, Doxygen does something similar for annotating the src code (in their "source code browser" feature). All the tooltips there are hidden divs at the bottom of the page (not generated by JS), and they just use JS (actually JQuery) to move the div/tooltip into view (and un-hide it). This is how I'm envisioning the build-time approach.

2bndy5:
At the risk of sounding like a noob, I have to ask what you mean by "print" or "printing"?

Much of Doxygen's site generation is _very poorly designed_ Their v2 is supposed to allow use of template engines like django, but I have yet to see that released. 

Doxygen's annotations aren't customizable (for the most part), so they just use CSS to <sup>1</sup>animate the tooltip visibility and <sup>2</sup>fix most of the tooltip's contents relative to the div's position. Then its just a matter of where the mouse is within the viewport; they don't have a problem with the "bottom of the page scenario" (https://github.com/squidfunk/mkdocs-material/issues/3639).

As for the PDF writer, I think it might be easier to leave the `(1)` in the literal block and render the tooltip contents as usual (like what I see in that link). We haven't been focusing on Sphinx' Latex writer, so I'd be surprised if everything works for this port of the theme (the CI doesn't complain though).

2bndy5:
If you're curious to see the Doxygen approach rendered, I still use it for the [RF24 lib](https://nrf24.github.io/RF24/RF24_8h_source.html)s.

docs/code_annotations.rst
```diff
@@ -0,0 +1,196 @@
...
+               :class: failure
+
+               This has no special effect because the :rst:dir:`code-annotations` directive
+               automatically enables the feature.
+
+Arbitrary list markers
+----------------------
+
+The :duref:`enumerated lists' <enumerated-lists>` markers are arbitrary and do not directly
+correspond to the identifying numbers used in the annotated comment. Because of this arbitration,
```
jbms, line 61: arbitration means something else --- you could use "arbitrariness" instead, or perhaps a different phrasing.

Also in next paragraph.

sphinx_immaterial/code_annotations.py
jbms: Thanks

sphinx_immaterial/code_annotations.py
jbms: You could simplify this to:

```python
annotation_parent = node
if isinstance(node.parent.children[0], nodes.caption):
  annotation_parent = node.parent
next_element = annotation_parent.next_node(nodes.Element, siblings=True, descend=False)
```

jbms:
Thanks.  Looks good to me.

PR #177, opened 2022-10-27
Fix 164

### LINKED ISSUE #164: [theme's graphviz ext] xref not friendly with node labels that require HTML chars
I'm writing a graph that shows data structures using [record-based nodes](https://graphviz.org/doc/info/shapes.html#record).

Here's a sample without any xrefs:

[homie-topology](http://magjac.com/graphviz-visual-editor/?dot=digraph%20g%20%7B%0A%20%20%20%20graph%20%5B%0A%20%20%20%20rankdir%20%3D%20%22LR%22%0A%20%20%20%20%5D%3B%0A%20%20%20%20%22device%22%20%5B%0A%20%20%20%20%20%20%20%20label%20%3D%20%22%3Cf0%3E%20device%20%7C%20%3Cf1%3E%20attributes%20%7C%20%3Cf2%3E%20nodes%22%0A%20%20%20%20%20%20%20%20shape%20%3D%20%22record%22%0A%20%20%20%20%5D%3B%0A%20%20%20%20%22node%22%20%5B%0A%20%20%20%20%20%20%20%20label%20%3D%20%22%3Cf0%3E%20node%20%7C%20%3Cf1%3E%20attributes%20%7C%20%3Cf2%3E%20properties%22%0A%20%20%20%20%20%20%20%20shape%20%3D%20%22record%22%0A%20%20%20%20%5D%3B%0A%20%20%20%20%22property%22%20%5B%0A%20%20%20%20%20%20%20%20label%20%3D%20%22%3Cf0%3E%20property%20%7C%20%3Cf1%3E%20attributes%22%0A%20%20%20%20%20%20%20%20shape%20%3D%20%22record%22%0A%20%20%20%20%5D%3B%0A%20%20%20%20%22device%22%3Af2%20-%3E%20%22node%22%3Af0%0A%20%20%20%20%22node%22%3Af2%20-%3E%20%22property%22%3Af0%0A%0A%7D)
Click this picture to go to the graphviz playground. Notice that cells in a record need to use the `<symbol>` syntax to identify the cell within the node when drawing connecting edges.

What I'd like to do is have a the top cell in the record link to a python class, but the replacements from this theme's graphviz ext crash the dot executable.

The warning emitted in the build log includes all my env vars (which I have removed), but doesn't show any of the dot code that was passed.
<details><summary>This code causes the following warning</summary>

```rst
.. graphviz::

    digraph g {
        graph [rankdir = "LR"]
        "device" [
            xref = "<f0> :py:class`HomieDevice` | <f1> attributes | <f2> nodes"
            shape = "record"
        ]
        "node" [
            label = "<f0> node | <f1> attributes | <f2> properties"
            shape = "record"
        ]
        "property" [
            label = "<f0> property | <f1> attributes"
            shape = "record"
        ]
        "device":f2 -> "node":f0
        "node":f2 -> "property":f0
    }
```
</details>

```
WARNING: Error running ['dot', '-Ncolor=#123457', '-Nstyle=solid,filled', '-Nfillcolor=#123458', '-Nfontcolor=#123456', '-Nfontsize=12', '-Nfontname=Roboto', '-Ecolor=#123459', '-Efontcolor=#123456', '-Efontsize=12', '-Efontname=Roboto', '-Gbgcolor=transparent', '-Gcolor=#123457', '-Gfontcolor=#123456', '-Gfontsize=12', '-Gfontname=Roboto', '-Tsvg'] with env {...} (dot.exe:17860): Pango-WARNING **: couldn't load font "Roboto Not-Rotated 12", falling back to "Sans Not-Rotated 12", expect ugly output.
Warning: node device, port f2 unrecognized
```

By adjusting the warning prompt in the theme's source, I was able to see the resulting altered dot syntax:
```dot
digraph g {
    graph [rankdir = "LR"]
    "device" [
        label=<&lt;f0&gt; :py:class`HomieDevice` | &lt;f1&gt; attributes | &lt;f2&gt; nodes>
        shape = "record"
    ]
    "node" [
        label = "<f0> node | <f1> attributes | <f2> properties"
        shape = "record"
    ]
    "property" [
        label = "<f0> property | <f1> attributes"
        shape = "record"
    ]
    "device":f2 -> "node":f0
    "node":f2 -> "property":f0
}
```

So, it looks like the line that escapes HTML syntax in the replaced label is causing dot to misinterpret the record-based node.

https://github.com/jbms/sphinx-immaterial/blob/21f2aeeb7fd182955a78893c30aaa6b84d6d5e8e/sphinx_immaterial/graphviz.py#L44

2bndy5:
resolves #164 

Implementation requires the manually specified `label` attribute be ~immediately~ after the `xref` attribute. I didn't add any unit tests because the docs demonstrate the purpose without fail, but we can add tests for certain graph types or use cases as we come across them.

### DIFF
--- a/docs/graphviz.rst
+++ b/docs/graphviz.rst
@@ -137,7 +137,7 @@ Overriding the cross-reference label
 The :graphvizattr:`label` generated with the new ``xref`` attribute uses escaped HTML characters.
 In some graphs, it is preferable to use unescaped HTML characters in the :graphvizattr:`label`.
 In this case, the generated :graphvizattr:`label` can be overridden by manually specifying the
-:graphvizattr:`label` immediately after the ``xref`` attribute.
+:graphvizattr:`label` after the ``xref`` attribute.
 
 .. rst-example:: A graph of "record" nodes
    
@@ -178,5 +178,5 @@ In this case, the generated :graphvizattr:`label` can be overridden by manually 
 .. admonition:: Implementation note
    :class: info
 
-   Using cross references in a single field of a record node (with multiple fields) is unsupported
+   Using a cross reference in a single field of a record node (with multiple fields) is unsupported
    because the manually specified :graphvizattr:`label` is used as is.
--- a/docs/graphviz.rst
+++ b/docs/graphviz.rst
@@ -123,10 +123,60 @@ example above.
 The reStructuredText will be parsed, and after resolving any cross references,
 will be substituted back into the graph definition as follows:
 
-- A :graphvizattr:`label` will be generated from the text content.
-- If there is at least one ``reference`` node, only the last such node is
+- A :graphvizattr:`label` will be generated from the cross reference's text content.
+- If there is at least one cross referable target, only the last such target is
   considered, and:
 
   - an :graphvizattr:`href` will be generated from its URL.
   - a :graphvizattr:`tooltip` will be generated from its title/tooltip, if any.
   - a :graphvizattr:`target` will be generated from its target, if any.
+
+Overriding the cross-reference label
+************************************
+
+The :graphvizattr:`label` generated with the new ``xref`` attribute uses escaped HTML characters.
+In some graphs, it is preferable to use unescaped HTML characters in the :graphvizattr:`label`.
+In this case, the generated :graphvizattr:`label` can be overridden by manually specifying the
+:graphvizattr:`label` immediately after the ``xref`` attribute.
+
+.. rst-example:: A graph of "record" nodes
+   
+   .. graphviz::
+
+      digraph {
+          graph [rankdir = "LR"]
+          "module" [
+              xref=":py:mod:`test_py_module.test`"
+              label="<f0> test_py_module.test | <f1> functions | <f2> classes"
+              shape = "record"
+          ]
+          "class" [
+              xref = ":py:class:`test_py_module.test.Foo`"
+              label = "<f0> Foo | <f1> attributes | <f2> methods"
+              shape = "record"
+          ]
+          "method" [
+              xref = ":py:meth:`test_py_module.test.Foo.capitalize()`"
+              label = "<f0> capitalize() | <f1> variables"
+              shape = "record"
+          ]
+          "function" [
+              xref = ":py:func:`test_py_module.test.func()`"
+              label = "<f0> func() | <f1> variables"
+              shape = "record"
+          ]
+          "class_attr" [
+              xref = ":py:attr:`~test_py_module.test.Foo.spam`"
+              shape = "record"
+          ]
+          "module":f2 -> "class":f0
+          "module":f1 -> "function":f0
+          "class":f2 -> "method":f0
+          "class":f1 -> "class_attr"
+      }      
+
+.. admonition:: Implementation note
+   :class: info
+
+   Using cross references in a single field of a record node (with multiple fields) is unsupported
+   because the manually specified :graphvizattr:`label` is used as is.
--- a/docs/graphviz.rst
+++ b/docs/graphviz.rst
@@ -120,11 +120,11 @@ This extension adds support for a special ``xref`` attribute, which may be set
 to a string value containing reStructuredText.  This is demonstrated in the
 example above.
 
-The reStructuredText will be parsed, and after resolving any cross references,
+The reStructuredText will be parsed as a document fragment, and after resolving any cross references,
 will be substituted back into the graph definition as follows:
 
-- A :graphvizattr:`label` will be generated from the cross reference's text content.
-- If there is at least one cross referable target, only the last such target is
+- A :graphvizattr:`label` will be generated from the parsed fragment's text content.
+- If the parsed fragment contains at least one hyperlink, only the last such hyperlink is
   considered, and:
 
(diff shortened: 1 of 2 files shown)


jbms:
It looks like graphviz allows an attribute such as `label` to be specified more than once in a given scope, with only the last value being used.  So I don't think we actually need any special code to handle this --- the user can just specify `label` afterwards and it will override the label set from xref.

2bndy5:
> It looks like graphviz allows an attribute such as label to be specified more than once in a given scope.

If so, then only the doc updates should be sufficient here. I'll do some testing and change this accordingly.

2bndy5:
> It looks like graphviz allows an attribute such as label to be specified more than once in a given scope.

This is correct, so I reverted the changes to the ext, and proofread the doc changes. I wish I knew this when I opened #164. At least it is now documented.

docs/graphviz.rst
```diff
@@ -123,10 +123,60 @@ example above.
 The reStructuredText will be parsed, and after resolving any cross references,
```
jbms, line 123: will be parsed as a document fragment

docs/graphviz.rst
```diff
@@ -123,10 +123,60 @@ example above.
...
 will be substituted back into the graph definition as follows:
 
-- A :graphvizattr:`label` will be generated from the text content.
-- If there is at least one ``reference`` node, only the last such node is
+- A :graphvizattr:`label` will be generated from the cross reference's text content.
```
jbms, line 126: parsed fragment's text content

docs/graphviz.rst
```diff
@@ -123,10 +123,60 @@ example above.
...
+- If there is at least one cross referable target, only the last such target is
```
jbms, line 127: If the parsed fragment contains at least one hyperlink, only the last such hyperlink is considered, and:

docs/graphviz.rst
```diff
@@ -123,10 +123,60 @@ example above.
...
   considered, and:
 
   - an :graphvizattr:`href` will be generated from its URL.
   - a :graphvizattr:`tooltip` will be generated from its title/tooltip, if any.
   - a :graphvizattr:`target` will be generated from its target, if any.
+
+Overriding the cross-reference label
+************************************
+
+The :graphvizattr:`label` generated with the new ``xref`` attribute uses escaped HTML characters.
```
jbms, line 137: generated from the new ``xref`` attribute uses the graphviz HTML-like attribute syntax which is not compatible with record-based nodes.  For record-based nodes, or in other cases where it is desired to override the label manually, simply specify the label after the xref attribute.

2bndy5: I might take a little liberty with this wording because
> uses the graphviz HTML-like attribute

doesn't imply that the text content within angle brackets has escaped HTML characters. Even though HTML-like labels very much support HTML characters.

I understand where you're going with this (there are a lot of recommended use cases for HTML-like labels), but the whole point of demonstrating overridden labels was because of the escaped HTML in the generated label.

Otherwise, a clever person might try the following without knowing about the escaped HTML:
```rst
.. graphviz::
    digraph {
        class [ shape=record ]
        record_node [
             xref=":ref:`\<f0\> field 1 \| \<f1> field 2 <target>`" 
             shape=record
        ]
        // notice the edge point to a specific field
        class -> record_node:f0
    }
```
Which results in a node that looks like 
image
but the edge can't locate the "f0 port" because the HTML was escaped escaped in the xref's label, resulting in a 
>Warning: node record_node, port f0 unrecognized

jbms: I see --- I guess you are saying that it is partially incorrect to say that the HTML-like attribute syntax is incompatible with record-based nodes --- because in your example the "|" is getting parsed correctly.

I suppose instead we can say:

generated from the new xref attribute cannot be used to specify record-based nodes, because angle brackets (``<`` and ``>``) in the fragment's text content will be escaped such that graphviz interprets them as literal text rather than specifying record fields.  For record-based nodes, or in other cases where it is desired to override the label manually, simply specify the label after the xref attribute.

2bndy5: Yes, exactly. Also, pointing directly at record nodes might be too narrow for the thesis of this section. I use record nodes in the example, so that can be implied as a use case that requires overriding the label. But HTML-like syntax is useful in a number of situations, not just record nodes. I'm having trouble picturing other cases where a user would need to have HTML-like label syntax and an xref attribute, but I'm sure someone will.

I'm currently working with this wording:
>The [label attribute](https://graphviz.org/docs/attrs/label/) generated with the new xref attribute uses graphviz’ HTML-like attribute syntax, but any HTML characters found within the xref parsed fragments’ text are escaped (eg. `< f0 >` becomes `&lt; f0 &gt;`). In some graphs, it is preferable to use unescaped HTML characters in the [label attribute](https://graphviz.org/docs/attrs/label/). In this case, the generated [label attribute](https://graphviz.org/docs/attrs/label/) can be overridden by manually specifying the [label attribute](https://graphviz.org/docs/attrs/label/) after the xref attribute.

jbms: Okay.

jbms:
Thanks.  Looks good to me.

Issue #46, opened 2022-03-19
production lists rendered incorrectly


2bndy5:
I noticed this a long time ago, but keep forgetting to file this issue.

An erroneous example can be seen [in our hosted docs](https://sphinx-immaterial.readthedocs.io/en/v0.7.3/additional_samples.html#production-lists)

I think the expected behavior is something that resembles a typical code-block but with cross-referencing links in certain parts of the code. I generally don't use the `productionlist` directive, but I figured I should say something.

The directive's docs are at [the Sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-productionlist)

An expected behavior can be observed from the attached simple project (uses furo theme)
[prod-list-proper.zip](https://github.com/jbms/sphinx-immaterial/files/8308599/prod-list-proper.zip)

![image](https://user-images.githubusercontent.com/14963867/159107100-9162ef2d-8623-436e-96f3-23e4c19e8bce.png)

2bndy5:
The problem seems to be related to misinterpreting the backticks `` ` `` within the directive's content.

jbms:
Yeah, I'm not sure what is going on.  I noticed this issue when first putting together this theme, but since I have no use for productionlist I didn't bother to look into it more.

2bndy5:
I have a feeling its related to CSS in upstream not compensated for production lists' style(s) or misunderstanding the `code` element's context.

2bndy5:
Peeked at the [`visit_productionlist()` in Sphinx](https://github.com/sphinx-doc/sphinx/blob/4bf9f65680e37038315f6b15ad4467a0a5033f75/sphinx/writers/html5.py#L490), and it doesn't enclose the literal block in a `<code>` element. Not sure if this violates HTML5 specs, but I think this should be elevated in sphinx before we address it here, which will likely involve another monkeypatch to (or rewrite of) the visitor.

With an added `<code>` element, the productionlist directive renders in a more expected way, but the backticks within the block are treated as references and triggering warnings in the build log since there's no definition to reference. According to the Sphinx docs (which is a bit vague in its terminology), the references are supposed to be defined by other production lists (see the [python docs](https://docs.python.org/3/reference/compound_stmts.html#the-try-statement) from which the the sample snippet in OP is taken from).

<details><summary>preliminary test with patched visitor</summary>
<br>

[...]

PR #178, opened 2022-10-29
patch visitor for production list directive

2bndy5:
Also un-comment the section in additional_samples.rst now that this patch addresses #46

Modified the production list example to use dot syntax grammar (and avoid unreferenced token warnings in the build log). I added use of the `:token:` role in production list example to better demonstrate the full utility of a production list.


jbms:
Thanks for fixing this.  Looks good to me.

PR #180, opened 2022-10-30
Fix search.highlight feature when loading from file:// URL

jbms:


jbms:
I don't believe there is any easy way to hide the search highlighting, though, after navigating to a search result.  Consequently, I didn't enable it by default for our docs.

2bndy5:
> I don't believe there is any easy way to hide the search highlighting, though, after navigating to a search result. Consequently, I didn't enable it by default for our docs.

This would actually matches my experience in other Sphinx themes. Typically I click the section permalink to disable search highlighting if it is too annoying.

I was excited to see the feature could be disabled at build time for this theme though. However, since the selected text follows the light/dark scheme CSS, it's not a bad idea to enable the feature for our docs. For other light-only schemed themes, I tend to use the "dark reader" browser extension, which looks rather ugly when highlighting search results.

jbms:
Not exactly related, but for the features that are solely implemented in JavaScript and therefore don't require a rebuild to enable, it could be neat to allow them to be interactively toggled in our docs in order to allow users to experiment with them.

2bndy5:
I think some features are excluded from the HTML in the templates if not specifically enabled. We could easily use the task-list directive (+ custom JS in our docs) to implement the toggling.

2bndy5:
Incidentally, I was contemplating something similar to the admonition icons because there are sample snippets that provide a complete suggestion for Octicons and FontAwesome.

2bndy5 (approved):
I can confirm this resolves it, thanks. It only seems to highlight exact matches though.

PR #181, opened 2022-10-30
Fix type_annotation_transforms for Sphinx 6

jbms:


setup.py
```diff
@@ -188,7 +188,7 @@ def run(self):
         ],
         "sphinx_immaterial.apidoc.cpp.cppreference_data": ["*.xml"],
     },
-    python_requires=">=3.7",
+    python_requires=">=3.8",
```
2bndy5, line 191: ❓ python 3.7 is still getting security fixes until 2023-06-27. Is there a reason to drop 3.7 support?

jbms: Sphinx 6 drops support for Python 3.7, and making the monkey patch work for 3.7 when using Sphinx < 6 would require even more branching.  But I see that I need to revise this anyway, so perhaps I can fix it.

2bndy5: That's good enough reason for me (it just wasn't obvious).

sphinx_immaterial/graphviz.py
```diff
@@ -15,7 +15,7 @@
 import sphinx.ext.graphviz
 import sphinx.util.docutils
 import sphinx.util.logging
-from sphinx.writers.html import HTMLTranslator
+from sphinx.writers.html5 import HTML5Translator
```
2bndy5, line 18: Seems to me that this change is the only remaining violation in mypy checks. If I revert the change, then the mypy step passes.

jbms:
I think I've fixed all of the typing issues now.  Unfortunately the setup-graphviz seems to be failing now on macOS, not sure how we can work around that.

2bndy5:
There seems to be a conflict between the installed python and the python homebrew tries to install. Although I can't clearly tell what the problem is. My instinct is to setup-graphviz before setup-python in the CI.

2bndy5:
We might be able to get [graphviz through conda-forge](https://anaconda.org/conda-forge/graphviz) instead of OS-specific pkg managers (like the current action does).

2bndy5:
Damn. I still have no idea what the problem is based on the homebrew output.

jbms:
I think this is a general issue with using homebrew on github actions.  Reordering the graphviz step before the Python setup step did not fix the issue, but skipping `brew update` seems to avoid the issue.

https://github.com/ts-graphviz/setup-graphviz/issues/457

2bndy5 (approved):

jbms:
Perhaps we should tag a new version with Sphinx 6 support?

2bndy5:
sure, but I was thinking a patch bump.

PR #182, opened 2022-11-01
add inline icon role and doc for it

2bndy5:
Add ability to use a builtin icon or a project specific SVG (located in static folder) as an inline text element.


sphinx_immaterial/inline_icons.py
```diff
@@ -0,0 +1,57 @@
+"""This module adds the md-icon role. This is specific to this this theme because,
```
jbms, line 1: Perhaps we should use `im-icon` or `im:icon` rather than `md` (for mkdocs)?

sphinx_immaterial/apidoc/apidoc_formatting.py
```diff
@@ -96,7 +96,8 @@ def depart_caption(
         "literal_block"
     ):
         self.body.append("</div>\n")
-    else:
+    elif not isinstance(node.parent, docutils.nodes.figure):
+        # calling super_func() on a figure results in 2 permalinks in caption
```
jbms, line 100: I assume this is an independent fix?

docs/customization.rst
```diff
@@ -244,15 +189,13 @@ Configuration Options
...
-            - |fa-git| ``fontawesome/brands/git``
-            - |fa-git-alt| ``fontawesome/brands/git-alt``
-            - |fa-git-square| ``fontawesome/brands/git-square``
-            - |fa-github| ``fontawesome/brands/github``
-            - |fa-github-alt| ``fontawesome/brands/github-alt``
-            - |fa-github-square| ``fontawesome/brands/github-square``
-            - |fa-gitlab| ``fontawesome/brands/gitlab``
-            - |fa-gitkraken| ``fontawesome/brands/gitkraken``
-            - |fa-bitbucket| ``fontawesome/brands/bitbucket``
+            - :md-icon:`fontawesome/brands/git` ``fontawesome/brands/git``
```
jbms, line 192: Could remove the redundancy using the jinja directive

docs/admonitions.rst
```diff
@@ -229,10 +231,27 @@ Any of the above builtin admonitions' icons can be changed using the
...
+              "admonition": {
+                  "seealso": "octicons/eye-24", # (1)!
+              },
+          },
+      }
+
+   .. code-annotations::
+      #. Uses the icon :md-icon:`octicons/eye-24`
+
+   Then, use the the generated CSS variable :css:`--md-admonition-icon--seealso` in the custom CSS for
```
jbms, line 253: I'm not sure I understand the intent of this section ---- what is the reason for giving special attention to the `seealso` admonition --- and why is there a mention of CSS variables, etc.?

docs/inline_icons.rst
```diff
@@ -0,0 +1,85 @@
...
+The following sets of icons are bundled with this theme:
+
+.. rst-example:: Using a builtin icon
+
+    - :md-icon:`material/material-design` `Material Design <https://materialdesignicons.com/>`_
+    - :md-icon:`fontawesome/regular/font-awesome` `Font Awesome <https://fontawesome.com/search?m=free>`_
+    - :md-icon:`octicons/mark-github-16` `Octicons <https://octicons.github.com/>`_
+    - :md-icon:`simple/simpleicons` `Simple Icons <https://simpleicons.org/>`_
+
+You can also browse these icons locally from the sphinx-immaterial theme's install path under
```
jbms, line 37: I suppose we could integrate the "icon" index from mkdocs-material into our docs...

sphinx_immaterial/inline_icons.py
2bndy5: I was thinking of maybe `si` for sphinx-immaterial. I only went with `md` because the other custom directives start with that (which I had assumed meant material design).

jbms: si sounds good to me

sphinx_immaterial/apidoc/apidoc_formatting.py
2bndy5: Yeah, I noticed this bug when using a figure to caption an image in the new doc. I can submit this an a separate PR though, if desired.

jbms: It is okay to combine it, just wanted to confirm.

docs/admonitions.rst
2bndy5: I could make it more generic in saying that

> Some admonitions inherit their styles from the note admonition. This specifically happens for the seealso admonition, which has no specific rules in CSS.

The reason for mentioning the CSS variable is a broad stroke of the implementation in the HTML template. There's no need to copy the SVG data to a custom CSS variable if the desired SVG ships with the theme.

I didn't alter the example CSS about custom admonitions because it may still be useful to demonstrate using SVG data that doesn't ship with the theme.

docs/inline_icons.rst
2bndy5: Yeah, I thought about that too. This is something that can be addressed in the next merge from upstream, assuming we currently exclude the part of the build script that creates the json.

docs/admonitions.rst
jbms: Oh I see --- the issue is that for `seealso` it is not sufficient to just specify an icon under `html_theme_options` because those only work for the admonitions specified in `src/assets/stylesheets/main/extensions/markdown/_admonition.scss`

Ideally we could fix that and add a nice Python interface for defining custom admonitions, but that is independent of this issue.

sphinx_immaterial/inline_icons.py
2bndy5: I'll also change the `.md-icon-inline` CSS class to `.si-icon-inline`. The `.md-icon` CSS class is something we get from upstream.

sphinx_immaterial/inline_icons.py
```diff
@@ -0,0 +1,57 @@
...
+            if svg.exists():
+                break
+        else:
+            raise FileNotFoundError(
+                f"{icon_name} not found in html_static_path and not bundled with theme"
+            )
+    svg_data = svg.read_text(encoding="utf-8")
+
+    css_classes = " ".join(node["classes"])
+    self.body.append(f'<span class="{css_classes}">{svg_data}</span>')
```
2bndy5, line 36: I was thinking about extracting this alogorithm into a separate importable function, so it could be used in other parts of the theme (eg for nav icons or #183 )

jbms: That would make sense.  One issue I noticed is that this will duplicate the svg for each use of the icon.  If the same icon is referenced many times (e.g. as may be the case of nav icons) that will bloat the generated output.   Instead we could emit it as CSS into the css bundle, once we have that implemented, and then it would only occur once.  Or maybe could switch between the two behaviors, depending on whether the icon is used more than once.

docs/customization.rst
```diff
@@ -244,15 +189,11 @@ Configuration Options
...
-            - |fa-git-square| ``fontawesome/brands/git-square``
-            - |fa-github| ``fontawesome/brands/github``
-            - |fa-github-alt| ``fontawesome/brands/github-alt``
-            - |fa-github-square| ``fontawesome/brands/github-square``
-            - |fa-gitlab| ``fontawesome/brands/gitlab``
-            - |fa-gitkraken| ``fontawesome/brands/gitkraken``
-            - |fa-bitbucket| ``fontawesome/brands/bitbucket``
+            .. jinja:: repo_icons
+
+                {% for icon in repo_icons %}
```
jbms, line 194: Can just use:

```
{% set repo_icons = ... %}
```

that way the icons can be specified inline rather than in `conf.py`.

jbms: Or it might work to just specify the list inline in the `{% for`

sphinx_immaterial/inline_icons.py
```diff
@@ -0,0 +1,68 @@
...
+from sphinx.application import Sphinx
+from sphinx.builders import Builder
+from sphinx.writers.html5 import HTML5Translator
+
+
+def read_svg_into_span(builder: Builder, icon_name: str, classes: List[str]) -> str:
+    """Reads the SVG data into a HTML span element."""
+    svg = Path(__file__).parent / ".icons" / icon_name
+    if not svg.exists():
+        static_paths: str = getattr(builder.config, "html_static_path")
```
jbms, line 16: I think this type annotation is wrong --- should be `List[str]`

Also could skip the extra `if not static_paths` check

sphinx_immaterial/inline_icons.py
```diff
@@ -0,0 +1,68 @@
...
+        else:
+            raise FileNotFoundError(
+                f"{icon_name} not found in html_static_path and not bundled with theme"
+            )
+    svg_data = svg.read_text(encoding="utf-8")
+    css_classes = " ".join(classes)
+    return f'<span class="{css_classes}">{svg_data}</span>'
+
+
+class md_icon(nodes.container):
```
jbms, line 32: rename to `si_icon` and also for the function below

sphinx_immaterial/inline_icons.py
2bndy5: type hint is wrong, good catch.

sphinx_immaterial/inline_icons.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+    )
+    raise nodes.SkipNode()
+
+
+def icons_role(
+    role: str, rawtext: str, text: str, lineno: int, inliner, options={}, content=[]
+) -> Tuple[List[nodes.Node], List[nodes.system_message]]:
+    path, classes = (text, "")
+    if ";" in text:
```
jbms, line 53: Are there other properties other than `class` that we may wish to support in the future?  If so it could make sense to use a syntax like `class=`.  On the other hand I guess we could just use the `name=value` syntax for those other options and keep the shorthand for `class`.

2bndy5: This implementation is mostly based on sphinx-design inline icons. They have some other options about sizing (by pixels), but it is flawed because not all SVG data behave well with specific sizes and it requires a certain order of properties (`path;classes;sizes;color` or something like that).

I personally think that any other properties can be managed in CSS more efficiently, but the embedded data in a span is about to change. We could re-assess when this feature is more finalized with #183. I will add a slice notation to make sure that only 2 list items are expanded.

jbms: One property that might be useful and cannot be handled from CSS is `title`, or perhaps `alt` if it is supported.

I imagine these icons would most commonly be used via a `|substitution|` and the title text could indicate the meaning of the icon, if the icon is intended to provide additional information rather than being purely decorative.

Anyway this is fine as is.

jbms:
Looks good to me.

Issue #183, opened 2022-11-01
Python API for custom admonitions


jbms:
Ideally, we would offer a Python API for defining custom admonitions, likely in the form of a config option to be specified in `conf.py`.

Currently it is possible to override the icons for existing admonitions using via the `icon/admonition` theme option, but to define a new admonition type, the user needs to:
1. Create a custom CSS file and add some boilerplate to it: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#custom-admonitions
2. Add an entry to the `icon/admonition` theme option.
3. Define a custom sphinx admonition directive for it

It would be nicer if we provided a single option where the user could specify:
- Directive name/names
- CSS class name (maybe same as directive name?)
- Default title
- Colors
- Icon

Ideally, this would emit the necessary additional CSS to a new combined bundle that contains the static pre-generated CSS bundle with any dynamically-generated CSS appended.  That way there is still a single bundle, and we avoid duplicating the additional CSS on every HTML page, as occurs with the current `icon/admonition` theme option.

2bndy5:
This reads like the details/summary tags (collapsible admonitions) are excluded.

If the CSS is written with a template that emits no whitespace, I suppose it wouldn't need to use software like the Sass compiler.

As far as HTML templates, we'd have to add our generated CSS into base.html inside a Jinja if statement. I'd prefer it if the current (`icon[admonition]`) upstream solution remained in the templates though. We can just discourage its use after this issue is resolved.

jbms:
Supporting the `details` directives as well would make sense --- I just didn't think about that.

Yes, seems reasonable to leave the existing `icons[admonitions]` code in place.

We don't necessarily need to add the generated CSS inline into the HTML at all --- instead we can just replace the normal CSS bundle with a modified one with the extra generated content appended.

Sass provides useful functionality but I think it would significantly complicate the sphinx build to have to invoke sass as part of it; instead we should just emit plain CSS directly (without extra whitespace, as you noted).  The mkdocs-material theme provides suggested CSS to add for defining custom admonitions:

https://squidfunk.github.io/mkdocs-material/reference/admonitions/#custom-admonitions

Perhaps that is basically exactly what is needed?

2bndy5:
[...]

Issue #132, opened 2022-07-15
Adding links to the index (genindex) and module index to the navigation sidebar


brechtm:
I'm currently using the RTD Sphinx theme, for which I've added templates to add links to these indices the the TOC as described in [this StackOverflow answer](https://stackoverflow.com/a/71686673/438249). See the _Indices and tables_ section in the sidebar at http://www.mos6581.org/rinohtype/0.5.4/ for the result.

I've been trying to do the same in sphinx-immaterial, but the templates do not seem to define blocks where these can easily be injected. My knowledge of Jinja is limited, but I think it is necessary to replace all of the `site_nav` block in _layout.html_, duplicating all of the block's contents and injecting the links. Here is my attempt so far:

<details>
  <summary><code>_templates/layout.html</code></summary>

```jinja
{% extends "!layout.html" %}

{% block site_nav %}

  <!-- Main navigation -->
  {% if nav %}
    {% if page and page.meta and page.meta.hide %}
      {% set hidden = "hidden" if "navigation" in page.meta.hide %}
    {% endif %}
    <div
      class="md-sidebar md-sidebar--primary"
      data-md-component="sidebar"
      data-md-type="navigation"
      {{ hidden }}
    >
      <div class="md-sidebar__scrollwrap">
        <div class="md-sidebar__inner">
          {% include "partials/nav.html" %}

          <ul class="md-nav__list">
            <li class="md-nav__item md-nav__item--nested">
              <nav class="md-nav" aria-label="User Manual" data-md-level="1">
                <span class="md-nav__icon md-icon"></span>
                <span class="md-ellipsis">Indices and tables</span>
                <ul class="md-nav__list">
                  <li class="md-nav__item">
                    <a href="{{pathto('genindex.html', 1)}}" class="md-nav__link">
                      <span title="intro (document)" class="md-ellipsis">Index</span>
                    </a>
                  </li>
                  <li class="md-nav__item">
                    <a href="{{pathto('py-modindex.html', 1)}}" class="md-nav__link">
                      <span title="install (document)" class="md-ellipsis">Module Index</span>
                    </a>
                  </li>
                </ul>
              </nav>
          </ul>

        </div>
      </div>
    </div>
  {% endif %}

  <!-- Table of contents -->
  {% if not "toc.integrate" in features %}
    {% if page and page.meta and page.meta.hide %}
      {% set hidden = "hidden" if "toc" in page.meta.hide %}
    {% endif %}
    <div
[...]

PR #186, opened 2022-11-03
customized admonition

2bndy5:
solves #183 and conforms #182 with improvements used here.


mhostetter:
Minor suggestion: perhaps instead of having the user instantiate a `CustomAdmonition` object in the `sphinx_immaterial_custom_admonitions` list in `conf.py`, maybe have the user specify a dictionary of kwargs. Those kwargs can be used internally to instantiate the `CustomAdmonition` objects. Might be a little more user-friendly.

### From

```python
from sphinx_immaterial.custom_admonitions import CustomAdmonition

sphinx_immaterial_custom_admonitions = [
    CustomAdmonition(
        name="example-admonition",
        color=(198, 177, 6),
        icon="material/school",
    ),
]
```

#### To

```python
sphinx_immaterial_custom_admonitions = [
    {
        "name": "example-admonition",
        "color": (198, 177, 6),
        "icon": "material/school",
    },
]
```

2bndy5:
I had considered it, but using dicts are problematic because undetected typos in the keys may result in its value not being used (& unused keys not noted during build time). We had worked around this in the past with the `object_description_options` config, but that is vastly more general of a configuration.

The other reason I didn't use a dict is because a class constructor can fill in unspecified default values for free (kwargs in c'tor). Plus, I can do type checking on the values (like making sure the color is `rgb()` friendly in CSS) before going any further with the docs build.

I'm not shooting it down, it should be possible, but I just found it quicker to get up and running with the class approach. Under the hood, it is still quite flexible to make a switch. I'll try it out in the coming days (still refining/testing the admonitions that override existing directives in sphinx).

2bndy5:
I have been rather annoyed by a problem in our docs which always shows [the `__new__()` function in the docs for the `CustomAdmonition` class](https://sphinx-immaterial--186.org.readthedocs.build/en/186/admonitions.html#sphinx_immaterial.custom_admonitions.CustomAdmonition.__new__) (despite try `no-special-members` and `no-undoc-members`). I think its a side effect of using autodoc and the python apigen in the same project.

jbms:
I also like Matt's suggestion.  If you use pydantic to define a data class and to parse the option in config_inited, that will take care of type checking and defaults.

2bndy5:
Will do. I'm just finishing up the todo directive override now.

2bndy5:
So switching to pydantic was easy, but now I'm trying to document the config values' datatypes... Currently trying to follow what was done for the obj_desc_opts, but it is complex. So far, I've created the new `admoniconf` directive/role and moved the params' description into the admonition.rst doc.

jbms:
> So switching to pydantic was easy, but now I'm trying to document the config values' datatypes... Currently trying to follow what was done for the obj_desc_opts, but it is complex. So far, I've created the new `admoniconf` directive/role and moved the params' description into the admonition.rst doc.

Does autodoc not work to document the dataclass?

2bndy5:
Oh, I hadn't thought to do it that way. I was trying to move away from autodoc because the rogue `__new__()` that I can't seem to get rid of.

jbms:
> Oh, I hadn't thought to do it that way. I was trying to move away from autodoc because the rogue `__new__()` that I can't seem to get rid of.

How about adding it to `:exclude-members:`, as done here:

https://github.com/jbms/sphinx-immaterial/blob/main/docs/apidoc/cpp/apigen.rst

2bndy5:
> How about adding it to `:exclude-members:`

I didn't see that in the autodocs' docs. It worked!

It doesn't show the default value for the class attrs, but I can use the `__init__` params instead.
image

2bndy5:
Opening this up for review.

Some changes to the generic admonitions should bring compliance to the deprecated aliased admonition CSS classes in upstream. While some of the classes getting removed in upstream v9.0 will have to be added back as Sphinx already has specific admonition directives that upstream CSS won't style. Namely, these classes are
- `error`
- `caution`
- `attention`
- `hint`
- `important`
- `todo`

On the plus side, this will give us a chance to style them more uniquely (if desired). The main reason these CSS classes are deprecated is to cut back on the CSS bundles footprint, but mkdocs isn't constrained by predefined admonitions like we are.

2bndy5:
I think we need to reduce the workflow CI triggers. It is excessive to have the workflow run 2x for every commit in an open PR. I'm running into bandwidth/connection problems and it takes a pretty long time to complete 1x workflow run.

I've gotten more comfortable using [pre-commit](https://pre-commit.com/), which we can use to reduce linting errors reported in CI, cache dependencies on the github runner, trim trailing whitespace, detect yaml/json syntax errors, and ensure uniform line endings.

FYI, the [coverage pkg's `run` command](https://coverage.readthedocs.io/en/6.5.0/#quick-start) can invoke pytest without installing the theme src (& build a coverage report if that sounds interesting), and conf.py can be configured to add the theme src to `sys.path` (I've been doing this locally) -- resulting in less need to install the theme in CI.

mhostetter:
> I think we need to reduce the workflow CI triggers. It is excessive to have the workflow run 2x for every commit in an open PR.

I think if you change

https://github.com/jbms/sphinx-immaterial/blob/09b41f950db5b3ba62143aca14276df6c6f3fcc8/.github/workflows/build.yml#L3

to

```yaml
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
  workflow_dispatch:
```

that should do the trick.

2bndy5:
not a fan of limiting by branch because we dev on any branch but main. I was thinking more like limiting the PR sync events:
```yml
on:
  push:
  pull_request:
    types: [opened, closed, repoened]
  workflow_dispatch:
```

mhostetter:
> not a fan of limiting by branch because we dev on any branch but main.

To be clear, my suggestion would run the CI on push to *any branch* that is tied to a PR to be merged onto `main` and anytime someone pushes to `main` (either manually or post-PR merge).

2bndy5:
I hadn't realized the PR target was the determining factor for trigger filters. That sounds like its worth a try.

docs/admonitions.rst
```diff
@@ -2,219 +2,309 @@
...
 Admonitions from mkdocs-material
 ********************************
 
-Some additional admonitions are supported via the source code from the mkdocs-material theme.
-These admonitions can still be used, but the syntax is a little different because it relies
-on the generic :dutree:`admonition` defined in the reStructuredText specifications.
-
-To use the following admonitions' styles from the mkdocs-material theme, the rST syntax is
-shown to demonstrate using the :rst:`:class:` option of generic admonitions.
+Some additional admonitions are supported via the CSS styles inherited from the mkdocs-material
```
jbms, line 49: Perhaps we should make these optional, in case they conflict with another extension or user-defined directive?

docs/admonitions.rst
```diff
@@ -2,219 +2,309 @@
...
+               It can even span multiple lines.
 
-Removing the title
-******************
+            Notice the blank line between the directive's beginning block and this content block.
+
+            This admonition's directive was created just for this documentation using the
+            sphinx-immaterial theme's `Custom Admonitions`_ feature.
+
+   .. rst:directive:option:: no-title
```
jbms, line 150: Perhaps we don't need this option --- instead just specifying an empty title with `:title:` could serve the same purpose.

Also, if the title is specified both as the argument and as this option, perhaps it should just result in an error rather than concatenating the two; that would be more consistent behavior, I think, if we decide to have an empty `:title:` mean no title.

docs/admonitions.rst
```diff
@@ -2,219 +2,309 @@
...
+   .. rst-example::
 
-         html_static_path = ["_static"]
-         html_css_files = ["extra_css.css"]
+      .. example-admonition::
+
+         This is simple a example.
+
+
+.. example:: Legacy approach inherited from the mkdocs-material theme.
```
jbms, line 258: Perhaps we don't need this example anymore?

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,320 @@
...
+    icon: str
+    title: str
+    override: bool = False
+
+    def __init__(
+        self,
+        name: str,
+        color: Tuple[int, int, int],
+        icon: str,
+        title: str = None,
```
jbms, line 41: `title: Optional[str] = None`

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,320 @@
...
+    env = jinja2.Environment(
+        loader=jinja2.FileSystemLoader(str(PurePath(__file__).parent))
+    )
+    template = env.get_template("custom_admonitions.css")
+    generated = template.render(
+        icons=custom_icons,
+        admonitions=custom_admonitions,
+    )
+    static_output = Path(app.outdir) / "_static" / "stylesheets"
+    for css in static_output.glob("main.*.min.css"):
```
jbms, line 235: Since we are modifying the CSS here, no purpose is served by including a hash in the name.  Instead, we could change `tools/build/` to leave the CSS under the original names `main.css` and `palette.css`, and not put them in `_static` to be copied automatically.  Then here we could produce a single combined CSS file (containing `main.css`, `palette.css` if `config.theme.palette` is set, and the custom admonitions), compute its hash, and copy it to the output directory.

sphinx_immaterial/inline_icons.py
```diff
@@ -9,22 +9,27 @@
...
-    return f'<span class="{css_classes}">{svg_data}</span>'
+def load_svg_into_builder_env(builder: Builder, icon_name: str) -> str:
+    """Reads the SVG data into a builder env variable for CSS generation."""
+    css_icon_name = icon_name.replace("/", "_").replace("\\", "_")
+    custom_icons = getattr(builder.env, "sphinx_immaterial_custom_icons", {})
+    icon_name += ".svg"
+    if css_icon_name not in custom_icons:
+        svg = Path(__file__).parent / ".icons" / icon_name
+        if not svg.exists():
+            static_paths: List[str] = getattr(builder.config, "html_static_path")
```
jbms, line 20: Since these icons don't actually need to be copied to the output directory, I think it would be better to create a new config option for specifying icon paths, rather than re-using `html_static_path`.  Additionally, It might be better to search these user-specified icon paths first, to allow the user to override builtin icons.

jbms:
> Opening this up for review.
> 
> Some changes to the generic admonitions should bring compliance to the deprecated aliased admonition CSS classes in upstream. Some of the classes getting removed in upstream v9.0 will have to be added back as Sphinx already has specific admonition directives that upstream CSS won't style. Namely, these classes are
> 
> * `error`
> * `caution`
> * `attention`
> * `hint`
> * `important`
> * `seealso`
> * `todo`
> 
> On the plus side, this will give us a chance to style them more uniquely (if desired). The main reason these CSS classes are deprecated is to cut back on the CSS bundle's footprint, but mkdocs isn't constrained by predefined admonitions like we are.

We could support aliases by remapping them in Python to a different CSS class, so that there is no need for additional CSS.

We could also keep track of which ones are used and only add the CSS as needed.

docs/admonitions.rst
2bndy5: Good idea. I was worried about this also because the terse names used (`bug`). I also considered using the `si` prefix for only the directive name.

sphinx_immaterial/custom_admonitions.py
2bndy5: Well the map files for the CSS would be useless. These are really only helpful in theme development because the absolute paths in the map files are specific to the machine that ran `npm run build`.

I do like this idea though, I just didn't want to mess with the build script and the HTML template. I'd actually be happier if the CSS file was named "sphinx_immaterial_theme" instead of just "main", so users are still free to use the common name "main" for their extra CSS file (s).

sphinx_immaterial/custom_admonitions.py
2bndy5: It's not actually optional. It needs to have a value after `super().__init__()` is called.

docs/admonitions.rst
2bndy5: I left it in a collapsed admonition because it will still work. Same reason I left the section below this about changing the default icon (but with a warning about HTML bloat).

I didn't mention anything about the CSS classes `inline` and `end` inherited from upstream. They seem buggy when I use them, because the HTML structure (sphinx vs mkdocs) seems different enough to cause a bunch of alignment problems.

docs/admonitions.rst
2bndy5: I actually had it to do that when I first wrote it, but an empty value to a `title` option doesn't feel user intuitive enough. It didn't seem too radical to have a `no-title` flag because this is similar to what autodoc does for some it's directive's options.

The way it is now felt more natural to write in RST.

sphinx_immaterial/custom_admonitions.py
2bndy5: I will remove the `None` default ass that is misleading and would be a invalid value.

sphinx_immaterial/inline_icons.py
2bndy5: I have this implemented, but our docs are using
```py
sphinx_immaterial_icon_path = html_static_path
```
because the custom icon (sphinx_logo.svg) is displayed using both `si-icon` role and the `image` directive for demonstration purposes.

docs/admonitions.rst
2bndy5: I have 
- renamed the inherited admonitions' directives with a `si-` prefix.
- added a config value that allows disabling generating these convenience directives (defaults to True)

2bndy5:
>We could support aliases by remapping them in Python to a different CSS class, so that there is no need for additional CSS.
>
>We could also keep track of which ones are used and only add the CSS as needed.

Using the aliases in python is certainly easier. We wouldn't need to add CSS or keep track of the admonitions used if we just change the default class used for the aliased admonitions. It would be similar to how I managed using a different name in the inherited admonition directives and still use the upstream CSS classes (currently not using the deprecated class aliases specific to mkdocs-material though).

sphinx_immaterial/custom_admonitions.py
2bndy5: This was always going to the grey area for me. I think the `build-finished` event is triggered too late for changing what CSS files are copied. I think the [`html-collect-pages` event](https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-html-collect-pages) may be more suitable to customizing the CSS files that will be copied to the output.

Additionally, _I think_ our custom builder's `init_css_files()` may need to be modified to temporarily exclude the theme's main & palette CSS bundles. Then they could be combined into a file that gets copied into the `outdir` path.

Where I get hazy is about the html templates because they currently look for main and palette bundles when generating each page. I didn't want to alter them, so I just let them behave normally and copied the generated CSS into the main.css that was copied.

Do you have a specific idea about which events to use for this request?

jbms: > Well the map files for the CSS would be useless. These are really only helpful in theme development because the absolute paths in the map files are specific to the machine that ran `npm run build`.

I think at least in some cases the map file also can contain the full original source, so that it doesn't matter if the paths are valid locally.

> 
> I do like this idea though, I just didn't want to mess with the build script and the HTML template. I'd actually be happier if the CSS file was named "sphinx_immaterial_theme" instead of just "main", so users are still free to use the common name "main" for their extra CSS file(s).

Agreed, I think that makes sense.  In fact we could concatenate the user's extra css files as well, though.

The change to the html template would be pretty minimal --- we could just delete the existing tags that load the css and rely on the extra css loading variables.

2bndy5: > we could concatenate the user's extra css files

If we do and the user's CSS has a syntactical error (like a missing `}` or `;`), then wouldn't it make that harder to diagnose when all the CSS cannot be loaded by the browser? I'm not sure how browsers handle erroneous CSS files (if the scope of the error is considered).

2bndy5: We can only partially minify the user CSS (remove LFs).

jbms: > This was always going to the grey area for me. I think the `build-finished` event is triggered too late for changing what CSS files are copied. I think the [`html-collect-pages` event](https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-html-collect-pages) may be more suitable to customizing the CSS files that will be copied to the output.
> 
> Additionally, _I think_ our custom builder's `init_css_files()` may need to be modified to temporarily exclude the theme's main & palette CSS bundles. Then they could be combined into a file that gets copied into the `outdir` path.
> 
> Where I get hazy is about the html templates because they currently look for main and palette bundles when generating each page. I didn't want to alter them, so I just let them behave normally and copied the generated CSS into the main.css that was copied.

I think it would be a simple enough change that it likely won't cause significant problems later.
> 
> Do you have a specific idea about which events to use for this request? 

I think env-check-consistency

2bndy5: > We can only partially minify the user CSS (remove LFs with any surrounding whitespace).

We could try using [csscompressor](https://github.com/sprymix/csscompressor).

2bndy5: This is done, but I didn't combine the user defined CSS files because sphinx doesn't provide a convenient method like `remove_css_file()` (as opposed to `add_css_file()`). When I peeked under the hood for `add_css_file()`, it seems to add the file to the `app.registry` and the builder (if any).

My attempts at using a 3rd party solution to minify CSS failed for SVG data (embedded in `url()` definitions). So, I figured it was just easier to just let the user minify their docs' extra CSS and let Sphinx do its thing.

2bndy5: Apparently the env-check-consistency event fires too early to generate the CSS. I found that the builder env var used to store aggregated icons' data is missing all inline icons (added to env by the `visit_si_icon()`).

I was able to produce desired behavior by only generating the CSS on build-finished event and calling `add_css_file()` on config-inited event. As of now, this didn't produce any warnings about copying non-existing assets.

2bndy5:
>>Some of the classes getting removed in upstream v9.0 will have to be added back as Sphinx already has specific admonition directives that upstream CSS won't style
>
> We could support aliases by remapping them in Python to a different CSS class, so that there is no need for additional CSS.

We are now prepared for this, but I left the code commented out until it is needed.

https://github.com/jbms/sphinx-immaterial/blob/70a25e57156f1137fd9361ba4eace43223c500e6/sphinx_immaterial/custom_admonitions.py#L211-L219

docs/admonitions.rst
jbms: Okay, makes sense.

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,354 @@
...
+from sphinx.writers.html5 import HTML5Translator
+from .inline_icons import load_svg_into_builder_env
+
+logger = getLogger(__name__)
+
+# treat the todo directive from the sphinx extension as a built-in directive
+admonitionlabels["todo"] = _("Todo")
+
+
+class CustomAdmonitionConfig(pydantic.BaseModel):
```
jbms, line 27: I am not too familiar with pydantic, but when I have used it, I have used `pydantic.dataclasses.dataclass` rather than `BaseModel` --- but perhaps you have a reason for preferring it this way.  Also, is it necessary to define `__init__` rather than relying on the built-in definition?  Is that needed to show the default values in autodoc?

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,354 @@
...
+    app.add_css_file(css_name)
+
+
+def setup(app: Sphinx):
+    """register our custom directive."""
+    app.add_config_value(
+        name="sphinx_immaterial_custom_admonitions",
+        default=[],
+        rebuild="env",
+        types=[List[Dict[str, Union[str, Tuple[int, int, int], bool]]]],
```
jbms, line 335: Sphinx doesn't actually do any deep validation of the types --- so you can just specify `List[CustomAdmonition]` as the type and then our `confval` directive will show more useful type information.

sphinx_immaterial/custom_admonitions.py
2bndy5: I was just following the pydantic docs about using the BaseModel class; I have no preference either way.

`__init__()` is needed to ensure the color components range is [0, 255], adopt the `name` as a `title` if not provided by the user, and yes the default values only show when documenting the params. If I just document the class's attributes, then autodoc doesn't show the default values - this may lead to confusion about what fields are required in the config's dict.

sphinx_immaterial/custom_admonitions.py
2bndy5: Every time I tried this, I get an error about CustomAdmonition class is not generic. It seems that the param needs to be a literal list of types. So, `types=[List[CustomAdmonitionConfig]],` will work, but it may be confusing because the config is actually a list of dicts, not a list of instantiated objects. 
image

sphinx_immaterial/custom_admonitions.py
2bndy5: I'm also using `__init__()` to limit the allowed `name` value to consist of only characters in the pattern `[a-zA-Z0-9\-_]`. Then the title can be extracted from the name with `" ".join(re.split(r"\W+", name)).title()`

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,354 @@
...
+        if isinstance(admonition_node, sphinx.ext.todo.todo_node):
+            # todo admonitions need extra info for the todolist directive
+            admonition_node["docname"] = admonition_node.source
+            self.state.document.note_explicit_target(admonition_node)
+        else:
+            self.add_name(admonition_node)
+        if "collapsible" in self.options:
+            admonition_node["collapsible"] = self.options["collapsible"]
+        if "no-title" in self.options and "collapsible" in self.options:
+            logger.warning(
```
jbms, line 188: Probably should be an error.

docs/admonitions.rst
```diff
@@ -2,219 +2,318 @@
...
 ********************************
 
-Some additional admonitions are supported via the source code from the mkdocs-material theme.
-These admonitions can still be used, but the syntax is a little different because it relies
-on the generic :dutree:`admonition` defined in the reStructuredText specifications.
+Some additional admonitions are supported via the CSS styles inherited from the mkdocs-material
+theme. These admonition styles can be used via the :rst:dir:`admonition` directive's :rst:`:class:`
+option. 
+
+.. confval:: sphinx_immaterial_generate_inherited_admonitions
```
jbms, line 53: I think originally you just defined an `xxx` directive for each `xxx` admonition class.  I was concerned about polluting the directive namespace.  However, I expect in practice users will want to use these directives with nice shorthand names, rather than `si-xxx`.  Therefore, it might be better for this option to just control whether we provide the `xxx` directives --- potentially we could always provide the `si-xxx` directives since they shouldn't cause any conflicts.

However, it might also be convenient to provide a way for a user to easily define their own admonition directives that map to the built-in admonition styles.  Perhaps instead of this option, we could provide a way for the user to specify these admonitions via `custom_admonitions`, e.g. with an option to have the admonition directive just map to an existing class rather than define a new class.

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,354 @@
...
+    jinja_env = jinja2.Environment(
+        loader=jinja2.FileSystemLoader(str(PurePath(__file__).parent))
+    )
+    template = jinja_env.get_template("custom_admonitions.css")
+    generated = template.render(
+        icons=custom_icons,
+        admonitions=custom_admonitions,
+    )
+
+    css_name = "stylesheets/sphinx_immaterial_theme.min.css"
```
jbms, line 311: We could include a hash of the contents in the filename as in mkdocs-material --- the purpose of this is to allow users to mark these bundle files as cachable with a long TTL.  I'm not sure how useful that is in practice, but it isn't hard for us to.

To do that, instead of writing directly to the output file, you can first write to an `io.BytesIO()` object.  Then you can use `hashlib.sha256(out.getvalue()).hexdigest()[:16]` to compute a hash and use the first 16 hex digits (8 bytes) of it.

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,354 @@
...
+            "bug",
+            "example",
+            "quote",
+        ):
+            app.add_directive(
+                f"si-{admonition}",
+                get_directive_class(admonition, _(admonition.title())),
+            )
+
+    user_defined_admonitions = getattr(config, "sphinx_immaterial_custom_admonitions")
```
jbms, line 277: Here you can normalize `sphinx_immaterial_custom_admonitions` as follows:

```python
config.sphinx_immaterial_custom_admonitions = pydantic.parse_obj_as(List[CustomAdmonition], config.sphinx_immateiral_custom_admonitions)
```

sphinx_immaterial/__init__.py
```diff
@@ -130,7 +129,12 @@ def onerror(filename: str, error: Exception) -> None:
                             )
                         excluded = sphinx.util.matching.Matcher(excluded_list)
                     else:
-                        excluded = sphinx.util.matching.DOTFILES
+                        excluded = sphinx.util.matching.Matcher(
```
jbms, line 132: If someone inherits from this theme, then anything matching stylesheets will also be excluded.  Instead, we could modify the build script to place the stylesheets outside of the `static` directory.

2bndy5: What location did you have in mind? Reverting this change will go back to copying everything except paths that begin with `.`. So, maybe have the build script put them in a `.theme_css` folder?

jbms: Perhaps instead of `sphinx_immaterial/static`, the css could be written to `sphinx_immaterial/bundles` -- we might eventually want to do something similar for the bundled javascript.

docs/admonitions.rst
2bndy5: >  Therefore, it might be better for this option to just control whether we provide the xxx directives

I'll revert the admonition name changes, so the option will be more useful about polluting directives' names.

>  Perhaps instead of this option, we could provide a way for the user to specify these admonitions via `custom_admonitions`, e.g. with an option to have the admonition directive just map to an existing class rather than define a new class.

I too was wondering how we could go about this. I hadn't considered adding a config option though, that would make things simpler. I think a `default_class` option would satisfy this well.

sphinx_immaterial/__init__.py
2bndy5: Just so I understand this correctly, you still want the built CSS bundles copied to `outdir` even though they aren't referred to by the base.html template anymore? I'm a little unclear about what "inherits from this theme" means.

sphinx_immaterial/custom_admonitions.py
2bndy5: I noticed that there will be multiple CSS files written to the output if re-building the docs with a change to the custom admonitions config. I think the expected behavior is to remove the outdated generated CSS files in this case.

docs/admonitions.rst
2bndy5: Done. I chose the name `classes` to accept a list of CSS class names. 
```py
sphinx_immaterial_custom_admonitions = [
    {
        "name": "my-note",
        "classes": ["note"],
    }
]
```
Although I'm not sure how useful it would be to specify multiple class names for a single directive...

sphinx_immaterial/custom_admonitions.py
jbms: It looks like pydantic has a built-in validation mechanism: https://pydantic-docs.helpmanual.io/usage/validators/

sphinx_immaterial/__init__.py
jbms: Sphinx theme's have an "inheritance" mechanism --- the `inherit` field in `theme.conf`.  Our theme inherits from `basic`, although perhaps we should take that out since I think in practice we don't actually use anything from `basic`.

I think inheriting just means the static files, and the HTML templates, are inherited.

I don't know if currently there are any themes that inherit from sphinx-immaterial, but in the future there might be.

No, I don't want the original built CSS bundles  to be copied to the outdir even though they won't be used --- I was just proposing that we change our build script to output them to somewhere other than `static`, so that they don't get copied automatically.  Then we don't have to change the logic here.

sphinx_immaterial/custom_admonitions.py
jbms: This is a general problem with Sphinx's build mechanism when not using a fresh output directory for each build invocation --- stale files from a prior build may remain.  For example, if you delete `xxx.rst`, `xxx.html` may remain in the output directory from a prior build.

We could add special logic to delete the stale CSS files, but I'm not sure it is particularly useful to do so, given that stale files can still remain for other reasons.

For the apigen modules, this problem is more severe because due to sphinx limitations we have to write files to the source directory, not just the output directory.  In that case we do have special logic to delete the stale files, but it is also a bit more trouble to get a fresh source directory for every build, than to get a fresh output directory for every build.  Additionally, the generated files in the source directory get processed by sphinx and therefore impact the result, while extra files in the output directory just get ignored.

sphinx_immaterial/__init__.py
2bndy5: Ah yes, I had forgotten about inheriting from other sphinx themes like that. I think we could get rid of the basic theme dependency and reduce the logic in this function a bit more.

If you don't want them copied and keep the old logic here, then they should be placed in a folder that begins with `.`. I'll play around a bit, but I think that (using the old logic) wherever we put the bundles in the theme distribution, they will get copied to the same relative location in outdir. The old logic ignored any distribution files whose path begins with `.`, that's why the icons aren't copied to the outdir.

sphinx_immaterial/custom_admonitions.py
2bndy5: Yeah, it's not a big concern since the problem would only occur when connective builds have a change in the amount of `si-icon`s or a change in custom admonitions. In practice I think most people would be hosting docs that were made from a fresh build.

sphinx_immaterial/custom_admonitions.py
2bndy5: Oh cool! I'll try that out. I'm not a big fan of the way the pydantic docs are written (lead by example then list what wasn't shown in the example).

2bndy5: Done. I also moved to documenting the class attributes now that all are optional except the `name` field.

docs/admonitions.rst
2bndy5: I moved all inherited approaches into a separate section titled "Legacy Approaches", and I gave examples for how to migrate to the custom admonitions feature.

sphinx_immaterial/__init__.py
2bndy5: >  Our theme inherits from basic, although perhaps we should take that out since I think in practice we don't actually use anything from basic.

I looked into this. There is a way to not inherit from any sphinx thene, but there are some templates that we're using.

1. [page.html](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/page.html) corresponds to the default value for the `templatename` param in [`StandaloneHTMLBuilder.handle_page()`](https://github.com/sphinx-doc/sphinx/blob/cd3f2e435000835dd98a11497440bc16a79ec31c/sphinx/builders/html/__init__.py#L1021). I was able to override this, but it led to other templates that were missing.
2. [genindex.html](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/genindex.html) - This is where I stopped looking into not inheriting from the basic theme. If it involves creating templates specific to Sphinx, then it defeats the point of not deviating from upstream templates. I'm not willing to forego support for these Sphinx-specific templates because... well you remember #132 inquired about supporting genindex and the like.

---
> Perhaps instead of sphinx_immaterial/static, the css could be written to sphinx_immaterial/bundles 

Using sphinx_immaterial/bundles worked. I must confess, I still don't fully understand how sphinx determines what files to copy...

sphinx_immaterial/__init__.py
```diff
@@ -272,19 +271,35 @@ def setup(app):
         "html_use_directory_uris_for_index_pages", False, rebuild="html", types=bool
     )
 
-    app.add_builder(_get_html_builder(app.registry.builders["html"]), override=True)
-    app.add_builder(_get_html_builder(app.registry.builders["dirhtml"]), override=True)
+    app.add_builder(
```
jbms, line 274: Perhaps change to for loop to avoid duplication

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,377 @@
...
+    # validate user defined config for custom admonitions
+    user_defined_admonitions = pydantic.parse_obj_as(
+        List[CustomAdmonitionConfig], getattr(config, confval_name)
+    )
+    setattr(config, confval_name, user_defined_admonitions)
+    user_defined_dir_names = [directive.name for directive in user_defined_admonitions]
+
+    # override the specific admonitions defined in sphinx and docutils
+    # these are the admonitions that have translated titles in sphinx.locale
+    for admonition, title in admonitionlabels.items():
```
jbms, line 303: I think we should have a config option to disable this, in case it conflicts with another extension that wants to modify these admonitions.  Also for the `admonition` directive itself.

sphinx_immaterial/custom_admonitions.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+def on_config_inited(app: Sphinx, config: Config):
+    """Add admonitions based on CSS classes inherited from mkdocs-material theme."""
+
+    # override the generic admonition directive
+    app.add_directive("admonition", get_directive_class("admonition", ""), True)
+
+    # generate directives for inherited admonitions from upstream CSS
+    if getattr(config, "sphinx_immaterial_generate_inherited_admonitions"):
```
jbms, line 278: Perhaps instead of "inherited" this should be "extra"?

2bndy5: This makes sense from a user perspective.

sphinx_immaterial/custom_admonitions.py
2bndy5: It really shouldn't conflict with other extensions, but I do understand the prudence. The syntax is practically the same (except for the optional arg). The visitor is still monkeypatched regardless.

While I'm at it, I'll add another config option to disallow overriding the built-in/specific admonitions in sphinx/docutils.

jbms:
Looks good to me, thanks for all your work on this!

PR #188, opened 2022-11-10
doc use of Giscus; add comments meta data tag

2bndy5:
Allows integration of page comments using [giscus](https://giscus.app/) (which uses github's discussions feature). Only thing we needed was a `comments` meta data field to enable comments for a specific page. Everything else is done by the user in the docs project's template overrides, which is now documented with instructions in customization.rst.

Tested this on a separate repo of mine (hosted at RTD.org), and it seems to work. It will not work when the URL is a `file://`. So, there's a preview at the bottom of [this page](https://circuitpython-homie.readthedocs.io/en/trying-doc-comments/tutorials/openhab.html#adding-a-homie-dev-as-an-oh-thing).

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -138,6 +138,18 @@ Each metadata is evaluated as a :rst:`:key: value` pair.
 
         :hide-feedback:
 
+.. themeconf:: show-comments
+
+    If specified, allows comments to be enabled at the bottom of the current page.
+
+    .. code-block:: rst
+        :caption: Enable comments at the bottom of the page:
+
+        :show-comments:
+
+    .. seealso::
+        Using comments requires extra steps. See `Adding a comment system`_ instructions.
+
 Configuration Options
 =====================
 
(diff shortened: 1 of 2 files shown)


2bndy5:
Giscus can derive a discussion's title from meta data, but we haven't been supporting the `meta` directive in a way that the inherited html templates can render the neccesary `<meta>` tags. So, I've added the necessary info to the pages' template context.

For example, users can specify the title of the github discussion to be embedded in a page by using the following rST source:
```rst

:show-comments:

.. meta::
    :property=og\:title: discussion name
```

---
Beneficially, I think this meta tag support also allows adding keywords/tags for search results:
```rst
.. meta::
    :keywords: aTag
    :keywords lang=en: english tags
```

docs/customization.rst
```diff
@@ -781,6 +793,92 @@ Lastly, make sure the project's documentation ``conf.py`` has the following line
...
+           data-category-id="..."
+           data-mapping="title"
+           data-reactions-enabled="1"
+           data-emit-metadata="1"
+           data-theme="light"
+           data-lang="en"
+           crossorigin="anonymous"
+           async>
+         </script>
+2. Override the theme's partials/comments.html template (blank by default) in your documentation
```
jbms, line 829: It seems rather unfortunate to ask the user to add all of this boilerplate.  But I see that this example comes from mkdocs-material.

2bndy5: Yeah. I'm not sure how much would change if using different comment systems (like Discus), but doing it this way allows the user more flexibility when following the docs for the chosen comment system.

jbms:
Looks good to me.

PR #191, opened 2022-11-15
adjust regressive z-index override from #171

### LINKED ISSUE #189: left TOC z-index incorrect in wide viewport with `toc.follow`
Small issue, but I noticed the left TOC from `python-apigen` generated pages have a display issue. It appears the contents of the TOC appear on top of the page title "API Reference" (in my case) when the TOC is long enough.

### Example

https://mhostetter.github.io/galois/latest/api/galois.ReedSolomon.k/

Scroll in the left TOC and observe the overlay on top of "API Reference".

image

2bndy5:
fixes #189 by using the inline CSS variable with fallback value inherited from upstream

### DIFF
--- a/src/assets/stylesheets/main/layout/_nav.scss
+++ b/src/assets/stylesheets/main/layout/_nav.scss
@@ -678,7 +678,7 @@
           > .md-nav__link {
             position: sticky;
             top: 0;
-            z-index: 1;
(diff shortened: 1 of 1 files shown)


jbms:
Looks good to me, thanks.

PR #205, opened 2023-01-12
Improve CSS/JavaScript bundling and fix miscellaneous issues

### LINKED ISSUE #193: Client Error: Bad Request for url when fetching some Google Fonts.
**Description**
I would like to use the _Recursive_ Google Fonts but when I configure the `html_theme_options`, [fonts](https://jbms.github.io/sphinx-immaterial/customization.html#themeconf-font), I get a Sphinx-Immaterial extension error.

**What's it all about?**
I want to use the _Recursive_ font as the text font in my project but I keep getting a `400 Client Error: Bad Request url` error when I build my Sphinx project.
From my research, I realised that the way Sphinx-Immaterial requests fonts from Google Fonts won't always work because some fonts do not provide other font styles like italics.

For example:

In my project, I want to use the Recursive font which does not provide [italic versions](https://fonts.google.com/specimen/Recursive?query=recurs) of the font.
As a result, if Sphinx-Immaterial tries fetching this URL: https://fonts.googleapis.com/css?family=Recursive:700i, it will get a `Bad Request error (status-code: 400)`.

**What should happen:**
Sphinx-Immaterial downloads the Google Fonts specified by the user, caches them (using the cache specified by [sphinx_immaterial_external_resource_cache_dir](https://jbms.github.io/sphinx-immaterial/customization.html#confval-sphinx_immaterial_external_resource_cache_dir)), and includes the fonts in the built documentation.

**What happened instead:**
I get the following Sphinx-Immaterial extension error:

```bash
Extension error (sphinx_immaterial.external_resource_cache):
Handler <function _builder_inited at 0x7fd07afc7370> for event 'builder-inited' threw an exception (exception: 400 Client Error: Bad Request for url: https://fonts.googleapis.com/css?family=Recursive:700i)
```
**How to reproduce this bug:**
1. Create `projectA`:
   ```bash
   mkdir projectA
   cd projectA
   sphinx-quickstart 
   ```
2. Install Sphinx-Immaterial and configure `projectA` by adding the following lines to `projectA/source/conf.py`:
   ```python
   extensions = ["sphinx_immaterial"]
   html_theme = "sphinx_immaterial"
   html_theme_options = {
	# Other Sphinx-Immaterial configuration
	"font": {
		"text": "Recursive"
	},
   }
   ```
3. Build documentation using `make html` or `sphinx-build -b html . ./_build/html`

**Browser(s) and Device(s) observed on:**
- Device:  Ubuntu Linux
- Browser: Google Chrome
- Python version: `3.9.13`
- Sphinx version: `5.3.0`
- Sphinx-Immaterial: `0.11.0`

I hope the description helps to reproduce the issue :)

jbms:
- Use Google Fonts metadata to fetch all available variants
    
  Previously, this theme retrieved a hardcoded list of font variants, as
  specified by the mkdocs-material theme.  However, it was reported in #193
  that some fonts do not have all of these variants, leading to an error
  when trying to use such a font.

  With this change, the precise list of variants for each font is used
  by fetching the complete list of fonts and variants from the Google
  Fonts API.

  This commit also factors out the Google Fonts retrieval logic from
  `external_resource_cache.py` into a separate `google_fonts.py` module.

  Fixes #193.

- Refactor CSS and JavaScript bundle output and support source maps
    
  This refactors the existing CSS bundling code in
  `custom_admonitions.py` in the following ways:

  - JavaScript bundling is supported in addition to CSS.

  - The `add_global_{css,javascript}` functions are added to allow other
    modules to register additional sources.

  - Source maps are supported, and are generated when
    `sphinx_immaterial_bundle_source_maps = True`.

  - The new `add_global_css` function is used to add the Google Fonts
    CSS to the CSS bundle, rather than as a separate .css file.

- Properly include required custom icons in parallel builds

- Exclude sphinx_highlight.js file from basic theme
    
  The file is not needed by this theme.

- Fix typos in Python domain customization documentation

### DIFF
--- a/sphinx_immaterial/css_and_javascript_bundles.py
+++ b/sphinx_immaterial/css_and_javascript_bundles.py
@@ -1,0 +1,200 @@
+"""Bundles CSS and JavaScript resources."""
+
+import hashlib
+import json
+import os
+import pathlib
+from typing import NamedTuple, Optional, List
+
+import sphinx.application
+import sphinx.environment
+
+
+class Entry(NamedTuple):
+    code: str
+    sourcemap: Optional[str]
+    priority: int
+
+
+_CSS_KEY = "_sphinx_immaterial_global_css"
+_JAVASCRIPT_KEY = "_sphinx_immaterial_global_javascript"
+_BUNDLE_SOURCE_MAPS_KEY = "sphinx_immaterial_bundle_source_maps"
+
+
+def _get_css_entries(app: sphinx.application.Sphinx) -> List[Entry]:
+    entries = getattr(app, _CSS_KEY, None)
+    if entries is None:
+        entries = []
+        setattr(app, _CSS_KEY, entries)
+    return entries
+
+
+def _get_javascript_entries(app: sphinx.application.Sphinx) -> List[Entry]:
+    entries = getattr(app, _JAVASCRIPT_KEY, None)
+    if entries is None:
+        entries = []
+        setattr(app, _JAVASCRIPT_KEY, entries)
+    return entries
+
+
+def add_global_css(
+    app: sphinx.application.Sphinx,
+    code: str,
+    sourcemap: Optional[str] = None,
+    priority: int = 500,
+):
+    _get_css_entries(app).append(
+        Entry(code=code, sourcemap=sourcemap, priority=priority)
+    )
+
+
+def add_global_javascript(
+    app: sphinx.application.Sphinx,
+    code: str,
+    sourcemap: Optional[str] = None,
+    priority: int = 500,
+):
+    _get_javascript_entries(app).append(
+        Entry(code=code, sourcemap=sourcemap, priority=priority)
+    )
+
+
+def generate_bundle(
+    app: sphinx.application.Sphinx,
+    env: sphinx.environment.BuildEnvironment,
+    entries: List[Entry],
+    source_mapping_url_prefix: str,
+    source_mapping_url_suffix: str,
+    output_prefix: str,
+    output_ext: str,
+) -> str:
+    entries.sort(key=lambda e: e.priority)
+
+    # Use the special "sections" support in Source Map v3 to combine source maps
+    # without needing to decode them.
+    #
+    # https://sourcemaps.info/spec.html
+    sourcemap_sections = []
+    lines: List[str] = []
+
+    entries.sort(key=lambda entry: entry.priority)
+    for entry in entries:
+        entry_lines = [
+            x
+            for x in entry.code.rstrip().splitlines()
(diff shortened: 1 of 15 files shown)


2bndy5:
Thanks for doing this! I was clearly fumbling around with the generated bundles...

Is the Google API key here allowed to be openly redistributed?

jbms:
> Thanks for doing this! I was clearly fumbling around with the generated bundles...
> 
> Is the Google API key here allowed to be openly redistributed?

This key is already effectively public --- if you visit

https://developers.google.com/fonts/docs/developer_api

open Dev Tools and click Execute in the "API explorer" side bar, you will see that a request is made from the client with this exact key.

In general I don't know why the Google Fonts API requires a key at all, since the entire API that requires a key just consists of receiving a single static 130KB JSON file, and the fonts themselves (which amount to a much larger amount of data and require more sophisticated request handling) can be retrieved without an API key.

It is also relatively easy to just create a Google Cloud project and obtain an API key (there are no charges), but I didn't want to bother with that as long as this key works.

sphinx_immaterial/css_and_javascript_bundles.py
```diff
@@ -0,0 +1,199 @@
...
+    css_bundle = generate_bundle(
+        app=app,
+        env=env,
+        entries=css_entries,
+        source_mapping_url_prefix="/*# sourceMappingURL=",
+        source_mapping_url_suffix=" */",
+        output_prefix=os.path.join(app.outdir, "_static", "sphinx_immaterial_theme"),
+        output_ext="css",
+    )
+    app.add_css_file(css_bundle, priority=200)
```
2bndy5, line 156: I think the resulting CSS path needs to be made relative to the `app.outdir`. The [RTD build looks broken](https://sphinx-immaterial--205.org.readthedocs.build/en/205/), but the CSS loads if I change the path (in the live DOM) to be relative.
```diff
-<link rel="stylesheet" type="text/css" href="/home/docs/checkouts/readthedocs.org/user_builds/sphinx-immaterial/checkouts/205/docs/_build/html/_static/sphinx_immaterial_theme.b26c02e5c3d38b40b.min.css">
+<link rel="stylesheet" type="text/css" href="_static/sphinx_immaterial_theme.b26c02e5c3d38b40b.min.css">
```

jbms: Thanks for catching that --- should be fixed now.

2bndy5: Indeed, it is fixed.

2bndy5 (approved):

PR #213, opened 2023-01-22
Fix source info

jbms:
- Remove unused imports and enable unused-imports lint

-  Add monkey patch to fix source file/line number info for object descriptions

   - This addresses the issue described at https://github.com/sphinx-doc/sphinx/pull/11147.

- Disable apidoc/python/section_titles monkey patch for Sphinx >= 5.3
    
  - This is unnecessary due to
    https://github.com/sphinx-doc/sphinx/issues/10887, which adds support
    for section titles in all object description content.

### DIFF
--- a/sphinx_immaterial/apidoc/python/section_titles.py
+++ b/sphinx_immaterial/apidoc/python/section_titles.py
@@ -1,8 +1,13 @@
 import docutils.nodes
 import docutils.statemachine
 
+import sphinx
 from sphinx.domains.python import PyObject
 import sphinx.util.nodes
+
+
+# Not needed in Sphinx 5.3
+assert sphinx.version_info < (5, 3)
 
 
 def _monkey_patch_python_domain_to_support_titles():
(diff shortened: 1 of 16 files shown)


sphinx_immaterial/apidoc/python/section_titles.py
```diff
@@ -1,10 +1,15 @@
 import docutils.nodes
 import docutils.statemachine
 
+import sphinx
 from sphinx.domains.python import PyObject
 import sphinx.util.nodes
 
 
+# Not needed in Sphinx 5.3
+assert sphinx.version_info < (5, 3)
```
2bndy5, line 10: Wouldn't this break usage with Sphinx prior to v5.3?

Is this file not imported elsewhere (making `assert` ok)?

jbms: This is only imported from `sphinx_immaterial/apidoc/python/default.py` where I've added a check of `sphinx.version_info`.

2bndy5 (approved):

PR #220, opened 2023-02-21
ci: do not fail fast

umarcor:
Currently, if any job of the matrix fails, all others are cancelled automatically. However, some versions of Python might work while other fail. Or some versions might works on some platforms but fail on others.

This PR unsets `fail-fast`, in order to allow all jobs to finish regardless of one or some of them failing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on: [push, pull_request, workflow_dispatch]
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


2bndy5:
pending #224 for CI failures

2bndy5:
@umarcor I think this branch needs to be re-based to include the fix from #224 .

2bndy5 (approved):

PR #224, opened 2023-02-28
Address pylint black updates

2bndy5:
- addresses warnings emitted by pylint v2.12.2
- ran black v23.1.0 on all py files

PR #225, opened 2023-02-28
allow use of system font in theme's graphviz ext

2bndy5:
This fixes regression about using system font when this theme's graphviz extension is enabled. I added a unit test to ensure future continuity. Also adjusted the docs about the `html_theme_options["font"]` field when using the system's available fonts.

EDIT: Now includes detailed step-by-step instructions for using a self-hosted font family.


2bndy5:
pending #224 for CI failures.

docs/customization.rst
```diff
@@ -429,6 +429,63 @@ Configuration Options
...
+               Always define fonts with the CSS variables shown above and not with
+               :css:`font-family` because :css:`font-family` uses the theme's CSS variables that
+               define the system font fallback.
+
+        .. rst-example::
+            :class: custom-font-example
+
+            This text is just an **example**.
+            *Notice* the font family used is different (monospaced) in the code snippet.
+
```
jbms, line 488: Perhaps add a note that the graphviz extension won't use the custom font.

Perhaps we should also allow the user to continue using `html_theme_options['font']` to set the fonts even when using custom fonts, and we should just have a config option to disable injecting the google font?  Then the user wouldn't need the extra `:root {` font block that kind of exposes implementation details of the theme.

docs/_static/custom_font_example.css
```diff
@@ -0,0 +1,64 @@
+/* comic-neue-latin-300-normal */
+@font-face {
```
jbms, line 2: Perhaps we can reduce the number of variants here, to avoid adding to the repo size unnecessarily?

2bndy5: I wasn't sure which weights I needed, but it is a good idea... I'll look into it.

jbms: I think if you skip variants the browser will just synthesize them, so the extra variants just increase the quality.  The default is 400.  But I have not verified to confirm.

docs/customization.rst
2bndy5: > Perhaps add a note that the graphviz extension won't use the custom font.

Will do.

> Perhaps we should also allow the user to continue using `html_theme_options['font']` to set the fonts even when using custom fonts

That's what the docs are actually doing for the limited example.

> we should just have a config option to disable injecting the google font?

Isn't that the same as using `html_theme_options['font'] = False`? I'm kinda hazy on how the `False` value is handled, but it does seem to work as expected in upstream.

> Then the user wouldn't need the extra `:root {` font block that kind of exposes implementation details of the theme.

I think the more important implementation detail is overriding the `--md-*-font-family` CSS vars as that is where fallback fonts are specified. I added the warning about not overriding the `font-family` property because it basically uses the `--md-*-font-family` value.

In creating the docs' example. I noticed the CSS specificity was an issue for a narrow scope (of custom/self-hosted font). I left a comment in the extra_css.css mods to explain why I had to override the `--md-*-font-family` vars and the `font-family` property... It works fine if the `--md-*-font` vars are altered at the doc's `:root` scope, but the example I used in the docs seems like a edge case exception.

---
Since this approach is the one recommended upstream, I think it'll do for now. But I would like to consider an idea that removes all the boilerplate CSS with some extra fields like
```py
html_theme_options = {
    "font": {
        "custom_text": "font-name",
        "custom_code": "mono-font-name",
    }
}
```
or maybe something similar to our custom admonitions config.

jbms: Well here it is unusual since you are just using the custom font in one place.  But I'm imagining that we have something like:

`sphinx_immaterial_inject_google_font = False`

that would prevent us from injecting the font specified by `html_theme_options["font"]`.  However, the font family would still get set to the specified fonts in the normal way, so they should either be system fonts or defined by the user with custom css as in your example.

It is true this wouldn't save much but it would mean the user wouldn't need to create a custom css file just to use a system font.

As for `"custom_text"` and `"custom_code"` --- did you have in mind exactly what I'm describing as far as `sphinx_immaterial_inject_google_font = False`, or do you mean somehow also adding css rules to define the font?

2bndy5: > the user wouldn't need to create a custom css file just to use a system font.

Oops, I should've made this clearer. No custom CSS is needed for using system fonts; only `"font": False` is needed for that. Or are you referring to a project that may want to use a system font not specified in the theme's fallback options? Because that would simply require the system fonts' name in `:root { --md-*-font: "name"; }`.

> or do you mean somehow also adding css rules to define the font?

Yes. One of the hardships I encountered when putting the example together was the CSS `url()`  path to the font file. I tried to make as much explicit in the docs, but there is clearly possible points of error in upstream's recommended approach.

hille721: Indeed it expects a custom.css when I set `font: false`. But not a blocker during build it will only try to load custom.css when accessing it and gives a 404 in developer mode.
I solved that by placing an empty `_static/custom.css`.

2bndy5: Again with the useful feedback! I didn't notice a 404 in the dev console. I will have to investigate and amend the docs about this.

@hille721 were you using chrome? I mainly just test in firefox...

hille721: Edge, thus kind of Chrome.

sphinx_immaterial/graphviz.py
```diff
@@ -235,10 +235,16 @@ def render_dot_html(
 ) -> Tuple[str, str]:
 
     theme_options = self.builder.config["html_theme_options"]
-    ttf_font_paths = google_fonts.get_ttf_font_paths(self.builder.app)
-    font = theme_options["font"]["text"]
+    if theme_options["font"]:
+        font = theme_options["font"]["text"]
+    else:
+        font = "Helvetica"
```
jbms, line 241: If we allow `html_theme_options["font"]` to be used even without injecting google fonts, then we can have graphviz still use the system or user-defined font in its output, just with the metrics of Helvetica.  I think that would be better than always actually displaying Helvetica if the user disables google fonts.

docs/customization.rst
jbms: That is odd -- not sure where `custom.css` is coming from....

sphinx_immaterial/graphviz.py
2bndy5: I did this to avoid
```
>       font = theme_options["font"]["text"]
E       TypeError: 'bool' object is not subscriptable
```
when font is false. This could be handled better though.
```py
    font: str = "Helvetica"
    if isinstance(theme_options["font"], dict) and "text" in theme_options["font"]:
        font = theme_options["font"]["text"]
```
Or did you have something else in mind?

docs/customization.rst
2bndy5: In loading the generated docs from this PR's added unit test, I don't see any 404 error. I do see a JS error (in all browsers: edge chrome, firefox), but that JS error is related to the unit test having a bare minimal config (namely an undefined `html_theme_options["palette"]`).

@hille721 You might want to check your conf.py for a `html_css_files` setting. If that is set to `["custom.css"]`, then I would expect the docs' src folder to have a  `_static/custom.css` file.

hille721: @2bndy5 : ok my bad, not directly related to sphinx-immaterial. It seems to be somehow a problem when combining sphinx-immaterial with [sphinx-argparse-cli](https://pypi.org/project/sphinx-argparse-cli/). Was able to reproduce in a fresh environment:

``` python
project = 'test'
copyright = '2023, ch'
author = 'ch'
extensions = [
    "sphinx_argparse_cli",
    "sphinx_immaterial"
    ]
templates_path = ['_templates']
exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
html_static_path = ['_static']
html_theme = 'sphinx_immaterial'
```

``` bash
sphinx==6.1.3
sphinx-argparse-cli==1.11.0
sphinx-immaterial==0.11.3
```

image

2bndy5: Yep, I [found the problem in that ext's src](https://github.com/tox-dev/sphinx-argparse-cli/blob/4a542ff54cb34e6400fece0e320e37025d4514bf/src/sphinx_argparse_cli/__init__.py#L9). Although I don't see a place where the `custom.css` is generated or distributed in that ext. It is worth [opening an issue there](https://github.com/tox-dev/sphinx-argparse-cli/issues), but the repo doesn't look very active.

docs/_static/custom_font_example.css
2bndy5: Default is 400, and it looks like rxjs might be doing some kind of synthesis with the font 🤷🏼‍♂️ .

I have removed all but the 400 weight for "Comic Neue" and "Comic Mono", and the output looks acceptable for an example. There's a slight decrease in quality but whatever.

sphinx_immaterial/graphviz.py
2bndy5: > then we can have graphviz still use the system or user-defined font in its output, just with the metrics of Helvetica

My main confusion is here. Sorry this post got real long...

The user doesn't need to explicitly specify a system font name. Rather, `font = False` will use whatever the theme's CSS var has set for system-specific fallbacks.
https://github.com/jbms/sphinx-immaterial/blob/cf059973255ee2595005765b941c81b47bbcc380/src/assets/stylesheets/main/_typeset.scss#L32-L35
Because **build system's offerings can differ from client system's offerings**, the exact system font used during browsing is unknown at build time when `font = False`.

Furthermore, as I mentioned before, explicitly setting a system font name would be done in user-defined CSS:
```css
:root {
  --md-text-font: "Arial";
}
```
Note, this var isn't required to be in `:root {` and can be specified more than once.
Because `font` is still `False` in this scenario, we have no easy way of telling graphviz which system font to use at build time. We could crawl the user-defined CSS for this var, but that would lead to a guessing game if more than 1 `--md-text-font` var is specified under different CSS selectors.

---
I suppose I could exclude the `-<x>fontname={ttf_font}` values
https://github.com/jbms/sphinx-immaterial/blob/cf059973255ee2595005765b941c81b47bbcc380/sphinx_immaterial/graphviz.py#L274-L283
when no google fonts are used or a system font is implied via `font = False`. But, I think that incites graphviz to use Helvetica (or Arial) by default anyway. I'm hoping this is the intentionally requested change because I'm rather confused.

docs/customization.rst
hille721: I think `custom.css` is normally coming with plain Sphinx (alabster):

image

Seems to be they are providing kind of a placeholder `custom.css`.

Will open a issue there as I also just realized that this problem occurs also when using the popular rtd theme.

2bndy5: hold up!

We actually don't inherit CSS from sphinx basic theme's (like many other popular sphinx theme's actually do), so we may need to consider a solution for this on our end.

2bndy5: I still don't see where custom.css is coming from the sphinx codebase. I do see a few assertions that the file exists in some of sphinx' unit tests, but I'm having trouble tracking it down.

Either way, I'm not convinced this is a sphinx-immaterial problem.

hille721: yeah as I said, problem occurs also with rtd theme. Already created a issue there (https://github.com/tox-dev/sphinx-argparse-cli/issues/82)

sphinx_immaterial/graphviz.py
2bndy5: Now that the CI was working through the linting/format checks, I see how my original approach only worked on Windows. So, I ended up only adding the `-<x>fontname={ttf_font}` values if `ttf_font` is not None, and prevented modifying the SVG elements' `font-family` attribute when `font` is None.

The unit test output renders fine: 
image
And the CI seems happy now.

docs/customization.rst
2bndy5: I think at some point sphinx abstracted the alabaster theme as a dependency...

2bndy5: Ok, so getting back on track here...

I added the note about the graphviz ext in the the graphviz.rst doc (with a link back to the `:themeconf:'font'` doc) when not using a Google font.

As for the rest, I think we were beginning to discuss a more user-friendly feature about specifying a custom or self-hosted font. But I feel like there was a disconnect between written word and my own comprehension. Maybe we should start a new issue to discuss in detail?

sphinx_immaterial/graphviz.py
jbms: Note that there are really two fonts used for the graphviz extension: the font whose metrics are used by graphviz itself to place things appropriately in the svg output ("metrics font"), and the font that actually gets displayed in the browser ("display font").  Ideally these two fonts are the same, but even if we just use the metrics from a default font like Helvetica, it only makes a small difference and isn't a big deal.

I agree that if the user is just setting the font manually via CSS, then we don't have a robust way of knowing what the font will be at build time and therefore have no hope of specifying a "metrics font".  Additionally, even if we did know what the font will be, we wouldn't be able to tell graphviz how to load it anyway.  I think we could still use the custom font as the "display font", though, since we could just set the font to `var(--md-text-font-family)`.  I think that would be better than just using the default graphviz font.

2bndy5: > we could just set the font to `var(--md-text-font-family)`

👀 I'll give that a try. It might even resolve the system font fallback used in CSS (for browsers only, not font metrics obviously).

2bndy5: Ok adding `var(--md-text-font-family)` to the inline style of SVG nodes that have the font-family attr set works to render the graph with the system font (if not using a cached google font):
image
Notice the graphviz default  font metrics are still used. In the above rendering, the graph nodes' `text` elements use `font-family = "Times New Roman"`, but the inline `style="font-family: var(--md-text-font-family);"` overrides the text elements' attr.
```xml
<text text-anchor="middle" x="27" y="-86.9" font-family="Times New Roman,serif" font-size="12.00"
  style="fill: var(--md-code-fg-color); font-family: var(--md-text-font-family)">A</text>
```

jbms: I think in case that we are setting `font-family` in the style we should also remove the `font-family` attribute.

docs/graphviz.rst
```diff
@@ -6,6 +6,12 @@ This theme adds several improvements to the built-in :py:mod:`sphinx.ext.graphvi
 - default fonts and colors are set to match the fonts and colors used by the
   theme, including support for both light and dark mode;
 
+  .. admonition:: Only compatible with Google fonts
+
+      When not using a Google font via :themeconf:`font` in conf.py (eg. using a
+      system/custom/self-hosted font), this extension will defer to the Graphviz default font
```
jbms, line 12: Will use the metrics from graphviz's default font , rather than the actual font, for the purpose of computing the graph layout.  However, the text will still be displayed in the browser using the user-specified font.

sphinx_immaterial/graphviz.py
2bndy5: done.

docs/graphviz.rst
2bndy5: done. It is now:
```rst
      When not using a Google font via :themeconf:`font` in conf.py (eg. using a
      system/custom/self-hosted font), this extension will use the metrics from graphviz's
      default font, rather than the actual font, for the purpose of computing the graph layout.
      However, the text will still be displayed in the browser using the user-specified font.
```

jbms (approved):

jbms:
Thanks!

PR #229, opened 2023-03-01
Fixed double numfig statements for Figues and no listing number for code blocks

glatosinski:
The explicit `add_fignumber` adds rendering of another figure number, resulting in double Fig entries in rendered doc. In addition, for code listings the Listing appears as a raw text above the code block:

double-fig

This PR removes the excessive number for tables and figures (code blocks require it). In addition, the Listing number, as well as listing caption are wrapped in `filename` class so they appear as listing content above the code block. The result is following:

double-fig-removed

### DIFF
--- a/sphinx_immaterial/apidoc/apidoc_formatting.py
+++ b/sphinx_immaterial/apidoc/apidoc_formatting.py
@@ -61,11 +61,15 @@ def visit_caption(
         # This is needed to trigger mkdocs-material CSS rule `.highlight .filename`
         self.body.append('<div class="code-block-caption highlight">')
         # append a CSS class to trigger mkdocs-material theme's caption CSS style
-        attributes["class"] += " filename"
+        # create a span wrapping caption's content and (optionally) number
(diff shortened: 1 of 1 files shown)


2bndy5:
I had noticed something similar to this (why the `else: super_func()` branch exists now). Is this specific to a certain version of sphinx?

2bndy5:
> Is this specific to a certain version of sphinx?

I checked with [the Sphinx source](https://github.com/sphinx-doc/sphinx/blame/ba080286b06cb9e0cadec59a6cf1f96aa11aef5a/sphinx/writers/html5.py#L425), and it doesn't seem effected by a particular version that this theme supports. And, the [docutils release notes](https://docutils.sourceforge.io/RELEASE-NOTES.html) don't mention any changes to captions.

2bndy5 (approved):

2bndy5:
Solution released in [v0.11.4](https://pypi.org/project/sphinx-immaterial/0.11.4/)

PR #230, opened 2023-03-02
add nox config; update CI

### LINKED ISSUE #223: update CI (again)
Opening this to discuss changes to the CI workflow...

- [x] Run Py and JS linting/format checks in a single job (agnostic of OS). OS specific checks (like unit tests and docs build) will require this job to complete first.
- [x] Use [nox](https://nox.thea.codes/en/stable/index.html) to pin certain versions of dev tools (namely the python tools like black, mypy, etc). ~There might be a way to incorporate the `npm run check` step with the `pre-commit run` command, but I'm not sure if this is preferable/possible.~ EDIT `npm run check` will still be used separately from pre-commit.
- [x] Generate code [coverage](https://coverage.readthedocs.io/en/7.2.1/) reports from unit tests (this may be optional but informative)
- [x] Avoid duplicate CI triggers in open PRs (using suggestion from https://github.com/jbms/sphinx-immaterial/pull/186#issuecomment-1304636511)

The main motivation here is to reduce the workflow run time by implementing task 1. Contributions from #220 and #219 will be retained.

2bndy5:
also ran pre-commit hooks via nox.

resolves #223


.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
+repos:
```
jbms, line 1: I think it is kind of unfortunate to duplicate versions and configurations here --- all we really need is like a 4-line shell script that invokes pylint, mypy, black, npm run check.  I guess pre-commit creates its own venv used only for running the checks?

2bndy5: >  I guess pre-commit creates its own venv used only for running the checks?

yep.

The added benefits I find with pre-commit is that I don't have to on the look out for
- CRLFs
- trailing whitespace
- yaml/toml/json syntax errors (which will always seem to crop up)
- EoF is a single blank line

I would not prefer to have an in-house script because it is just more work to maintain. I personally can't use a .sh script as that isn't friendly with Windows (without an env created in WSL which is significantly slower). Using pre-commit also provides some useful feedback in the event that a check failed.

Having the pre-commit config filled out also shows what deps are needed for which check. I suppose we could keep comments in the dev-requirements.txt file about this, but that is more of a nuisance to me.

---
In re-writing the CI, I noticed that the only thing needed from dev-requirements.txt is `wheel`, so that file is no longer used in CI.

2bndy5:
It also occurred to me that we could include the docs' build in the code coverage, but I didn't pursue that because it isn't as definitive as unit tests.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.0.1
+    hooks:
+      - id: mypy
+        # this hook's behavior doesn't respect toml config about `files` and `exclude`.
+        # we must re-specify using a regex pattern here.
+        files: "^sphinx_immaterial/.*\\.py$|^tests/.*\\.py$"
+        exclude: ".*/setup\\.py$"
+        additional_dependencies: [
+          pydantic, # needed for pydantic.mypy plugin (as respected from toml)
```
jbms, line 34: The duplication of our dependencies here and for pylint is the biggest thing I have an issue with.

There seems to be a `system` language option for pre-commit where it doesn't create a new venv -- perhaps that would allow us to avoid this duplication?

2bndy5: Interesting, I'll look into that `system` option...

2bndy5: ## system's tool versions used as is
So, using the system hook presents a problem which pre-commit is designed to solve by default:
- pre-commit hooks that use `system` env are ignoring the `rev` field which is required in the config. This means that any version of tool installed in the system env is used without validating that the tool's version matches the specified `rev` tag.

To summarize, the versions specified in both pre-commit config and dev-requirements.txt are even more required when using the system env.

> The duplication of our dependencies here and for pylint is the biggest thing I have an issue with.
>
>There seems to be a system language option for pre-commit where it doesn't create a new venv -- perhaps that would allow us to avoid this duplication?

Yes, using the system env does not mandate that `additional_dependencies` be specified in pre-commit config.

The `black` tool doesn't need any additional deps, but using the system env means we can't cache the `black` tool's install.

## Caching the pre-commit's hooks' venvs doesn't cache the system env's site_pkgs

With pylint and mypy hooks using pre-commit's isolated venvs, the specified `additional_dependencies` are cached as well. Consequently, this means the CI runner's cache (of mypy and pylint pre-commit venvs) could be out-of-date until the cache expires (which I think happens after a week) or until the pre-commit config changes.

If we want to use `actions/cache` to cache mypy and pylint tools' system install (which inherently need the theme's deps as well), then we should also use stricter version specs in the theme's requirements.txt. For this reason, it makes sense to use the non-cached system env for the mypy and pylint hooks and only cache the pre-commit venvs used for the other checks.

Remember, we've had problems in the past with "silent" updates to the black, so it  makes sense to cache a specific version of black via pre-commit hook's venv.

## hooks specific to pre-commit
I have found a way to use [pre-commit/pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks) to check for

- CRLFs
- trailing whitespace
- yaml/toml/json syntax errors
- EoF is a single blank line

without needing pre-commit installed, but this means the CI steps will go from 1 `pre-commit run --all-files` to a multitude of steps that does the same thing as pre-commit. Again, I don't want to maintain a generic script that does the same as pre-commit.

jbms: I agree that it would make sense to pin versions of all the tools in any case.  We don't want any usage of actions/cache to affect which versions are used (currently does not, I believe).

Note that on Windows outside of WSL you can still install/use bash --- e.g. a copy is included with git.

But probably it would be simplest to write the lint script in Python rather than shell script in any case.

2bndy5: It just seems redundant and more work to write a repo-specific script. I've seen it done in other repos and it isn't a set-and-forget option, especially if we pin theme deps to a specific version.

In writing the coverage steps, I thought it might be easier to run unit tests on various python versions (and deps' versions like sphinx) using [tox](https://tox.wiki/en/latest/). I'll look into that...

2bndy5: Not loving tox for its configuration... I think that [nox](https://nox.thea.codes/en/stable/index.html) could be the compromise we're looking for.

2bndy5: After exploring nox, I think it is the route that would be preferable in terms of writing our own script. It is so flexible that we can rewrite the entire CI in the noxfile.py (a config similar to pytest script) with the exception of the pypi deployments steps, So I'm wondering how much we want to put in the noxfile.py (`npm run <script>` can be performed as well).

I'll probably segregate the dev-requirements.txt using nested requirements.txt files for different nox sessions.
```
# in dev-requirements.txt

-r requirements/pylint.txt
-r requirements/mypy.txt
-r requirements/optional-cpp.txt
-r requirements/optional-json.txt
-r requirements/optional-keys.txt
```
This way I can keep the sessions' env cleaner/quicker rather than blindly installing everything. We could even use this in setup.py (or pyproject.toml) as well instead of re-specifying the optional dependencies.

jbms: I haven't used nox before, but it looks good.  I have used tox, but have not been all that happy with it.

2bndy5: I'm having fun with nox... Maybe too much fun 🤓.

Once I'm ready to commit, I want to figure out how to get nox to use the various [python versions from the github-hosted runners' tool cache](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#hosted-tool-cache). That way, when building the dists and docs, the CI won't have to create a separate job for each version of python, just a job for each OS. This idea may result in long CI runs though.

BTW, I included the docs' builds in the coverage data (locally), and the total coverage spiked to 84-89% (from 55-57% for only the unit tests) depending on which versions of sphinx, builder, and python I'm using in the nox session. This doesn't require the package installed because the `coverage` cmd adds the invoking directory to python's `sys.path`. I'm not sure if _all_ the monkey patching is observed in the coverage data though.

requirements/dev-pre_commit_hooks.txt
```diff
@@ -0,0 +1 @@
+pre-commit-hooks==4.4.0
```
jbms, line 1: is this still used?

noxfile.py
```diff
@@ -0,0 +1,228 @@
...
+
+@nox.session(python=False)
+def pylint(session: nox.Session):
+    """Run pylint using in default env"""
+    session.run("pip", "install", "-r", "requirements/dev-pylint.txt")
+    session.run("pylint", "sphinx_immaterial")
+
+
+@nox.session(python=False)
+def mypy(session: nox.Session):
```
jbms, line 35: I believe mypy results are Python version-dependent, so we should do this for all Python versions.

noxfile.py
```diff
@@ -0,0 +1,228 @@
...
+    """Checks formatting linting and typing"""
+    session.install("-r", "requirements/dev-black.txt")
+    session.run("black", ".")
+
+
+@nox.session(python=False)
+def pylint(session: nox.Session):
+    """Run pylint using in default env"""
+    session.run("pip", "install", "-r", "requirements/dev-pylint.txt")
+    session.run("pylint", "sphinx_immaterial")
```
jbms, line 31: Not sure if pylint results are python version-dependent, but if so we should also do this for all python versions.

requirements/dev-pre_commit_hooks.txt
jbms: Oh I see it is needed for check-json and similar rules, vermind.

2bndy5:
On a side note, I would've liked the coverage data to be combined from each OS-specific job, but the concurrency of the matrix in CI would definitely be a problem. So, I just have the CI create a coverage report for each OS.

noxfile.py
2bndy5: > I thought there might be a way to cache nox venvs without having any of the deps (which are set to use latest versions) in the cache, but I didn't look into it enough.

Nope. Anything installed while using a nox venv is installed to the venv. `session.run("pip install")` doesn't use the system env if using a nox venv. Note, `session.install()` can't be used if using the system env (when `@nox.session(python=False)`). So, we'll just have the CI ignore the nox venvs for mypy and pylint (only caching the black and pre-commit-hooks venvs).

PS - I think it would be overkill to run pylint and mypy with different versions of sphinx (like the unit tests are now doing).

.github/workflows/build.yml
```diff
@@ -17,135 +66,170 @@ jobs:
...
+      - name: Compare Windows wheel to Linux wheel
+        run: diff -ur --color=always pkg-Windows pkg-Windows
 
   python-publish-package:
     # Only publish package on push to tag or default branch.
     if: ${{ github.event_name == 'push' && github.repository == 'jbms/sphinx-immaterial' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main') }}
     runs-on: ubuntu-latest
-    needs:
-      - build
+    needs: [build, test]
```
2bndy5, line 216: I wasn't sure if it was desired to deploy to pypi _after_ the docs and unit tests complete/pass. Because the pkg build is now done in an isolated job matrix, I figured I'd keep the original behavior in CI even though the build jobs complete much earlier than the tests jobs.

.github/workflows/build.yml
```diff
@@ -17,135 +66,170 @@ jobs:
...
+        with:
+          path: "**/node_modules"
+          key: ${{ runner.os }}-${{ matrix.node-version }}-node_modules-${{ hashFiles('**/package-lock.json') }}
+      - run: npm install
+      - name: Check for dirty working directory
+        run: git diff --exit-code
+      - name: Build wheel
+        run: pipx run nox -s dists
+      - name: Show package version
+        run: python setup.py --version
```
2bndy5, line 97: I'm curious as to the intention here. I left it in because it doesn't prevent builds' success, but this step seems to add around 10-20 seconds to the CI runtime. Remember, the build jobs run concurrent to unit tests/docs jobs (as they no longer need the pkg installed), so it doesn't really matter if this is removed; I'm just curious.

noxfile.py
2bndy5: IDK if the python version effects the results. It might depending on certain deps used (c-ext libs come to mind). So, I just made pylint do the same that mypy does now (use all supported versions of python).

.github/workflows/build.yml
jbms: I think it is probably better to keep the original behavior.  The idea was that we only tag a release after manually verifying that the build (on main branch) was successful.  Therefore it is not necessary to wait again for the tests to complete (or even run them at all).  Also there might be a case where a test is known to be failing for some reason but we still want to tag a release.

.github/workflows/build.yml
jbms: The intent was to be able to check the logs to verify that the setuptools-scm version generation was working as expected.  I'm a bit surprised that it is taking a long time.  Maybe a call to `git describe` by setuptools-scm is taking a long time for some reason?

jbms:
> It also occurred to me that we could include the docs' build in the code coverage, but I didn't pursue that because it isn't as definitive as unit tests.

Yeah, in my experience coverage metrics are most useful when used carefully, e.g. measuring the coverage from just a specific set of unit tests known to actually validate the code's behavior.  It is easy to achieve high coverage by just executing the code without properly validating the behavior.

.github/workflows/build.yml
2bndy5: Thanks for the answer. The delay seems to the worst with macos and windows runners. It might something to do with the docker containers that github uses (trying to reference how the FS is exposed to the workflow).

In the bigger picture, I'm not worried about an extra 20 seconds when the docs take nearly 10 minutes to build with 3 different sphinx builders.

.github/workflows/build.yml
2bndy5: ok, so I'll leave this as is then

.github/workflows/build.yml
2bndy5: I can have nox output the name of the wheel using `GITHUB_OUTPUT` env var. This should give us the version used without needing to rerun setup.py again.

2bndy5:
Docs are now built after installing the wheel (as was done before).

I know that the build and test jobs could be combined, but using the concurrency of 2 separate jobs makes the CI finish faster. And, we don't need to install all supported versions of python to build the pkg and docs, so it still makes sense to run the unit tests separately.

jbms:
> On a somewhat related note, I'd like to use a env var to conditionally add the repo's root folder to the `sys.path` in conf.py, so I don't have to keep installing the theme when developing a solution.

I normally use `pip install -e .` --- that installs the package in development mode, where the "installed version" just points back to the source directory, so you can modify the source files without having to reinstall again.

2bndy5:
I can do that since it doesn't seem to go through the long steps of copying everything... I'll remove the env var stuff from the conf.py and noxfile.py

jbms:
This looks good to me --- the one thing you might consider is splitting up some of the configurations over multiple runs, especially on macos and windows where` it is taking 13-15 minutes.

2bndy5:
How so? You mean only certain sphinx versions tested on the macos and windows runners?

jbms:
> How so? You mean only certain sphinx versions tested on the macos and windows runners?

Previously we had a separate job for each python version and each platform.  That is probably excessive but we could split the macos and windows configurations across 2 or 3 different jobs, e.g. by specifying as part of the matrix configuration a parameter that will be passed to nox to restrict which configurations are tested.

2bndy5:
I can adjust the tests matrix to use different nox sessions...
```yml
test:
  strategy:
    matrix:
      include:
        - os: 'ubuntu-latest'
          nox-sessions: 'tests' # test all sphinx versions on all python versions
        - os: 'windows-latest'
          nox-sessions: 'tests-3.11'  # only test py 3.11 on windows
        - os: 'macos-latest'
          nox-sessions: 'tests-3.11(sphinx6)'  # only test sphinx v6 on python v3.11
  steps:
    run: nox -s ${{ matrix.nox-sessions }}
```

jbms:
Yeah exactly like that except that we can have multiple windows nacos entries in the matrix so that we still test all configurations, just in parallel rather than sequentially

2bndy5:
Ah. It will fail if a certain python version is not installed (or added to PATH) -- this behavior is the default when the env var `CI` is asserted. This will require an extra param to `nox` to allow skipping a python version that is not present in PATH.

I'm currently doing it sequentially because there's no reliable way to cache (or save as artifacts) the multiple coverage data files across concurrent jobs and then combine them when all tests are finished.

2bndy5:
I can cache each set of coverage data files and then use a second job to combine and report the results. But this would be _very_ long CI script since each concurrent job's cache will have to be restored individually (`actions/cache` keys doesn't support globs). I thought about using artifacts and downloading all artifacts after the tests are done, but this would include downloading all built docs and pkg artifacts.

jbms:
Artifact upload/download would definitely be a better fit than the cache.  You can choose to download just a particular named artifact -- you don't have to download everything.

2bndy5:
>  You can choose to download just a particular named artifact -- you don't have to download everything.

Not if we're creating a single artifact concurrently. I tried this using 1 artifact with a single name, but the CI will fail if an artifact is currently being written to by another concurrent job.

2bndy5:
Similar to `actions.cache`, downloading artifacts by name doesn't support glob patterns.

jbms:
Yeah I imagined you would only download the artifacts from a separate job that runs after all the other jobs complete, and combine the coverage data.

2bndy5:
Yeah. Doing that will require separate artifacts per concurrent job, and the final reporting job will have to download all artifacts regardless of name (just to get the coverage data artifacts).

2bndy5:
I'll submit the change and see if it desirable...

2bndy5:
ok, I got the coverage report to combine from artifacts of concurrently run tests. I also applied concurrency to the check_conventions job.

Now we should discuss which jobs to exclude from the test matrix. I was thinking that
- macos should run tests only on latest py and latest sphinx versions because it is the slowest runner there is.
- windows should run tests either
   1. on all sphinx versions but only on latest py version
   2. on all py versions but only only on latest sphinx version
- ubuntu will still run tests on all supported sphinx and py versions because it is the champion of runners' speed.

The biggest detriment in speed for non-ubuntu runners is the "install graphviz" step, so...

2bndy5:
I'm just remembering that py3.11 release notes boasted ["improved performance on Windows"](https://github.com/python/cpython/issues/89279), so I'll just use latest py version with all sphinx versions on windows runner.

2bndy5:
While I'm still at it... Did you want to limit the sphinx builders used for the docs in macOS or Windows?

jbms:
To be clear, I wasn't suggesting skipping any variants, just having multiple windows and macos jobs in the matrix and distributing the variants among them.

2bndy5:
oh! I can revert the limitations.

jbms:
The current configuration is maybe more parallelized than necessary, but it is okay.

What I had in mind was that for ubuntu, there would just be a single test job.  For macos and windows, we might have one test job (i.e. matrix `include` entry) that sequentially tests on python 3.8 and 3.9 (all sphinx versions), and another test job (i.e. matrix `include` entry) that sequentially tests on python 3.10 and 3.11.

2bndy5:
Oh. Well you're welcome to tweak it further. I'm kinda an all-or-nothing guy. The logs are somewhat clearer when it is all parallelized.

jbms (approved):

2bndy5:
I just learned about [`just`][just]. It seems like a feasible alternative to `nox` (and makefiles) because a python venv is not mandated (and the [shell used is configurable](https://just.systems/man/en/chapter_27.html#python-3)). The caching of a python venv in CI would be slightly less implicit (venv names would have to be defined in the [justfile](https://just.systems/man/en/chapter_20.html)).

Using `@nox.session(python=False)` installs things to the system python env. Lately I've been avoiding the use of `nox` locally to prevent `nox -s docs` from polluting my system python env. [`just`][just] seems like a viable approach, but it isn't installed via pip (see [using a pkg-mgr](https://just.systems/man/en/chapter_4.html) or [using a gh action](https://just.systems/man/en/chapter_6.html)).

[just]: https://just.systems/man/en/

PR #231, opened 2023-03-02
add mermaid.js and mathjax to npm deps

### LINKED ISSUE #150: caching (locally) npm pkgs' dist at docs build stage
> ~This may be off topic, but I figured I'd mention it now that this aims to solve #111.~ I think the fetching of mermaid.js should be added to `external_resource_cache`. Currently, the `sphinx_immaterial.mermaid_diagrams` ext is on by default because the theme src seems to be fetching the mermaid.js resource regardless.

Agreed, and also for MathJax.  But that could be deferred for a separate PR.

_Originally posted by @jbms in https://github.com/jbms/sphinx-immaterial/issues/144#issuecomment-1218796196_

2bndy5:
- mermaid dist is copied to built docs' static files.
- mathjax dist is copied to built docs' static files if the sphinx.ext.mathjax is enabled.
  mathjax script tag attributes intentionally exclude async.

resolves #150

### DIFF
--- a/sphinx_immaterial/external_resource_cache.py
+++ b/sphinx_immaterial/external_resource_cache.py
@@ -2,7 +2,7 @@ import hashlib
 import json
 import os
 import tempfile
-from typing import Dict, Optional
+from typing import Dict, Optional, Any
 
 import appdirs
 import requests
@@ -81,6 +81,20 @@ def get_cache_dir(app: sphinx.application.Sphinx) -> str:
     return getattr(app.config, _RESOURCE_CONFIG_KEY)
 
 
+def _config_inited(
+    app: sphinx.application.Sphinx, config: sphinx.config.Config
+) -> None:
+    """Inject path to local cache of mathjax dist.
+    Sphinx machinery will do the rest of the work for us."""
+    if "mathjax_path" in config:
+        config["mathjax_path"] = "tex-mml-chtml.js"
+
+        assert "mathjax_options" in config
+        mathjax_options: Dict[str, Any] = config["mathjax_options"]
+        # async attr is not applicable as the script is self-hosted
+        mathjax_options.update({"id": "MathJax-script", "async": None})
+
+
 def setup(app: sphinx.application.Sphinx):
     app.add_config_value(
         _RESOURCE_CONFIG_KEY,
@@ -88,7 +102,7 @@ def setup(app: sphinx.application.Sphinx):
         rebuild="env",
         types=(str,),
     )
-
+    app.connect("config-inited", _config_inited)
     return {
         "parallel_read_safe": True,
         "parallel_write_safe": True,
--- a/sphinx_immaterial/external_resource_cache.py
+++ b/sphinx_immaterial/external_resource_cache.py
@@ -2,7 +2,7 @@ import hashlib
 import json
 import os
 import tempfile
-from typing import Dict, Optional, Any
+from typing import Dict, Optional
 
 import appdirs
 import requests
@@ -81,20 +81,6 @@ def get_cache_dir(app: sphinx.application.Sphinx) -> str:
     return getattr(app.config, _RESOURCE_CONFIG_KEY)
 
 
-def _config_inited(
-    app: sphinx.application.Sphinx, config: sphinx.config.Config
-) -> None:
-    """Inject path to local cache of mathjax dist.
-    Sphinx machinery will do the rest of the work for us."""
-    if "mathjax_path" in config:
-        config["mathjax_path"] = "tex-mml-chtml.js"
-
-        assert "mathjax_options" in config
-        mathjax_options: Dict[str, Any] = config["mathjax_options"]
-        # async attr is not applicable as the script is self-hosted
-        mathjax_options.update({"id": "MathJax-script", "async": None})
-
-
 def setup(app: sphinx.application.Sphinx):
     app.add_config_value(
         _RESOURCE_CONFIG_KEY,
@@ -102,7 +88,6 @@ def setup(app: sphinx.application.Sphinx):
         rebuild="env",
         types=(str,),
     )
-    app.connect("config-inited", _config_inited)
     return {
         "parallel_read_safe": True,
         "parallel_write_safe": True,
--- a/sphinx_immaterial/external_resource_cache.py
+++ b/sphinx_immaterial/external_resource_cache.py
@@ -88,6 +88,7 @@ def setup(app: sphinx.application.Sphinx):
         rebuild="env",
         types=(str,),
     )
+
     return {
         "parallel_read_safe": True,
         "parallel_write_safe": True,
--- a/sphinx_immaterial/local_mathjax.py
+++ b/sphinx_immaterial/local_mathjax.py
@@ -28,7 +28,7 @@ def copy_mathjax_dist(app: Sphinx, env: BuildEnvironment) -> None:
     if app.registry.html_assets_policy == "always" or domain.has_equations():
         # copy mathjax fonts only if equations exists
         shutil.copytree(
-            str(Path(__file__).parent / "static" / "mathjax"),
+            str(Path(__file__).parent / "bundles" / "mathjax"),
             str(Path(app.outdir, "_static", "mathjax")),
             dirs_exist_ok=True,  # for consecutive builds
         )
--- a/sphinx_immaterial/local_mathjax.py
+++ b/sphinx_immaterial/local_mathjax.py
@@ -1,0 +1,64 @@
+from pathlib import Path
+import shutil
+from typing import Dict, Any, cast
+from sphinx.application import Sphinx
+from sphinx.config import Config
+from sphinx.errors import ExtensionError
+from sphinx.ext.mathjax import MATHJAX_URL, MathDomain
+
+
+def copy_mathjax_dist(
+    app: Sphinx,
+    pagename: str,
+    templatename: str,
+    context: Dict[str, Any],
+    event_arg: Any,
+) -> None:
+    if (
+        app.builder.format != "html"
+        or app.builder.math_renderer_name != "mathjax"  # type: ignore[attr-defined]
+    ):
+        return
+    if not app.config.mathjax_path:
+        raise ExtensionError(
+            "mathjax_path config value must be set for the mathjax extension to work"
+        )
+    if (
+        "mathjax_path" in app.config
+        and app.config["mathjax_path"] != "mathjax/tex-mml-chtml.js"
+    ):
+        return  # not using theme's cache of mathjax dist
+
+    domain = cast(MathDomain, app.env.get_domain("math"))
+    if app.registry.html_assets_policy == "always" or domain.has_equations(pagename):
+        # copy mathjax fonts only if equations exists (and not already done)
+        dst = Path(app.outdir, "_static", "mathjax")
+        if Path(app.outdir, "_static", dst).exists():
+            return
+        shutil.copytree(
+            str(Path(__file__).parent / "static" / "mathjax"),
+            str(dst),
+        )
+
+
+def _config_inited(app: Sphinx, config: Config) -> None:
+    """Inject path to local cache of mathjax dist.
+    Sphinx machinery will do the rest of the work for us."""
+    if "mathjax_path" in config:
+        is_default = config["mathjax_path"] == MATHJAX_URL
+        if is_default:
+            config["mathjax_path"] = "mathjax/tex-mml-chtml.js"
+
+            assert "mathjax_options" in config
+            mathjax_options: Dict[str, Any] = config["mathjax_options"]
+            # async attr is not applicable as the script is self-hosted
+            mathjax_options.update({"id": "MathJax-script", "async": None})
+
+
+def setup(app: Sphinx):
+    app.connect("config-inited", _config_inited)
+    app.connect("html-page-context", copy_mathjax_dist)
+    return {
+        "parallel_read_safe": True,
+        "parallel_write_safe": True,
+    }
--- a/sphinx_immaterial/local_mathjax.py
+++ b/sphinx_immaterial/local_mathjax.py
@@ -5,15 +5,10 @@ from sphinx.application import Sphinx
 from sphinx.config import Config
 from sphinx.errors import ExtensionError
 from sphinx.ext.mathjax import MATHJAX_URL, MathDomain
+from sphinx.environment import BuildEnvironment
 
 
-def copy_mathjax_dist(
-    app: Sphinx,
-    pagename: str,
-    templatename: str,
-    context: Dict[str, Any],
-    event_arg: Any,
-) -> None:
+def copy_mathjax_dist(app: Sphinx, env: BuildEnvironment) -> None:
     if (
         app.builder.format != "html"
         or app.builder.math_renderer_name != "mathjax"  # type: ignore[attr-defined]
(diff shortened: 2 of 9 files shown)


jbms:
I think rather than download mermaid.js when the docs are built, we could just add it as a dependency to `packages.json` and then include it in the built package / wheel with the other javascript bundles (but not bundle it).  We can do the same thing for mathjax.

2bndy5:
I had considered that as well, but I wasn't sure if changing the package.json was a welcome change. I'll rework this and get back to you...

I'm still not convinced that the mathjax lib needs to be handled by this theme. Sphinx' [`mathjax_path`](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-mathjax_path) can be used by users to do the same thing that was suggested. Maybe I'm missing something.

jbms:
It is true that mathjax isn't really related to this theme.  Perhaps it would be better if Sphinx itself included a version of mathjax so that there is no dependency on external resources by default.  But I'm not sure how likely that is to go through...

Particularly since we may soon provide a way to use mathml instead, I think it is reasonable to just ignore mathjax for now...

But in general providing a way to avoid any dependence on external resources is a useful feature, and one that can be a bit annoying for users to implement themselves.

2bndy5:
>  Perhaps it would be better if Sphinx itself included a version of mathjax so that there is no dependency on external resources by default

Oh, you're focused on the _default_ behavior. 🤦🏼 That would be so easy to just change the default value of `mathjax_path` if the config has the attr defined... I'm not sure where we'd document this adjustment. Maybe it can be mentioned with the solution for #233 .

src/assets/javascripts/components/content/code/mermaid/index.ts
```diff
@@ -69,7 +69,7 @@ let sequence = 0
  */
 function fetchScripts(): Observable<void> {
   return typeof mermaid === "undefined" || mermaid instanceof Element
-    ? watchScript("https://unpkg.com/mermaid@9.1.7/dist/mermaid.min.js")
+    ? watchScript("_static/mermaid.min.js") // sphinx-immaterial uses cached dist
```
jbms, line 72: This path needs to take into account that the document may be a in sub-dir.

2bndy5: I was under the impression that the path specified here is relative to the base site path. The python code (in \_\_init\_\_.py) doesn't override user-defined mathjax path, so I'm not sure how to get this JS to coincide with the Py side without overriding the user-defined path (which will ignore any custom version of mathjax dist that users try to employ).

sphinx_immaterial/external_resource_cache.py
```diff
@@ -81,14 +81,28 @@ def get_cache_dir(app: sphinx.application.Sphinx) -> str:
     return getattr(app.config, _RESOURCE_CONFIG_KEY)
 
 
+def _config_inited(
+    app: sphinx.application.Sphinx, config: sphinx.config.Config
+) -> None:
+    """Inject path to local cache of mathjax dist.
+    Sphinx machinery will do the rest of the work for us."""
+    if "mathjax_path" in config:
+        config["mathjax_path"] = "tex-mml-chtml.js"
```
2bndy5, line 90: I have to doube check this. I think this will override the user-defined path, but I'm not sure if it is a guarantee that mathjax is being used.

src/assets/javascripts/components/content/code/mermaid/index.ts
jbms: I believe the mathjax path is indeed handled by sphinx automatically, but this code is for mermaid and is using the mkdocs-material `watchScript` function which just seems to copy the argument directly as the `<script src="...">` attribute.

2bndy5: > but this code is for mermaid 

oops! 😆 

> using the mkdocs-material `watchScript` function which just seems to copy the argument directly as the `<script src="...">` attribute.

Ah yes. Currently, there's no way to specify where the mermaid dist is being copied from (by Sphinx at docs' build time). The build script should always copy the mermaid dist to the same folder for theme dist. So, I think the subfolder concern is null, yes?

sphinx_immaterial/external_resource_cache.py
2bndy5: Looks like this path will always be present if the mathjax ext is added to the conf.py `extensions`. Resolving this.

src/assets/javascripts/components/content/code/mermaid/index.ts
jbms: The issue is if we have foo/bar.rst

2bndy5: Oh, I see. Good catch! I'm not sure how I would go about specifying the full path to the site base. I'll have to investigate further...

2bndy5: Any suggestion as to how to do this for both local `file::///` and web-hosted `https://`? I see `window.location.origin` only works for the later.

jbms: You can use the approach from sphinx_search.ts: https://github.com/jbms/sphinx-immaterial/blob/d91eacc94250c7fafb8b930ac744d3bbcffe50da/src/assets/javascripts/sphinx_search.ts#L30

2bndy5: TBH, I don't fully understand how that works better than my previous attempts involving `location.href`, (must be something to do with the `URL` object), but I pushed the changes after verifying that they work in rst-cheatsheet/rst-cheatsheet.rst.

jbms: Basically we have:

```
config.base = `${new URL(base_url, new URL(location.href))}`
```

where `base_url` is defined in Python as:

```
    num_slashes = builder.get_target_uri(pagename).count("/")
    if num_slashes == 0:
        base_url = "."
    else:
        base_url = "/".join(".." for _ in range(num_slashes))
```

The key is that we need an external source to tell us know many levels up we need to go.

2bndy5: Ah, the `base_url` bit is what I was missing. Thanks for the extra details!

src/assets/javascripts/components/content/code/mermaid/index.ts
```diff
@@ -69,7 +75,7 @@ let sequence = 0
  */
 function fetchScripts(): Observable<void> {
   return typeof mermaid === "undefined" || mermaid instanceof Element
-    ? watchScript("https://unpkg.com/mermaid@9.1.7/dist/mermaid.min.js")
+    ? watchScript(`${config.base}/_static/mermaid.min.js`) // sphinx-immaterial uses cached dist
```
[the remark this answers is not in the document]
2bndy5: Damn I forgot to check that.

sphinx_immaterial/external_resource_cache.py
```diff
@@ -81,14 +81,28 @@ def get_cache_dir(app: sphinx.application.Sphinx) -> str:
     return getattr(app.config, _RESOURCE_CONFIG_KEY)
 
 
+def _config_inited(
```
jbms, line 84: This would probably be better in a separate file, e.g. `local_mathjax.py`, since it doesn't really relate to the rest of this file (and note that the google fonts are now handled by a separate file as well).

We also probably want a way to disable this behavior, so that users can still set their own mathjax path if desired.

Finally, it would be nice to only copy mathjax if actually used.  See this logic here:
https://github.com/sphinx-doc/sphinx/blob/4eb706eb21bc3aaa849f036593369ee1c503743d/sphinx/ext/mathjax.py#L82

Similarly for mermaid it would be nice to also only copy it if actually needed.

2bndy5: > Similarly for mermaid it would be nice to also only copy it if actually needed.

I could make the mermaid extension optional, but you probably want a more context-specific solution (as suggested for mathjax.

jbms: Yes that's what I was thinking.

2bndy5: Well, I abstracted the config-inited code into its own module and removed the lines from `sphinx_immaterial/__init__.py` that copy the mathjax dist into the build dir. This seems to still work well: Mathjax script is still added to the HTML accordingly via `sphinx.ext.mathjax.install_mathjax()`.

I also added a condition to only modify the `mathjax_path` conf option if the value is set to the mathjax ext default path value. This tested well when I copied the mathjax dist into docs/_static/dist/mathjax.js & modified the conf.py accordingly.

---
However, I'm noticing new font issues spawn from the self-hosted mathjax script. From the web console:
```
downloadable font: download failed (font-family: "MJXZERO" style:normal weight:400 stretch:100 src index:0): status=2152857618 source: file:///path/to/sphinx-immaterial/docs/_build/html/_static/output/chtml/fonts/woff-v2/MathJax_Zero.woff
downloadable font: download failed (font-family: "MJXTEX" style:normal weight:400 stretch:100 src index:0): status=2152857618 source: file:///path/to/sphinx-immaterial/docs/_build/html/_static/output/chtml/fonts/woff-v2/MathJax_Main-Regular.woff
downloadable font: download failed (font-family: "MJXTEX-I" style:normal weight:400 stretch:100 src index:0): status=2152857618 source: file:///path/to/sphinx-immaterial/docs/_build/html/_static/output/chtml/fonts/woff-v2/MathJax_Math-Italic.woff
```

I think it means I need to copy these fonts from the mathjax npm pkg in node_modules; in which case it would probably be better to put the mathjax dist files in their own folder during npm build. Since the `sphinx.ext.mathjax` doesn't seem to handle the fonts, this seems to complicate things even further.

2bndy5: Ok, I've got a working solution for mathjax (including the mathjax fonts now). I'll push it shortly.

---
The JS bundle already conditionally mounts the mermaid script if the page has a `<div class="mermaid">` in it. So, I should only need to devise a way to detect if any mermaid directives are invoked, then copy the mermaid dist when needed.

My initial guess is to use the builder env to persist a theme-specific flag for this, but maybe you have a better approach in mind?

jbms: Yes, adding it in the env (and remember to handle merging in order to support parallel builds) is exactly what I had in mind.

jbms:
Regarding mathjax, now that all major browsers support MathML, it may make sense to explore whether we can just recommend using mathml instead of mathjax, since if it works, it would require no additional libraries and would render without a delay.

This comment describes how it might simply require modifying the docutils.conf file in the doc root directory:

https://github.com/sphinx-doc/sphinx/issues/6092#issuecomment-1501088319

Not sure if anyone has actually tested that, though.

2bndy5:
I was going to approach the #233 separately. The newer MathML feature does look much more appealing, but this work is more like retrofitting when it comes to mathjax use. The sphinx dev cycle isn't too slow (for the most part), so hopefully there will be more development/testing with MathML in Sphinx by the time we get around to #233 .

sphinx_immaterial/local_mathjax.py
```diff
@@ -0,0 +1,64 @@
...
+        "mathjax_path" in app.config
+        and app.config["mathjax_path"] != "mathjax/tex-mml-chtml.js"
+    ):
+        return  # not using theme's cache of mathjax dist
+
+    domain = cast(MathDomain, app.env.get_domain("math"))
+    if app.registry.html_assets_policy == "always" or domain.has_equations(pagename):
+        # copy mathjax fonts only if equations exists (and not already done)
+        dst = Path(app.outdir, "_static", "mathjax")
+        if Path(app.outdir, "_static", dst).exists():
```
jbms, line 36: I think rather than checking if the path exists in the output dir, it would be better to just always copy the files if mathjax is being used, but not more than once per build.

We could accomplish that by adding a hook to the `env-updated` event.

2bndy5: I only used the `html-page-context` event because `domain.has_equations(pagename)` seemed vital. Most of this function's beginnings were straight copied from the `mathjax.install_mathjax()` function. Though I think the `env-updated` event can be used similarly, I just have to figure out how it could fit with the `domain.has_equations()` call.

2bndy5: Oh looks like the `docname` param to `domain.has_equations` is optional... commiting shortly.

2bndy5: done

sphinx_immaterial/local_mathjax.py
```diff
@@ -0,0 +1,57 @@
...
+        "mathjax_path" in app.config
+        and app.config["mathjax_path"] != "mathjax/tex-mml-chtml.js"
+    ):
+        return  # not using theme's cache of mathjax dist
+
+    domain = cast(MathDomain, env.get_domain("math"))
+    if app.registry.html_assets_policy == "always" or domain.has_equations():
+        # copy mathjax fonts only if equations exists
+        shutil.copytree(
+            str(Path(__file__).parent / "static" / "mathjax"),
```
jbms, line 31: Since the `static` directory is by default always copied, it would be simpler to put mathjax and mermaid in a different directory and then there is no need to specifically exclude them.

2bndy5: like "bundles" or were you thinking of a new folder?

sphinx_immaterial/mermaid_diagrams.py
```diff
@@ -34,6 +40,7 @@ def run(self) -> List[nodes.Node]:
             self.add_name(diagram_div)
         diagram_div += diagram
         self.set_source_info(diagram_div)
+        setattr(self.env, _COPY_MERMAID_DIST_ENV_KEY, True)
```
jbms, line 43: To support parallel building, a handler for `env-merge-info` is needed so that this information gets propagated back to the main process.  Otherwise, it will only be set in the workers that happen to process documents with mermaid directives.

https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-env-merge-info

sphinx_immaterial/local_mathjax.py
jbms: I think bundles is fine.

jbms:
Can you also ensure that the licenses for these two packages are added to the generated license file (see `tools/buil.d/transform/index.ts`),; the existing logic only includes packages that are bundled by esbuild.

2bndy5:
> Can you also ensure that the licenses for these two packages are added to the generated license file 

The mermaid dist includes a copy of the License alongside the mermaid.min.js\* files. Did you want to keep that in the bundles folder as well? Or is it sufficient to have it only in the generated LICENSE file?

sphinx_immaterial/mermaid_diagrams.py
2bndy5: I think I did this as requested. I basically took a hint from the inline_icons.py module:
https://github.com/jbms/sphinx-immaterial/blob/6713712c7211b7c684b7be271aebd8559941e1af/sphinx_immaterial/inline_icons.py#L80-L90

jbms:
> > Can you also ensure that the licenses for these two packages are added to the generated license file
> 
> The mermaid dist includes a copy of the License alongside the mermaid.min.js* files. Did you want to keep that in the bundles folder as well? Or is it sufficient to have it only in the generated LICENSE file?

Keeping a copy of the license there as well makes sense.

sphinx_immaterial/mermaid_diagrams.py
```diff
@@ -54,13 +61,38 @@ def depart_mermaid_node_latex(self, node: mermaid_node):
...
+        dst = Path(app.outdir, "_static", "mermaid")
+        if dst.exists():
+            return
+        shutil.copytree(str(Path(__file__).parent / "bundles" / "mermaid"), dst)
+
+
+def _merge_env_key(
+    app: Sphinx, env: BuildEnvironment, docnames: List[str], other: BuildEnvironment
+) -> None:
+    setattr(env, _COPY_MERMAID_DIST_ENV_KEY, getattr(other, _COPY_MERMAID_DIST_ENV_KEY))
```
jbms, line 83: This should do a "logical or" rather than just unconditionally copying the value.

2bndy5: done

jbms (approved):
Thanks!

Issue #222, opened 2023-02-27
Using without Google Fonts


hille721:
Really like this project and would love to use it, as with it's dark theme it is so much better than sphinx-material. But the fact that it will download the Google Fonts from remote server is currently blocking me to use this project. In my CI I don't have internet access, thus downloading is not a option. 
Yes, I could place the fonts in a directory and pointing the `sphinx_immaterial_external_resource_cache_dir`, but in CI also this is not so easy. I could put them directly in my GIT repo, but then I would double the size. 

Is there any possibility to disable the usage of Google fonts?

2bndy5:
This is the second time this theme's external resources extension has caused a problem concerning access to google's fonts.

I think this is worth looking into. The upstream mkdocs theme has support for manually specifying the font in a situation where downloading the fonts are not an option (see [here](https://squidfunk.github.io/mkdocs-material/setup/changing-the-fonts/#autoloading)), but I haven't tested this since the external resources extension was added to this theme.

hille721:
> I think this is worth looking into. The upstream mkdocs theme has support for manually specifying the font in a situation where downloading the fonts is not an option (see [here](https://squidfunk.github.io/mkdocs-material/setup/changing-the-fonts/#autoloading)), but I haven't tested this since the external resources extension was added to this theme.

Oh, just realized that this seems to work

``` python
html_theme_options = {
    ...
    "font": False,
    ...
}

2bndy5:
Yes, I see that too, but I'm looking into ways to compensate the graphviz custom extension... 
This patch seems to have a semi-acceptable result, but the font color needs to be fixed.
```diff
--- a/sphinx_immaterial/google_fonts.py
+++ b/sphinx_immaterial/google_fonts.py
@@ -181,7 +181,9 @@ def add_google_fonts(app: sphinx.application.Sphinx, fonts: List[str]):


 def get_ttf_font_paths(app: sphinx.application.Sphinx) -> Dict[Tuple[str, str], str]:
-    return getattr(app, _TTF_FONT_PATHS_KEY)
+    if hasattr(app, _TTF_FONT_PATHS_KEY):
+        return getattr(app, _TTF_FONT_PATHS_KEY)
+    return None


 def _builder_inited(app: sphinx.application.Sphinx):
diff --git a/sphinx_immaterial/graphviz.py b/sphinx_immaterial/graphviz.py
index d242e63..eb7faf9 100644
--- a/sphinx_immaterial/graphviz.py
[...]

PR #232, opened 2023-03-02
Added embedding default fonts in the theme wheel

glatosinski:
This PR is related to https://github.com/jbms/sphinx-immaterial/issues/222, and proposes an implementation for one of the discussed variants: `Add the default Google fonts to this theme's package`.

It is a mechanism that we currently use to prevent downloading fonts every time we start a fresh build of documentation

Changes made in the PR:
* Added `sphinx_immaterial.resources` module that contains pre-downloaded fonts delivered with the wheel.
* Updated `add_google_fonts` method so it first checks the `sphinx_immaterial.resources` module for bundled fonts, then it checks local cache, and in the end it downloads fonts from remote server
* Added downloading default fonts during stage/wheel building to `sphinx_immaterial.resources` module to embed them in release wheel
* Added stub `sphinx_immaterial.resources` module


requirements.txt
```diff
@@ -4,3 +4,4 @@ pydantic
 typing-extensions
 appdirs
 requests
+importlib-resources
```
2bndy5, line 7: It looks like there's a [`importlib.resources` module](https://docs.python.org/3/library/importlib.resources.html) that has been available as part of the std libs since python v3.7. Since python 3.6 has reached end-of-life and this theme requires v3.8+, does this mean we don't need a new external dependency? Is there some advantage to using the thrid-party [`importlib_resources` lib](https://importlib-resources.readthedocs.io/en/latest/using.html)?

setup.py
```diff
@@ -30,6 +30,12 @@
 import setuptools.command.develop
 import setuptools.command.install
 import setuptools.command.sdist
+from sphinx_immaterial.google_fonts import install_google_fonts
+
+from importlib_resources import files
+from sphinx_immaterial import resources
+
+from sphinx_immaterial import DEFAULT_THEME_OPTIONS
```
2bndy5, line 38: This is causing a "chicken before the egg" problem because you are trying to import from the very pkg that this installs. The CI seems fine because the pkg deps are installed before the pkg is built. However, when I run this in a fresh env, I get `ModuleNotFoundError`.

<details><summary>full traceback</summary>

```
Traceback (most recent call last):
  File "path\to\repo\.venv-temp\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
    main()
  File "path\to\repo\.venv-temp\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "path\to\repo\.venv-temp\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 118, in get_requires_for_build_wheel
    return hook(config_settings)
           ^^^^^^^^^^^^^^^^^^^^^
  File "path\to\temp\pip-build-env-415k6g8g\overlay\Lib\site-packages\setuptools\build_meta.py", line 338, in get_requires_for_build_wheel
    return self._get_build_requires(config_settings, requirements=['wheel'])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "path\to\temp\pip-build-env-415k6g8g\overlay\Lib\site-packages\setuptools\build_meta.py", line 320, in _get_build_requires
    self.run_setup()
  File "path\to\temp\pip-build-env-415k6g8g\overlay\Lib\site-packages\setuptools\build_meta.py", line 485, in run_setup
    self).run_setup(setup_script=setup_script)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "path\to\temp\pip-build-env-415k6g8g\overlay\Lib\site-packages\setuptools\build_meta.py", line 335, in run_setup
    exec(code, locals())
  File "<string>", line 33, in <module>
  File "path\to\repo\sphinx-immaterial\sphinx_immaterial\__init__.py", line 6, in <module>
    import docutils.nodes
ModuleNotFoundError: No module named 'docutils'
```

</details>

setup.py
```diff
@@ -146,6 +152,11 @@ def run(self):
         target = {"min": "build", "dev": "build:dev"}
 
         try:
+            install_google_fonts(
+                files(resources),
+                files(resources),
+                DEFAULT_THEME_OPTIONS["font"].values(),
+            )
```
2bndy5, line 159: I don't think this will really solve the problem reported in #222. If a CI installs this theme, then the install process will error out if the CI runner doesn't have access to Google's font API (as similarly reported in #222).

I was under the impression that we'd have to include copies of the default fonts with the theme src to properly address #222. I do appreciate the attempt to update the default fonts at pkg install time, but it isn't really feasible (with respect to #222).

#222 would be better resolved by updating the default fonts' cache when `npm run build` is performed.

requirements.txt
2bndy5: After some more thought, I don't think the cached default fonts needs to be an importable resource. We could just use 

```py
# from google_fonts.py or external_resource_cache.py
pathlib.PurePath(__file__).parent / f"{default_font_location}" / f"{default_font_name}"
```

The `importlib.resources` module seems designed to let external API easily find a non-python asset from our pkg.

setup.py
jbms: To avoid this we could set a special environment variable that causes `sphinx_immaterial/__init__.py` not to import other stuff.

setup.py
jbms: When installing from the git repository directly, we have to download the npm packages (which requires internet access).  However, the wheel package that we publish to PyPI includes the pre-built javascript and css bundles, and icons, and I think the idea here is that it could also include the Google fonts.  So it would in fact solve the problem reported in #222 as long as the user does not wish to change the fonts from their default values.

2bndy5: Yeah that was my thinking exactly.

I'm not sure if installing from git is preferred when the users' env has limited access to other domains.

requirements.txt
glatosinski: `importlib-resources` is added due to the fact that Python versions below 3.9 do not have `files` in `importlib.resources`. But sure, I can do a workaround that will not use `importlib-resources` dependency

setup.py
2bndy5: Installing from git/zip is also common for envs that don't have python's pip tool. For example, the python executable shipped with Ubuntu usually doesn't have `pip` or `venv` and the admin group (needed to use `apt`) isn't viable for the active user(s). Sure, there are alternative install methods (docker containers), but it usually falls on pkg maintainers when all else fails.

umarcor: @2bndy5 I might be wrong but my understanding is that users who don't have pip or venv and who are not in the admin group will need to install npm and the JS deps in order to install this package. Therefore, they have more things to worry about than whether their env/runner has internet access to Google. Setting that up falls into "packaging", which can be done by the distro maintainers or by the end user.

Hence, that's a different use case than the one covered in this PR. I believe that this PR should be compatible with that one, but it should not be constrained by envs not having access to Google at build time.

2bndy5: > I believe that this PR should be compatible with that one, but it should not be constrained by envs not having access to Google at build time.

That is the exact problem that #222 poses. The OP of this PR aims to implement a solution discussed in #222. I feel like we're getting off track again (you and I seem to have a "chemistry" 😜 ).

2bndy5: BTW, if installing from a zip, the archive should be in the form of the sdist deployed to pypi which shouldn't need npm or the repo's JS deps because `npm run build` bundles all of that stuff into the sphinx_immaterial directory _before_ compressing the archive. Then the setup.py just indicates (via [`package_data`](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data)) what is contained in the sphinx_immaterial pkg.

umarcor: Let's see if I'm getting it right:

- #222 is the discussion of a problem and 5 different solutions are proposed, each covering slighly different use cases.
- #232, this PR, implements one of the 5 solutions to support one of the use cases.
- We want to make sure that implementing this use case does not break all other 4 use cases. We all agree on that.

What's the need to portrait this as a competitive solution instead of a collaborative one? I'm referring to *"https://github.com/jbms/sphinx-immaterial/issues/222 would be **better** resolved by updating the default fonts' cache when npm run build is performed"* (bolds added by me). Precisely, as discussed in #222, it is only solved by covering the five use cases, or by making it explicit which of them are not to be supported (and why). Implementing a single one and ignoring others does not resolve #222, no matter which one it is. This PR is "related to #222", not closing or fixing it.

> I feel like we're getting off track again (you and I seem to have a "chemistry" 😜 ).

It's ok, it's just a matter of experience in the open-source galaxy 😃. We have very different points of view about the project. You try to bring attention towards other related maintenance/development issues which you feel important, and I'm trying to ponderate attention from an external position. In several other projects, I do play the role you are playing here.

umarcor: > BTW, if installing from a zip, the archive should be in the form of the sdist deployed to pypi which shouldn't need npm or the repo's JS deps because npm run build bundles all of that stuff into the sphinx_immaterial directory before compressing the archive. Then the setup.py just indicates (via [package_data](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data)) what is contained in the sphinx_immaterial pkg.

This is interesting! So, the proposal is to install an sdist instead of the zip of this repo? Where would that sdist zip be hosted?

2bndy5: Oh, they aren't 5 use cases; they're [5 possible solutions](https://github.com/jbms/sphinx-immaterial/issues/222#issuecomment-1446913574) to the reported problem. We agreed to pursue solution 5 because it is the most unobtrusive approach. The other solutions (except solution 1) would still be supported, but they require user-implementation. Solution 1 has been unsupported ever since the external resources ext was added. Consequently, not supporting solution 1 has instigated #222...

I'm not sure how other spoken languages work out, but English is often too imprecise for us analytical thinkers 🤓 

---
FYI, we had discussed something like solution 5 when the external resources ext was added (#144), but because that was lumped into the theme's custom graphviz addition, we decided it deserved its own PR.

umarcor: Agree. Code speaks better. 

Is the npm based approach something you expect to do in the following weeks? We (I) can hold this discussion until we can better see and understand what's the difference.

2bndy5: I have an idea about using the fontsource npm dists for [Robot](https://www.npmjs.com/package/@fontsource/roboto) and [Robot Mono](https://www.npmjs.com/package/@fontsource/roboto-mono). But I haven't tackled how to integrate it into google_fonts.py yet.

2bndy5: > With "in the form" you mean "something that looks like it" or "it will be uploaded to pypi and installable through pip"?

I mean the archive would be the actual sdist that is uploaded to pypi but agnostic to the compression used (tar.gz vs zip).

2bndy5: I completely missed that comment, so I'm answering it in parts...

> Where would that sdist zip be hosted?

Currently, we don't publish releases on github, but we could upload the archive as a release asset if we do. Maybe that is something that should be discussed as part of #223. However, installing from git does require npm and stuff.

umarcor: @2bndy5 any update to this feature? Is there an alternative implementation?

2bndy5: If you or @glatosinski could satisfy the review concerns here, then I'd be happy to move this PR forward and implement the switch to fontsource CDN in a separate PR.

I'm currently enthralled with exploring my idea for #261.

glatosinski: Ok, I'll look into it

PR #244, opened 2023-04-26
support version directives

### LINKED ISSUE #243: BUG/ENH: Common Sphinx directives have no style applied
Sphinx admonitions such as 

```rst
.. deprecated:: 3.0
   This is deprecated since 3.0
```

have no style applied. 

If you search for deprecated in the [specimen page](https://jbms.github.io/sphinx-immaterial/demo_api.html#test-py-module) you will see unstyled markup of

```rst
.. versionadded:: 1.0
   This was added in 1.0
.. versionchanged:: 2.0
   This was changed in 2.0
.. deprecated:: 3.0
   This is deprecated since 3.0
```

2bndy5:
resolves #243

- Overrides version directives `versionadded`, `versionchanged`, and `deprecated` with custom admonition directives.
- Adds new conf.py option `sphinx_immaterial_override_version_directives` to disable this new behavior.
- Adds CSS to italicize the version directive "label" as is done in the Sphinx basic theme CSS.
- Updates the admonitions.rst doc
- Admonition content is optional for newly overridden version directives. Content is still required for existing admonition overrides (as is the default behavior)
- Overridden version directives require 1 argument and still support 1 optional argument.

### DIFF
--- a/docs/admonitions.rst
+++ b/docs/admonitions.rst
@@ -53,6 +53,34 @@ usage in other sphinx-based themes.
    .. error::
       You have made a grave :dutree:`error`.
 
+Version directives
+******************
+
+The `versionadded`, `versionchanged`, and `deprecated` directives defined by Sphinx are also
+overridden by the sphinx-immaterial theme to render as an admonition.
+
+.. confval:: sphinx_immaterial_override_version_directives
+
+   This theme can be configured to disallow these overrides if the new features cause a conflict
+   with other Sphinx extensions.
+
+   .. code-block:: python
+      :caption: Disallow overriding version directives in conf.py
+
+      sphinx_immaterial_override_version_directives = False
+
+.. rst-example::
+
+    .. versionadded:: 1.0
+        Description in title.
+
+        Some additional context.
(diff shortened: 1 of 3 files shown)


jbms:
Thanks for working on this!

I hate to give you more work, but given that Sphinx's implementation of these version directives is not that simple, and has various conditions as far as how to handle the title, and also registers the entries in the `ChangeSetDomain`, I'm thinking it may be better to leave the version directives alone.

In order to style them, we could either do it purely with CSS, using Sphinx's original HTML output, or could add an override to our HTML translator for the versionmodified node.  That would avoid adding complexity to the admonition directive, avoid duplicating the existing logic of the version directives, and allow the ChangeSet domain to continue working.

2bndy5:
I had thought of doing that at one point in the dev process, but I was too curious to see how the no-content stuff would work out.

I'll start a new branch and begin approaching the issue from the CSS side instead (probably tomorrow). The only drawback that I can think of is that CSS is harder for users to undo since we can't conveniently have a `confval` like we do here, but maybe I can figure that out on the way.

Issue #162, opened 2022-09-13
Add MyST markdown example?


mgielda:
First of all, thanks for this work and the awesome theme - looking forward to adopting it in some projects (worked for me out of the box!).

I was curious if you were aware of MyST markdown - https://myst-parser.readthedocs.io/en/latest/syntax/roles-and-directives.html - which is a nice twist given this theme's mkdocs origins. MyST markdown has the benefit of being practically a drop-in replacement for RST, except being an extension of CommonMark, which means that the syntax feels much more familiar to markdown people. The added benefit is that a JS parser exists (the key bit is here: https://github.com/executablebooks/markdown-it-docutils) so that it integrates much better than RST with the web ecosystem.

Overall I think it might be a nice idea to have a little shout-out or example of using this theme with MyST. I can confirm it works out of the box, just add the relevant extension and use an .md file.

jbms:
Would you care to contribute an example to the documentation?

2bndy5:
We are aware of the MyST extension, and we've had other users report positive experiences about using it in this theme.

My only main concern with adding it to the docs is that people should be aware that it isn't exactly the same MD syntax used in the mkdocs-material theme. So, some tweaking should be expected if users are migrating from mkdocs-material to sphinx-immaterial.

A PR would be appreciated.

2bndy5:
It's never good to assume things. I personally migrated a (small) project from mkdocs-material to sphinx-immaterial, but I rewrote the MD files in RST because that's just my personal preference (and it doesn't require more 3rd party Sphinx extensions).

Off the top of my head, I can think of 3 other instances of user feedback (from different users) that had questions about migrating from mkdocs-material to sphinx-immaterial; 2 of the users were documenting an API, the other user was using MyST (and just wanted to hide the toc).

2bndy5:
Good news for those migrating: `pymdownx` extensions will be switching to a new "directive" oriented block syntax that mimics MyST block syntax. This means that migrating from mkdocs to sphinx (or vice-versa) should be less work with the new/proposed syntax. I'm told there will be a minimum 2 year transition away from the current "block syntax soup", then it will be moved to a legacy mkdocs plugin.

2bndy5:
[...]

PR #247, opened 2023-04-27
Backlogged doc updates

### LINKED ISSUE #190: Version aliases usage
Are the version aliases functional in Sphinx Immaterial? I think some extra code or HTML is required, but since that is not my forte I may be mistaken.

From [Material for Mkdocs' documentation](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=version#versioning) and [Sphinx Immaterial's documentation](https://jbms.github.io/sphinx-immaterial/customization.html#version-information-structure), my understanding is:

- To make multiple folders with different versions, `docs/build/v1/` and `docs/build/v2/` in my example.
- Add this to `conf.py`:
```python
html_theme_options = {
    "version_dropdown": True,
    "version_json": "../versions.json",
}
```
- Add the versions JSON file at the same level as the folders, `docs/build/versions.json` in my example.
```json
[
    {
        "version": "v2",
        "title": "v2",
        "aliases": ["latest"]
    },
    {
        "version": "v1",
        "title": "v1",
        "aliases": []
    }
]
```
- Serve the `docs/build/` folder from a webserver.

[foo-190.zip](https://github.com/jbms/sphinx-immaterial/files/9997569/foo-190.zip)

I can access `http://localhost:8080/v1/` and `http://localhost:8080/v2/`, but `http://localhost:8080/latest/` returns 404. I was expecting it to return the same content as `http://localhost:8080/v2/` or simply redirect to `http://localhost:8080/v2/`. Am I misunderstanding how the feature works?

It seems there may need to be a special `docs/build/index.html` or something like a `docs/build/latest -> docs/build/v1` symlink.

### LINKED ISSUE #234: Better document globaltoc_collapse
Hello ! 

It took me a while to realize that the`globaltoc_collapse` theme option not only collapses everything that is not a parent of the current page in the Table of Contents AND prevents those items from being expanded/collapsed.

It was not clear to me, so it may be a good idea to add a line to the documentation ([here](https://jbms.github.io/sphinx-immaterial/customization.html#themeconf-globaltoc_collapse)) to make it more explicit.

Thank you!

### LINKED ISSUE #242: 0.11.3: pytest is failing
I'm packaging your module as an rpm package so I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix> using 'installer` module
- run pytest with $PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>
- build is performed in env which is *`cut off from access to the public network`* (pytest is executed with `-m "not network"`)

Here is pytest output:
<details>

```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-sphinx-immaterial-0.11.3-2.fc35.x86_64/usr/lib64/python3.8/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-sphinx-immaterial-0.11.3-2.fc35.x86_64/usr/lib/python3.8/site-packages
+ /usr/bin/pytest -ra -m 'not network'
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.8.16, pytest-7.3.1, pluggy-1.0.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/tkloczko/rpmbuild/BUILD/sphinx-immaterial-0.11.3
configfile: pyproject.toml
testpaths: tests
collected 30 items / 1 error

========================================================================================== ERRORS ===========================================================================================
____________________________________________________________________ ERROR collecting tests/issue_134/issue_134_test.py _____________________________________________________________________
ImportError while importing test module '/home/tkloczko/rpmbuild/BUILD/sphinx-immaterial-0.11.3/tests/issue_134/issue_134_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.8/site-packages/_pytest/python.py:622: in _importtestmodule
    mod = import_path(self.path, mode=importmode, root=self.config.rootpath)
/usr/lib/python3.8/site-packages/_pytest/pathlib.py:564: in import_path
    importlib.import_module(module_name)
/usr/lib64/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1014: in _gcd_import
    ???
<frozen importlib._bootstrap>:991: in _find_and_load
    ???
<frozen importlib._bootstrap>:975: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:671: in _load_unlocked
    ???
/usr/lib/python3.8/site-packages/_pytest/assertion/rewrite.py:172: in exec_module
    exec(co, module.__dict__)
tests/issue_134/issue_134_test.py:6: in <module>
    from sphinx_immaterial_pybind11_issue_134 import Example  # type: ignore
E   ModuleNotFoundError: No module named 'sphinx_immaterial_pybind11_issue_134'
================================================================================== short test summary info ==================================================================================
ERROR tests/issue_134/issue_134_test.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
===================================================================================== 1 error in 0.35s ======================================================================================
```
</details>

Here is list of installed modules in build env
<details>

```console
Package                       Version
----------------------------- -----------------
alabaster                     0.7.13
appdirs                       1.4.4
Babel                         2.12.1
build                         0.10.0
charset-normalizer            3.1.0
clang                         16.0.1.1
distro                        1.8.0
docutils                      0.19
exceptiongroup                1.0.0
gpg                           1.19.0
idna                          3.4
imagesize                     1.4.1
importlib-metadata            6.6.0
iniconfig                     2.0.0
installer                     0.7.0
Jinja2                        3.1.2
libcomps                      0.1.19
Markdown                      3.4.1
MarkupSafe                    2.1.2
packaging                     23.1
pluggy                        1.0.0
pydantic                      1.10.7
Pygments                      2.15.1
pymdown-extensions            9.11
pyproject_hooks               1.0.0
pytest                        7.3.1
python-dateutil               2.8.2
pytz                          2023.2
PyYAML                        6.0
requests                      2.28.2
setuptools                    67.6.1
setuptools-scm                7.1.0
six                           1.16.0
snowballstemmer               2.2.0
Sphinx                        6.1.3
sphinx-jinja                  2.0.2
sphinxcontrib-applehelp       1.0.4
sphinxcontrib-devhelp         1.0.2.dev20230415
sphinxcontrib-htmlhelp        2.0.0
sphinxcontrib-jsmath          1.0.1.dev20230415
sphinxcontrib-qthelp          1.0.3.dev20230415
sphinxcontrib-serializinghtml 1.1.5
tomli                         2.0.1
typing_extensions             4.5.0
urllib3                       1.26.15
wheel                         0.40.0
zipp                          3.15.0
```
</details>

2bndy5:
Some docs updates that have accumulated.

Also (as part of the suggestions from #242 ) the pytest hard-coded options have been moved from pyproject.toml to nox_file.py

resolves #190
resolves #234
resolves #242


tests/README.rst
```diff
@@ -0,0 +1,51 @@
...
+
+The sphinx-immaterial theme should also be installed. For developer purposes, we recommend
+using an editable install, but this also requires that theme assets be assembled (as is done for
+distribution).
+
+.. code-block:: shell
+
+    npm install
+    npm run build
+    pip install -e .
```
jbms, line 21: I believe `pip install -e` also does the npm build automatically.  Is that not correct?

jbms:
Thanks!

tests/README.rst
2bndy5: No, we had someone ask recently why sphinx failed to find the html templates when it was using an editable install.

2bndy5:
I added a MyST example doc to satisfy #162. The MyST example doc is a bit raw, but it is meant to
encourage further contributions by providing a foundation to build on. Apparently, the blocking factor to user contribution is that [the original doc](https://github.com/executablebooks/MyST-Parser/blob/master/docs/syntax/typography.md) uses a private API (in [`myst_parser._docs` module](https://github.com/executablebooks/MyST-Parser/blob/a784880c7fbd893566087b83deec081376740e2c/myst_parser/_docs.py#L276)) to behave similarly to our `rst-example` directive. However, the `myst-example` directive relies on a `div` directive that is hidden elsewhere in the MyST ecosystem (probably not meant to be used publicly as well).

tests/README.rst
jbms: Looks like the issue is that `pip install -e .` no longer runs the `develop` command, which is what we were relying on in setup.py to do the nodejs build.  It looks like there is a new mechanism that can be used to handle editable builds, but it is not clear to me exactly how to use it:

https://setuptools.pypa.io/en/latest/userguide/extension.html#supporting-sdists-and-editable-installs-in-build-sub-commands

2bndy5: I honestly don't mind the way it works now. It cuts down on the initial editable install time significantly.

Although, I do understand the convenience for newcomers not having to remember the npm build step. Should this become a separate issue?

jbms:
> I added a MyST example doc to satisfy #162. The MyST example doc is a bit raw, but it is meant to encourage further contributions by providing a foundation to build on. Apparently, the blocking factor to user contribution is that [the original doc](https://github.com/executablebooks/MyST-Parser/blob/master/docs/syntax/typography.md) uses a private API (in [`myst_parser._docs` module](https://github.com/executablebooks/MyST-Parser/blob/a784880c7fbd893566087b83deec081376740e2c/myst_parser/_docs.py#L276)) to behave similarly to our `rst-example` directive. However, the `myst-example` directive relies on a `div` directive that is hidden elsewhere in the MyST ecosystem (probably not meant to be used publicly as well).

I think our existing `rst-example` directive could easily be modified to support myst as well.  In fact I think the only thing that needs to change is the language used for syntax highlighting, because when used from myst the `nested_parse` call will already do the right thing.  (I think the `div` directive used by the myst-example directive defined by myst-parser comes from sphinx-design.)

2bndy5:
> I think our existing rst-example directive could easily be modified to support myst as well.

I had the same thought a while ago: https://github.com/jbms/sphinx-immaterial/issues/162#issuecomment-1303338182. But that was more geared toward converting rST to MyST on-the-fly and rendering both with the appropriate parsers.

2bndy5:
I think the `myst-example` directive is more of a long-term and worthwhile goal. The solution here is only meant to satisfy #162 per user's request/feedback. I'm really just trying to close out as many docs-related issues with this PR.

tests/README.rst
jbms: I created an issue.  You don't need to re-run `pip install -e .` that often -- that is, after all, the whole point of an editable install.  So I think it is definitely desirable for it to just work by default.

jbms:
> I think the `myst-example` directive is more of a long-term and worthwhile goal. The solution here is only meant to satisfy #162 per user's request/feedback. I'm really just trying to close out as many docs-related issues with this PR.

Okay, sounds reasonable.

jbms (approved):

Issue #233, opened 2023-03-03
Consider switching from MathJax to MathML for math output


jbms:
For a long time Chrome has been the lone holdout in MathML support, making MathML basically unusable, but there is now support in Chrome 109, which is the current stable version:

https://caniuse.com/mathml

In a short time it should be even more widely deployed.

docutils already has support for generating mathml output, but I don't think Sphinx yet provides an easy way to take advantage of it.  I think we may also need some CSS from docutils for it to work well.

MathML's advantage over MathJax is that we don't need an extra javascript dependency, and it would render instantly with the rest of the content.  In contrast, with MathJax there is a lag during which you see the unrendered math markup and then the page reflows after it is rendered.

2bndy5:
>  In contrast, with MathJax there is a lag during which you see the unrendered math markup and then the page reflows after it is rendered.

BTW, I noticed the same lag with mermaid.js

2bndy5:
For reference to sphinx/docutils supporting MathML: sphinx-doc/sphinx#6092

2bndy5:
- original MathML specs are here: https://www.w3.org/TR/MathML3/
- https://developer.mozilla.org/en-US/docs/Web/MathML#mathml_reference has some useful references.

PR #251, opened 2023-04-28
Add support for html_math_renderer = 'mathml'

jbms:
Fixes https://github.com/jbms/sphinx-immaterial/issues/233.


jbms:
I did a bit of experimentation and was able to get MathML output working fairly easily.  However, it could benefit from some CSS styling to fix spacing issues.

This would actually be better suited as a change to Sphinx itself, though.

2bndy5:
Tagging https://github.com/sphinx-doc/sphinx/issues/6092 for visibility.

2bndy5:
> could benefit from some CSS styling to fix spacing issues

We could probably use the fonts from the npm mathjax dist when #231 gets finalized, but the code points might need translation or metrics might not match up. It might be easier to switch the font caching source to fontsource. I'm not nearly an expert though.

sphinx_immaterial/mathml_math_renderer.py
```diff
@@ -0,0 +1,40 @@
...
+from docutils.writers._html_base import HTMLTranslator as BaseHTMLTranslator
+from sphinx.application import Sphinx
+
+# Sphinx overrides Docutils' math rendering.  Here, we override Sphinx's
+# override to revert to docutils' math rendering.
+
+
+def visit_math(self: BaseHTMLTranslator, node: docutils.nodes.math):
+    self.math_output = "mathml"
+    self.math_output_options = []
```
2bndy5, line 14: Should this be extended to user confval(s)?  Does this directly influence the docutils' converter used?

jbms: This is basically equivalent to putting:

```
[html writers]
math_output: mathml
```

in a `docutils.conf` file next to the `conf.py` file.

There is one option that could be specified, an external LaTeX-syntax to MathML converter program to use in place of the one that is built into docutils.  Probably there should be a config option for that.

Alternatively this extension could be renamed "docutils_math_renderer" and we could require the user to create the separate docutils.conf file to specify the math format.  However, it seems unlikely that a user would want to use any of the docutils math formats other than mathml:
 - the `mathjax` format is already separately supported by sphinx, so I don't think there is any point in supporting that;
 - I don't know about the "HTML" format, but if we wanted to support that, we'd have to arrange to include the appropriate math.css stylesheet.

jbms:
> > could benefit from some CSS styling to fix spacing issues
> 
> We could probably use the fonts from the npm mathjax dist when #231 gets finalized, but the code points might need translation or metrics might not match up. It might be easier to switch the font caching source to fontsource. I'm not nearly an expert though.

Assuming it is a regular font I expect it will work fine.  MathML just encodes math as an element tree but leaves all of the actual layout up to the browser.  The MathML converter doesn't know anything about fonts.

I think the choice of font is independent of the issue I had noticed, where using "&" for alignment caused there to be extra space on either side, due to the default padding.

sphinx_immaterial/mathml_math_renderer.py
2bndy5: I'm really just working with the [info from here](https://docutils.sourceforge.io/docs/user/config.html#math-output). I think it would be fine to only use the docutils converter script (latex2mathml). But I was thinking it might be useful if a user already has external converter installed (like pandoc or blahxml) and might want to use that instead.

2bndy5:
I played around a bit with the rendered RTD build... I think the `columnspacing` can't be adjusted via CSS:
```css
mtable.align {
  columnspacing: 1em;
}
```
But, setting the `columnspacing` attr on the element does have a desirable effect. Setting `comunmalign="right left"` seems to work well also .

docs/additional_samples.rst
```diff
@@ -209,7 +209,7 @@ Math
 
    .. math::
 
-      (a + b)^2 = a^2 + 2ab + b^2
+      (a + b)^2 = a^2 + 2ab + b^2 \\
```
2bndy5, line 212: It might be worth noting when the escaped line breaks are needed/useful in a doc dedicated to this theme-ext

2bndy5:
docutils [release notes](https://docutils.sourceforge.io/RELEASE-NOTES.html) list 
> - "html5" writer:
>    - Change the default value for [math_output](https://docutils.sourceforge.io/docs/user/config.html#math-output) to "MathML" in Docutils 0.22.

jbms:
I added some minimal CSS to try to account for missing attribute support in Chrome.  I also copied in the docutils math example doc, and modified it to use rst-example.

I couldn't figure out how to make `\mathrm` and similar styles work in Chrome with CSS.  Firefox supports them by default.

jbms:
Overall it seems like the MathML support isn't quite ready, though maybe it is adequate for light use.

PR #157, opened 2022-08-31
Eliminate colon from headings in field lists

jbms:
Fixes #156.

This also refactors the various HTMLTranslator overrides.

PR #252, opened 2023-04-30
remove unnecessary permalink to `depart_term()`

### LINKED ISSUE #248: Glossary directive produces 2 permalinks
I just noticed this so I'm opening this to track it (& as a reminder). In [additional_samples.rst](https://github.com/jbms/sphinx-immaterial/blob/a738925ea04ec4e56000f81194fa04be561297bd/docs/additional_samples.rst?plain=1#L193):
image image

2bndy5:
resolves #248

The problem doesn't seem instigated by any changes to docutils or sphinx. So, it may have been something that was originally overlooked in #157.

Use `rst-example` for the `glossary` directive in additional_samples.rst

### DIFF
--- a/sphinx_immaterial/apidoc/object_toc.py
+++ b/sphinx_immaterial/apidoc/object_toc.py
@@ -294,7 +294,11 @@ def depart_term(
     node: docutils.nodes.term,
     super_func: html_translator_mixin.BaseVisitCallback[docutils.nodes.term],
 ) -> None:
-    self.add_permalink_ref(node, _("Permalink to this definition"))
+    next_node: docutils.nodes.Element = node.next_node(descend=False, siblings=True)
+    if not isinstance(next_node, docutils.nodes.classifier) and not isinstance(
+        node.parent.parent.parent, sphinx.addnodes.glossary
+    ):
+        self.add_permalink_ref(node, _("Permalink to this definition"))
     super_func(self, node)
 
 
--- a/sphinx_immaterial/apidoc/object_toc.py
+++ b/sphinx_immaterial/apidoc/object_toc.py
@@ -295,8 +295,11 @@ def depart_term(
     super_func: html_translator_mixin.BaseVisitCallback[docutils.nodes.term],
 ) -> None:
     next_node: docutils.nodes.Element = node.next_node(descend=False, siblings=True)
-    if not isinstance(next_node, docutils.nodes.classifier) and not isinstance(
-        node.parent.parent.parent, sphinx.addnodes.glossary
+    if (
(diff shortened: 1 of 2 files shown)


jbms:
The original code is intended to support this use case:

```rst
.. _a-term:

a term:
  a definition
```

In fact I was not aware that the default `depart_term` definition in Sphinx has a special case for glossary nodes:

https://github.com/sphinx-doc/sphinx/blob/ba7408209e84ee413f240afc20f3c6b484a81f8f/sphinx/writers/html5.py#L361

In fact I was not even familiar with the glossary directive at all.

I think it would be good to still support permalinks for definition list definitions outside of a glossary.

2bndy5:
Ah, ok. Should I revert and add a condition to exclude the extra permalink for the glossary directive?

jbms:
> Ah, ok. Should I revert and add a condition to exclude the extra permalink for the glossary directive?

Yeah, though it looks like a special case is also required for `nodes.classifier`.

sphinx_immaterial/apidoc/object_toc.py
```diff
@@ -294,7 +294,11 @@ def depart_term(
     node: docutils.nodes.term,
     super_func: html_translator_mixin.BaseVisitCallback[docutils.nodes.term],
 ) -> None:
-    self.add_permalink_ref(node, _("Permalink to this definition"))
+    next_node: docutils.nodes.Element = node.next_node(descend=False, siblings=True)
+    if not isinstance(next_node, docutils.nodes.classifier) and not isinstance(
+        node.parent.parent.parent, sphinx.addnodes.glossary
```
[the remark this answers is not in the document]
2bndy5: Good catch. I updated the condition.

sphinx_immaterial/apidoc/object_toc.py
```diff
@@ -294,7 +294,14 @@ def depart_term(
     node: docutils.nodes.term,
     super_func: html_translator_mixin.BaseVisitCallback[docutils.nodes.term],
 ) -> None:
-    self.add_permalink_ref(node, _("Permalink to this definition"))
+    next_node: docutils.nodes.Element = node.next_node(descend=False, siblings=True)
+    if (
+        not isinstance(next_node, docutils.nodes.classifier)
+        and node.parent is not None
```
jbms, line 300: This should be `and not (node.parent is not None and node.parent.parent is not None and isinstance(node.parent.parent.parent, sphinx.addnodes.glossary))`

2bndy5: done

jbms (approved):

PR #246, opened 2023-04-26
Styling version directives

2bndy5:
resolves #243

This does not override the original sphinx directives `versionadded`, `versionchanged`, and `deprecated`. Rather, the CSS styles have been adjusted to render these directives as admonitions. The default styles can be customized using the custom admonitions feature, but only `color` and `icon` will apply (as long as the `name` matches the directive name).

Also adds the italicized style inherent in Sphinx' basic theme CSS.

Updated admonitions.rst accordingly.

PR #253, opened 2023-05-01
support admonition options in version directives (follow-up to #246)

2bndy5:
The version directives now support admonition options including `collapsible`, `class`, and `name`.

Note `title` and `no-title` options are not supported because
1. the default title cannot be changed
2. the directives arguments (both required and optional) are used in the title

The custom admonitions feature's `classes` option is now also respected when set in conf.py.

### DIFF
--- a/docs/admonitions.rst
+++ b/docs/admonitions.rst
@@ -57,9 +57,28 @@ Version Directives
 ******************
 
 The `versionadded`, `versionchanged`, and `deprecated` directives defined by Sphinx are also
-rendered by the sphinx-immaterial theme as admonitions.
+overridden by the sphinx-immaterial theme to render as :rst:dir:`admonition`\ s with added
+functionality (like optionally making them :rst:`:collapsible:`).
 
-These admonition styles can be customized using the `Custom Admonitions`_ feature provided the
+Beware, these directive overrides do not support the :rst:`:title:` and
+:rst:`:no-title:` options because
+
+1. They require 1 argument (and optionally 1 additional argument) that gets used in the title.
+2. The default title cannot be changed.
+
+.. confval:: sphinx_immaterial_override_version_directives
+
+   This theme can be configured to disallow these overrides if the new features cause a conflict
+   with other Sphinx extensions.
+
+   .. code-block:: python
+      :caption: Disallow overriding builtin version directives in conf.py
+
+      sphinx_immaterial_override_version_directives = False
+
+These admonition styles can be customized (despite the
+:confval:`sphinx_immaterial_override_version_directives` value) using the `Custom Admonitions`_
+feature provided the
 :py:attr:`~sphinx_immaterial.custom_admonitions.CustomAdmonitionConfig.name` matches the directive
 name, but only the
 :py:attr:`~sphinx_immaterial.custom_admonitions.CustomAdmonitionConfig.icon`,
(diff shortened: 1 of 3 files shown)


sphinx_immaterial/custom_admonitions.py
```diff
@@ -413,6 +486,10 @@ def setup(app: Sphinx):
     app.connect("env-check-consistency", add_admonition_and_icon_css)
     app.connect("config-inited", on_config_inited)
 
+    # override original version directives with custom derivatives
+    for name, __ in VERSION_DIR_STYLE.items():
```
jbms, line 490: not a bug but probably you intended just one underscore there?

jbms, line 490: Perhaps this should be gated on a config option?

2bndy5: Originally, yes. But `_` is actually already used as a translation function (see module imports).

2bndy5: Well the visit/depart methods are always overridden using the `html_translator_mixin.override` decorator. The directive overrides in `setup()` would also have to be conditional as well (using the same config option).

I didn't attempt this because the original functionality is still employed. All, I'm doing here is adding CSS styling in a way that fits with our other Custom Admonitions' features.

2bndy5: Done. I had to move some things into the visitor to ensure the CSS customization still works despite the value of `sphinx_immaterial_override_version_directives`.

Docs updated accordingly.

sphinx_immaterial/custom_admonitions.py
```diff
@@ -408,6 +490,12 @@ def setup(app: Sphinx):
         rebuild="html",
         types=bool,
     )
+    app.add_config_value(
+        name="sphinx_immaterial_override_version_directives",
+        default=True,
+        rebuild="html",
```
jbms, line 496: This should be "env" I think since it affects the doctree.

docs/admonitions.rst
```diff
@@ -57,13 +57,33 @@ Version Directives
 ******************
 
 The `versionadded`, `versionchanged`, and `deprecated` directives defined by Sphinx are also
-rendered by the sphinx-immaterial theme as admonitions.
+overridden by the sphinx-immaterial theme to render as :rst:dir:`admonition`\ s with added
```
jbms, line 60: perhaps leave as "rendered by this theme as admonitions, and also optionally overridden to support the additional admonition options collapsible, classes, ...

docs/admonitions.rst
```diff
@@ -57,13 +57,33 @@ Version Directives
...
+functionality (like optionally making them :rst:`:collapsible:`).
 
-These admonition styles can be customized using the `Custom Admonitions`_ feature provided the
+Beware, these directive overrides do not support the :rst:`:title:` and
```
jbms, line 63: This could be in a `.. warning`

docs/admonitions.rst
```diff
@@ -57,13 +57,33 @@ Version Directives
...
+
+   This theme can be configured to disallow these overrides if the new features cause a conflict
+   with other Sphinx extensions.
+
+   .. code-block:: python
+      :caption: Disallow overriding builtin version directives in conf.py
+
+      sphinx_immaterial_override_version_directives = False
+
+These admonition styles can be customized (despite the
```
jbms, line 79: perhaps "regardless of the setting of ..." instead of "despite"

docs/admonitions.rst
2bndy5: I know, but I didn't want to overuse admonitions in a doc the demonstrates admonitions (seemed like admonition litter).

docs/admonitions.rst
2bndy5: I'm not a fan of nesting prepositional phrases; "of ... of ..." may fail in translation to non-English languages. I like the "regardless" word choice though.

docs/admonitions.rst
2bndy5: It became a run-on sentence. I used the suggested wording in a additional sentence.

jbms (approved):

PR #256, opened 2023-05-05
allow user-interactive color changes in customization doc

2bndy5:
I revised the list of color `scheme`, `primary` and `accent` options, so users can click on the "badge" (actually a styled inline literal) to dynamically test the supported values. The corresponding example code-block's content is also dynamically adjusted accordingly. Similar to upstream's doc, the color changes do not persist after page (re)load.


docs/customization.rst
```diff
@@ -325,40 +282,69 @@ Configuration Options
...
 
             Primary color options are
 
-            :red:`red`, :pink:`pink`, :purple:`purple`, :deep-purple:`deep-purple`, :indigo:`indigo`, :blue:`blue`,
-            :light-blue:`light-blue`, :cyan:`cyan`, :teal:`teal`, :green:`green`, :light-green:`light-green`,
-            :lime:`lime`, :yellow:`yellow`, :amber:`amber`, :orange:`orange`, :deep-orange:`deep-orange`,
-            :brown:`brown`, :grey:`grey`, :blue-grey:`blue-grey`, :black:`black`, and :white:`white`.
+            .. jinja:: colors
+
+                {% for color in supported_primary -%}
```
jbms, line 315: How about just using the `raw` directive to directly emit HTML with inline `style` and `data-md-color-primary=xxxx` attributes here?  Then there would be no need to add any CSS styles, and instead you could access the color via `var(--md-primary-fg-color)`.  To support PDF output there could be a separate `.. only:: not html` section emitted as well.  (I think `not html` is allowed in only directives.)

jbms: You could also emit the inline javascript event handlers, which may be simpler than the separate script you have below.

2bndy5: > How about just using the raw directive 

This is originally how I started, but I wanted something that could easily translate (via sphinx native behavior) to other writers (especially non-HTML). I can't simply `app.add_node(name, cls, <visitors>)` in conf.py because the added node's class isn't picklable (due to the scope of the namespace for conf.py).

jbms: > > How about just using the raw directive
> 
> This is originally how I started, but I wanted something that could easily translate (via sphinx native behavior) to other writers (especially non-HTML). I can't simply `app.add_node(name, cls, <visitors>)` in conf.py because the added node's class isn't picklable (due to the scope of the namespace for conf.py).

For the other builders, though, it will just be a plain list of color names, with no actual colors applied (and obviously no dynamic behavior), so you could just emit plain literals.

I guess to bypass the pickle issue the node could be defined in a separate sphinx extension, e.g. `docs/sphinx_immaterial_docs_ext.py`.

2bndy5: > For the other builders, though, it will just be a plain list of color names, with no actual colors applied (and obviously no dynamic behavior), so you could just emit plain literals.

That's why I'm currently using the `docutils.nodes.literal`, just with no syntax highlighting and extra CSS classes. The CSS classes are still used to avoid having to compensate (or compromise) other writers, but maybe it wouldn't matter if the style was inlined.

> I guess to bypass the pickle issue the node could be defined in a separate sphinx extension, e.g. docs/sphinx_immaterial_docs_ext.py.

I was thinking this could be migrated to a badge extension (similar to sphinx-design's badges) in the theme later on. Upstream uses
```html
<button attr=color-value>color value</button>
```
But trying to mimic this from conf.py is how I hit the pickle problem. I like how the button tag automatically changes the mouse cursor, but that seemed like a personal preference.

2bndy5: Wouldn't using `nodes.raw` prevent the text from showing up entirely for writers that don't comply with the raw node's given `format` attr?

jbms: Yes, but then you can provide an alternative version using `only`.

2bndy5: Inlining the JS is proving painful because of the curly brackets.

2bndy5: The `only` directive can't be nested within a `jinja` directive, so there's still some duplicate code.

docs/conf.py
```diff
@@ -527,7 +557,72 @@ def _parse_confval_signature(
     return signature
 
 
+class TestColor(SphinxRole):
+    color_type: str
+    style = (
+        "background-color: %s;"
```
jbms, line 563: Can we get this from a CSS var so that we don't have to duplicate the colors used by the theme in `supported_colors`?

docs/customization.rst
```diff
@@ -337,40 +294,92 @@ Configuration Options
...
-            :light-blue:`light-blue`, :cyan:`cyan`, :teal:`teal`, :green:`green`, :light-green:`light-green`,
-            :lime:`lime`, :yellow:`yellow`, :amber:`amber`, :orange:`orange`, :deep-orange:`deep-orange`,
-            :brown:`brown`, :grey:`grey`, :blue-grey:`blue-grey`, :black:`black`, and :white:`white`.
+            .. only:: html
+
+                .. jinja:: colors
+
+                    {% for color in supported_primary -%}
+                    :test-color-primary:`{{color}}`
+                    {%- if color != 'white' %},{% if color == 'black' %} and{% endif %} {% endif %}
```
jbms, line 336: Rather than checking for specific colors you could use `loop.index` and similar variables:
https://stackoverflow.com/questions/12145434/how-to-output-loop-counter-in-python-jinja-template

Alternatively could just use a bullet list and not worry about the punctuation

jbms: Or maybe hlist

docs/conf.py
2bndy5: The palette names are only used in CSS selectors. It may be inefficient to crawl the CSS bundle for the supported palette colors.

On a side note, I've been working up a `social_cards` ext that should mimic the upstream's [builtin social plugin](https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/#+social.cards_layout_options). Coincidentally, the list of [supported palette colors are saved in the python source for the plugin](https://github.com/squidfunk/mkdocs-material/blob/1b3b8a719db8209c2bd9e6ee9ddaef5c78420fc7/src/plugins/social/plugin.py#L497-L519). So, this dict will be migrated out of conf.py when I have something ready to submit.

docs/customization.rst
2bndy5: Yeah, I was thinking hlist as well.

2bndy5: done

docs/conf.py
jbms: I found that the color names can be extracted from the css using the following script:

```python
import pathlib
import re


def get_colors(t):
    unique_colors = {m.group(1) for m in re.finditer(
        r"\[data-md-color-" + t + r"=([a-z\-]+)\]",
        pathlib.Path("sphinx_immaterial/bundles/stylesheets/palette.css").read_text(
            encoding="utf-8"
        ),
    )}
    return sorted(unique_colors)


print(f'Primary: {get_colors("primary")}')
print(f'Accent: {get_colors("accent")}')
```

Getting the color values corresponding to these names from the CSS might be kind of annoying, but I think there is a workaround: set the `data-md-color-primary` or `data-md-color-accent` attribute on the `button` element; then you can access the color values from other style rules on the button via css variables.

2bndy5: > set the data-md-color-primary or data-md-color-accent attribute on the button element

Yep, I was going to do that also, but I got hung up writing #261 proposal. I'll also play around with the other inlined CSS to see what can go or is still required.

2bndy5: As for the scripting idea, I think it is kinda moot since I still need the color data for the social_cards ext. I can add it as a temp workaround for now. For the social_cards ext, I'll also try expanding on your snippet to extract the color data that corresponds to the color name.

2bndy5: done. I added a lambda to sort the colors a little better, but it still looks a bit helter-skelter:
image

jbms (approved):
Thanks, looks great!

Issue #115, opened 2022-07-02
namespace need to exactly match `external_cpp_references` config


2bndy5:
~This is unrelated to the changes here, but~ I'm reviewing the docs and I found the `external_cpp_references` example doesn't work when the parameter data type is explicitly of global scope (`::nlohmann::json`) rather it works when using `nlohmann::json` (as exactly defined in the conf.py setting). I'm not sure if there's a way to strip the global namespace using `cpp_strip_namespaces_from_signatures` option.

_Originally posted by @2bndy5 in https://github.com/jbms/sphinx-immaterial/issues/99#issuecomment-1172808163_

2bndy5:
relevant [comment from @jbms](https://github.com/jbms/sphinx-immaterial/pull/99#issuecomment-1172809367):
> The `external_cpp_references` extension won't be affected by `cpp_strip_namespaces_from_signatures` --- that option only affects how they are displayed in the HTML output, but not how the cross-referencing works.
> 
> In `external_cpp_references` I think we should always strip a leading `::` from any targets before trying to match them.
> 
> I agree that it would also be nice if `cpp_strip_namespces_from_signatures` could also support stripping a leading `::` though.

2bndy5:
In trying to reproduce this, I think the build env is somehow effecting this. If I add the global `::` and rebuild with `-a`, then the data type is still correctly hyperlinked (contradictory of the observation in OP that was based on a fresh build).

2bndy5:
While the original experience seems to have been solved in https://github.com/jbms/sphinx-immaterial/pull/99/commits/315e3bc710f3b34909e910ca99ad2b72e928b6a7 , the remaining observations about global scope operators still should be addressed.

2bndy5:
> In `external_cpp_references` I think we should always strip a leading `::` from any targets before trying to match them.

Locally, I've altered `external_cpp_references` to strip any leading `::` from the config's keys (on builder-inited event). I can submit a PR for review...

---
> I agree that it would also be nice if `cpp_strip_namespces_from_signatures` could also support stripping a leading `::` though.

In exploring this, I think I stumbled onto a separate issue. It looks like Sphinx has trouble interpreting C++ type identifiers if the type begins with a `::`.

When I made the following changes
```diff
--- a/docs/apidoc/cpp/index.rst
+++ b/docs/apidoc/cpp/index.rst
@@ -20,9 +20,9 @@ C++ domain customization
 
[...]

PR #257, opened 2023-05-05
update external_cpp_references

2bndy5:
- strips the global scope operator from confval's keys
- updates doc to use `autoclass` (obsoletes ignored nitpicky error)

This partially addresses the observations in https://github.com/jbms/sphinx-immaterial/issues/115#issuecomment-1172809791

### DIFF
--- a/docs/apidoc/cpp/external_cpp_references.rst
+++ b/docs/apidoc/cpp/external_cpp_references.rst
@@ -20,39 +20,24 @@ define the :confval:`external_cpp_references` configuration option:
     ]
 
 .. literalinclude:: /conf.py
-   :language: python
-   :start-after: # BEGIN: sphinx_immaterial.apidoc.cpp.external_cpp_references extension options
-   :end-before: # END: sphinx_immaterial.apidoc.cpp.external_cpp_references extension options
+    :language: python
+    :start-after: # BEGIN: sphinx_immaterial.apidoc.cpp.external_cpp_references extension options
+    :end-before: # END: sphinx_immaterial.apidoc.cpp.external_cpp_references extension options
 
 .. rst-example::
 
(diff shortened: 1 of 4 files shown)


docs/apidoc/cpp/external_cpp_references.rst
```diff
@@ -20,39 +20,24 @@ define the :confval:`external_cpp_references` configuration option:
...
-
-   - :python:`"class"`
-   - :python:`"union"`
-   - :python:`"function"`
-   - :python:`"member"`
-   - :python:`"type"`
-   - :python:`"concept"`
-   - :python:`"enum"`
-   - :python:`"enumerator"`
+    .. autoclass:: sphinx_immaterial.apidoc.cpp.external_cpp_references.ExternalCppReference
```
jbms, line 38: Can we exclude `__new__` from being documented since it isn't helpful here?

jbms:
This change looks good since we are already stripping leading ":" in `_strip_template_arguments`.  However, I'm rather confused why we weren't getting nitpick warnings before.

docs/apidoc/cpp/external_cpp_references.rst
2bndy5: Oh I missed that. Good catch!

2bndy5:
Nitpick warnings about what exactly? The only nitpick warning this relieves is the one about the python class.

docs/apidoc/cpp/external_cpp_references.rst
2bndy5: done

jbms:
> Nitpick warnings about what exactly? The only nitpick warning this relieves is the one about the python class.

Oh sorry, I was getting confused between `external_cpp_references` and the nitpick warnings on namespaces.

jbms (approved):

jbms (approved):

Issue #163, opened 2022-09-22
Theme does not work with rST Vscode Extension


twodrops:
I really like the immaterial theme, but it does not work with [rST VsCode Extension](https://marketplace.visualstudio.com/items?itemName=lextudio.restructuredtext). The problem is with the language server ([Esbonio](https://github.com/swyddfa/esbonio)) used by the extension that it crashes with the async Google font download.

I did not find a way to turn off the font download. It would be nice, if the font download is made optional.

2bndy5:
Can you post anything that tells us why the font is causing the extension to crash?

If it is a problem with the path to the fonts not being relative to the sphinx-build output, then you can use the config option [`sphinx_immaterial_external_resource_cache_dir`](https://jbms.github.io/sphinx-immaterial/customization.html#confval-sphinx_immaterial_external_resource_cache_dir) to tell the theme to download the fonts to your project's docs' build folder.

twodrops:
Unfortunately the extension does not seem to write much about this problem on the console even with debug log level.
> /usr/local/lib/python3.10/site-packages/esbonio/lsp/sphinx/__init__.py:164: RuntimeWarning: coroutine 'add_google_fonts.<locals>.do_fetch' was never awaited

[sphinx_immaterial_external_resource_cache_dir](https://jbms.github.io/sphinx-immaterial/customization.html#confval-sphinx_immaterial_external_resource_cache_dir) worked well with local build but not with the build triggered by the extension,

There is some suggestion by the Esbonio developers here
https://github.com/swyddfa/esbonio/issues/451#issuecomment-1255633191

twodrops:
Maybe it is notable to say that I am using a VsCode DevContainer. But most other themes work fine with this setup and Esbonio.

twodrops:
@2bndy5 I found another issue which is potentially due to nesting of asyncio.

I have a python module which is documented with autodoc. The module has a "async def" method and immaterial theme does not seem to like this. Other themes work fine with the module.

```python
async def run(
    name: str,
    start_commands: Sequence[Dict[str, Any]] = ()
) -> None:
    """Runs something.

    :param name: a unique name.
    :param start_commands: some command.
    """
```

Throws during sphinx-build

> module.py:docstring of module.run:1: WARNING: Parameter name 'start_commands' does not match any of the parameters defined in the signature: ['name', 'start_commands=()']

[...]

PR #258, opened 2023-05-05
Fix interoperability with Esbonio language server (#163)

tim-nordell-nimbelink:
The Esbonio language server runs its own asyncio loop context when it invokes the Sphinx code.  This causes the asyncio.run() call to complain since it's already being run in the same thread.

Since we're already running a set of threads here using PoolExecutor, we can run one more thread which runs the asyncio.run() context in a separate thread.  This mitigates the interoperability issue with the Esbonio language server at the cost of an additional thread, and at the cost of blocking the coroutine loop within the Esbonio language server for the closing of the download set.

(See also swyddfa/esbonio#451)

### DIFF
--- a/sphinx_immaterial/google_fonts.py
+++ b/sphinx_immaterial/google_fonts.py
@@ -67,7 +67,7 @@ def add_google_fonts(app: sphinx.application.Sphinx, fonts: List[str]):
     font_dir = os.path.join(static_dir, "fonts")
     os.makedirs(font_dir, exist_ok=True)
 
-    with concurrent.futures.ThreadPoolExecutor(max_workers=32) as executor:
+    with concurrent.futures.ThreadPoolExecutor(max_workers=33) as executor:
 
         def to_thread(fn, *args, **kwargs) -> asyncio.Future:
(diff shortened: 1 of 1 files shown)


jbms:
Thanks!

2bndy5:
Does this mean we can close out #163? My dev env doesn't seem to like the esbonio extension in VSCode (hit some other non-related issues with the esbonio settings), so my local attempt failed.

tim-nordell-nimbelink:
Probably can be closed?  I'm using Esbonio within NeoVIM, but did test these changes within VSCode, too.  Maybe @twodrops could test it since he opened the original issue in this project?

2bndy5:
I got it to work, so I closed the issue. It would be good to get a second opinion, but @twodrops noted using a dev container and this fix hasn't been deployed to pypi yet (can be installed with the nightly https://test.pypi.org/project/sphinx-immaterial/0.11.3.post1.dev23/).

twodrops:
Works like a charm from within DevContainer!
Thanks a lot @tim-nordell-nimbelink @2bndy5

PR #260, opened 2023-05-06
remove repo_type and make repo_name intuitive

### LINKED ISSUE #259: `repo_type` not used and `repo_name` can be intuitive
## `repo_type`
I think this conf.py option `repo_type` was actually inherited from the sphinx-material theme. But it isn't used in any JS or HTML templates. Removing it could be considered a breaking change because removing the option from theme.conf would invalidate its usage. Unfortunately, the current docs state (as inherited from sphinx-material):

> Must be one of github, gitlab or bitbucket.

which sounds misleadingly required.

## `repo_name`
In accordance with [upstream's behavior](https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#repository-name), the `repo_name` conf.py option can be automatically extracted from a given `repo_url` value. Currently, it is mandatory (as inherited from sphinx-material) without this added intuition, so this idea isn't a breaking change.

The URL structure for github, gitlab, and bitbucket all ends in `user/repo_name`. Sourceforge URLs or a URL starting with `git://` do not use the same structure, but these aren't explicitly supported by upstream. In fact the JS really only expects github or gitlab URLs for fetching repo info (if consent permits). In conclusion, the `repo_name` should only be extracted if the domain is of github, gitlab, or bitbucket; any other git-hosting service (like svn) will require the `repo_name` set explicitly.

2bndy5:
resolves #259

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -176,7 +176,7 @@ Configuration Options
 
     .. themeconf:: repo_url
 
-        The link to the repository will be rendered next to the search bar on big screens and as
+        The link to the repository will be rendered next to the search bar on large screens and as
         part of the main navigation drawer on smaller screen sizes. Additionally, for public
         repositories hosted on `GitHub <https://github.com>`_ or `GitLab
         <https://about.gitlab.com/>`_, the number of stars and forks is automatically requested and
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -176,15 +176,27 @@ Configuration Options
 
     .. themeconf:: repo_url
 
-        Set the repo url for the link to appear.
+        The link to the repository will be rendered next to the search bar on big screens and as
+        part of the main navigation drawer on smaller screen sizes. Additionally, for public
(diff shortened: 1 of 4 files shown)


2bndy5:
I feel I should apologize the barrage of PRs in the last week. I just happened to find the time, and... Well, you know how passionate I am about this theme.

As usual, there's no rush with any of my proposals. Since the changes to customization.rst here might conflict with #256, I'd prefer #256 get reviewed first.

docs/customization.rst
```diff
@@ -176,15 +176,27 @@ Configuration Options
 
     .. themeconf:: repo_url
 
-        Set the repo url for the link to appear.
+        The link to the repository will be rendered next to the search bar on big screens and as
```
jbms, line 179: "large screens"

jbms (approved):

docs/customization.rst
2bndy5: +1 Admittedly, this was copied from upstream's docs.

PR #266, opened 2023-05-31
document the icon[logo] option

2bndy5:
This option is already used in the HTML template.
So I figured I'd add the necessary documentation for it.

PR #267, opened 2023-05-31
satisfy warning about TypedDict

2bndy5:
Not sure what instigated this warning, but it reads:
>/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/sphinx_immaterial/apidoc/cpp/external_cpp_references.py:docstring of sphinx_immaterial.apidoc.cpp.external_cpp_references.ExternalCppReference:1: WARNING: py:class reference target not found: typing_extensions.TypedDict

I think there were some [changes to typing-extensions](https://github.com/python/typing_extensions/blob/main/CHANGELOG.md) about [`typing-extensions.TypedDict`](https://typing-extensions.readthedocs.io/en/latest/#TypedDict), but I don't see why we can't just use the [implementation available in native python](https://docs.python.org/3.11/library/typing.html?highlight=typing#typing.TypedDict) (v3.8 or newer).

### DIFF
--- a/sphinx_immaterial/apidoc/cpp/api_parser.py
+++ b/sphinx_immaterial/apidoc/cpp/api_parser.py
@@ -37,6 +37,7 @@ import json
 import os
 import pathlib
 import re
+import sys
 import time
 import typing
 from typing import (
@@ -51,6 +52,7 @@ from typing import (
     Pattern,
     Literal,
     Callable,
+    TypedDict,
 )
 from textwrap import dedent
 
@@ -70,9 +72,13 @@ import docutils.nodes
 import pydantic.dataclasses
 import sphinx.domains.cpp
 import sphinx.util.logging
-from typing_extensions import TypedDict, NotRequired
 
 from . import ast_fixes  # pylint: disable=unused-import
+
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
 
 logger = sphinx.util.logging.getLogger(__name__)
 
--- a/sphinx_immaterial/apidoc/cpp/api_parser.py
+++ b/sphinx_immaterial/apidoc/cpp/api_parser.py
@@ -37,7 +37,6 @@ import json
 import os
 import pathlib
 import re
-import sys
 import time
 import typing
 from typing import (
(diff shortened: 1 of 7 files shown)


2bndy5:
The warning seems like an intersphinx problem, and we can now add the typing-extensions intersphinx inventory.
```py
intersphinx_mapping = {
    "python": ("https://docs.python.org/3", None),
    "sphinx_docs": ("https://www.sphinx-doc.org/en/master", None),
    "MyST parser docs": ("https://myst-parser.readthedocs.io/en/latest", None),
    "typing-extensions": ("https://typing-extensions.readthedocs.io/en/latest", None),
}
```
Although this isn't really needed for our docs with the changes made here. I'm just commenting this for posterity I guess.

jbms:
We now actually require Python 3.8 so we can just always use `typing.TypedDict`.

2bndy5:
Most of our other typing-extensions usage is for importing `Literal`, but there's one [import of `NotRequired`](https://github.com/jbms/sphinx-immaterial/blob/f5e3213c511e4960aa5b4758b92de4652d1807c0/sphinx_immaterial/apidoc/cpp/api_parser.py#L73) which is [now available in python v3.11](https://docs.python.org/3.11/library/typing.html?highlight=literal#typing.NotRequired).

I 'm going to re-work this PR to factor out the dependency where possible.

2bndy5:
Done. This is needed to get CI passing again. The warning stated in OP is from the CI build logs for #266.

sphinx_immaterial/apidoc/cpp/api_parser.py
```diff
@@ -70,10 +72,14 @@
...
 import sphinx.domains.cpp
 import sphinx.util.logging
-from typing_extensions import TypedDict, NotRequired
 
 from . import ast_fixes  # pylint: disable=unused-import
 
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
```
jbms, line 81: This logic is basically already present in typing_extensions, but I'm not sure it is worth moving this condition into our code to avoid a dependency on typing_extensions when using 3.11, given that typing_extensions may likely already be installed and is pretty lightweight.  It is likely that we will depend on other stuff from typing_extensions in the future, so we may end up having to add the dependency back even for 3.11.

2bndy5: You want me to revert this and re-add the dependency to the theme (instead of the cpp optional deps)?

jbms: Yeah I think it makes sense to just keep an unconditional dependency on typing_extensions

2bndy5: done

jbms:
Thanks!

Issue #265, opened 2023-05-21
Sphinx will have better support for multi-line python signatures.


2bndy5:
I was just [notified of an error when building with Sphinx (master branch)](https://github.com/spatialaudio/nbsphinx/pull/737#issuecomment-1556209339):

```
Exception occurred:
  File "/home/mg/git/sphinx/sphinx/domains/python.py", line 575, in handle_signature
    signode += _parse_arglist(arglist, self.env, multi_line_parameter_list)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: _monkey_patch_python_parse_arglist.<locals>.parse_arglist() takes from 1 to 2 positional arguments but 3 were given
```
This stems from https://github.com/sphinx-doc/sphinx/commit/86b07d4a97a225e79150d14e25a768ebc4c087cc

Concerns the monkey_patch for default values to args:
https://github.com/jbms/sphinx-immaterial/blob/f5e3213c511e4960aa5b4758b92de4652d1807c0/sphinx_immaterial/apidoc/python/style_default_values_as_code.py#L9-L16

2bndy5:
This specific to sphinx v7.1.0

pradyunsg:
FWIW, the Sphinx-side changes influence C, C++, JS *and* Python.

jbms:
I think we will at least need some css changes to just directly support this option.

This theme has two forms of signature line wrapping:

- a simple parameter wrapping that is similar to the feature added to shinx, although implemented with css rather than by inserting dl tags.  In practice this works well enough for Python and probably c, but doesn't work that well for c++ which has more complicated signatures.
- wrapping based on clang-format

They are basically independent, so we could remove the simple wrapping and leave the clang-format wrapping.

Our feature does have the advantage that it can be more selectively applied since it is configured using our object description options mechanism.

2bndy5:
> we could remove the simple wrapping and leave the clang-format wrapping.

I'm leaning towards this idea. Better to rely on functionality in sphinx than maintain our hacky workaround with CSS.

PR #271, opened 2023-07-13
Fix minor compatibility issues with Sphinx 7

jbms:


sphinx_immaterial/apidoc/python/style_default_values_as_code.py
```diff
@@ -12,9 +10,9 @@ def _monkey_patch_python_parse_arglist():
     orig_parse_arglist = sphinx.domains.python._parse_arglist
 
     def parse_arglist(
-        arglist: str, env: Optional[sphinx.environment.BuildEnvironment] = None
+        arglist: str, *args, **kwargs
```
2bndy5, line 13: This is the same solution I was going to suggest for #265. So, I think this should close #265, but I was wondering if their multi-line approach could be coupled with this theme's multi-line formatting of signatures. Probably better to keep them separate due to the options this theme has for clang-format on C family domains.

2bndy5 (approved):

2bndy5:
Yesterday, sphinx released v7.1.0 which requires this fix for compatibility with this theme. I'm pushing v0.11.6 tag to deploy this fix.

2bndy5:
Probably should have added sphinx v7 to the CI test matrix (and nox session).

PR #278, opened 2023-08-18
fix compatibility with Sphinx v7.2.0

### LINKED ISSUE #277: Issue with self.css_files attributes?
This might be a version issue, but I just hit this:

```
Exception occurred:
  File "/home/runner/.local/lib/python3.10/site-packages/sphinx_immaterial/__init__.py", line 90, in init_css_files
    self.css_files = [
AttributeError: can't set attribute 'css_files'
The full traceback has been saved in /tmp/sphinx-err-hu_d8fkw.log, if you want to report the issue to the developers.
```
Is it related to Python version or immaterial version? I'm not currently pinning anything. Thank you (and sorry for always opening bugs, I am terrible :laughing: )

2bndy5:
This fixes #277 along with other problems found with v7.2.0 changes. See https://github.com/jbms/sphinx-immaterial/issues/277#issuecomment-1683192295.


jbms (approved):
Thanks!

2bndy5:
Looks like Sphinx v7.2.0 dropped support for python v3.8. This complicates the CI a bit... My "all-or-nothing" trait wants to remove python v3.8 from the workflow, but I need to think on a better solution.

jbms:
> Looks like Sphinx v7.2.0 dropped support for python v3.8. This complicates the CI a bit... My "all-or-nothing" trait wants to remove python v3.8 from the workflow, but I need to think on a better solution.

I think it might be tricky to avoid repetition entirely but this may help:
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations

2bndy5:
The `exclude` option should work for the unit tests, but mypy and pylint don't seem to understand the changes here when using python v3.8. I can `type: ignore` the problems until we drop 3.8 entirely or I can just skip pylint and mypy for python v3.8 in CI.

2bndy5:
In updating the the unit tests, I found a warning about `python.apigen`:
```
tests/python_apigen_test.py: 91 warnings
  path\to\sphinx_immaterial\apidoc\python\apigen.py:1235: RemovedInSphinx80Warning: The tuple interface of ObjectMember is deprecated. Use (obj.__name__, obj.object) instead.
    members.sort(key=lambda entry: member_order.get(entry[0], float("inf")))
```

jposada202020:
Thank you @2bndy5!

PR #284, opened 2023-09-03
Add support for "scope" option

Bizordec:
Didn't find where to set [the "scope" option](https://github.com/jbms/sphinx-immaterial/blob/main/src/partials/javascripts/base.html#L30), so here it is.


2bndy5:
What does exposing this accomplish? I think the doc entry could be explained more. Is there info about this in the mkdocs-material theme?

Bizordec:
It's useful in case when you build subsites separately. For example, for this structure
```
.
└── /
    ├── A/
    ├── B/
    └── C/
```
each site will have its own state for palette (`/A/.palette`, `/B/.palette`, `/C/.palette)`. To share the same state between all subsites, you could set `"scope": "/"`.

In mkdocs-material it's somehow still not documented, but here's the description in this issue: https://github.com/squidfunk/mkdocs-material/issues/1914#issuecomment-882025594

2bndy5:
This seems useful for different versioned sets of docs. We do support custom cookies, but that's another feature that has gone undocumented here.

How have you tested this?

Bizordec:
Added test and better description.

I've also created a PR in mkdocs-material (https://github.com/squidfunk/mkdocs-material/pull/5967), guess it'll soon be documented here too 😃

2bndy5 (dismissed):

tests/extra_scope_test.py
```diff
@@ -0,0 +1,44 @@
...
+from bs4 import BeautifulSoup
+import re
+import textwrap
+
+if TYPE_CHECKING:
+    from sphinx.testing.util import SphinxTestApp
+    from bs4.element import Tag
+
+
+def test_extra_scope(immaterial_make_app):
```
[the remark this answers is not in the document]
Bizordec: Yeah, done.

docs/customization.rst
```diff
@@ -664,6 +664,12 @@ Configuration Options
 
         A list of dictionaries used to populate the :ref:`version_dropdown` selector.
 
+    .. themeconf:: scope
+
+        Scope of site preferences (i.e. cookie consent, content tabs, color palette).
+        If you have multi-site project, you can set this to "/"
+        to share preferences between all sub-sites.
```
2bndy5, line 671:
```suggestion
        to share preferences between all sub-sites. See the :ref:`version_dropdown` section 
        on how to setup a multi-site project.
```

tests/extra_scope_test.py
```diff
@@ -0,0 +1,64 @@
...
+            }}
+            """
+        ),
+        files={
+            "index.rst": "Sample text",
+        },
+    )
+    app.build()
+
+    with open(app.outdir / "index.html", mode="r") as file:
```
2bndy5, line 50: older versions of Sphinx (`<7.2`) treated `outdir` like a `str`. The following mypy warning is telling you this as well:
```
nox > mypy 
tests/extra_scope_test.py:50: error: Unsupported left operand type for /
("str")  [operator]
        with open(app.outdir / "index.html", mode="r") as file:
                  ^~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 94 source files)
```
```suggestion
    with open(pathlib.Path(app.outdir) / "index.html", mode="r") as file:
```

tests/extra_scope_test.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
```
2bndy5, line 3:
```suggestion
import pathlib
from typing import TYPE_CHECKING
```
See other comment.

docs/customization.rst
2bndy5, line 1: Ci check found trailing whitespace in here:
```
nox > Session check_trailing_space aborted: 
Fixing docs/customization.rst
.
```
You can double check these locally before committing by installing nox and just running it:
```
pip install nox
nox
```
Some checks actually auto-fix the errors they complain about (like the one above).

tests/extra_scope_test.py
Bizordec: Fixed

tests/extra_scope_test.py
Bizordec: Fixed

docs/customization.rst
Bizordec: TIL it's always better to add changes locally than by clicking "Commit suggestion" on  GitHub 😅 
Fixed!

2bndy5:
I should add `nox` info to our contributing guidelines.

2bndy5:
I've re-run the CI multiple times, but the docs build on the windows runner fails to download/cache the fonts from google's API:
```
http.client.RemoteDisconnected: Remote end closed connection without response
```
I doubt that this is specific to this PR, but I just wanted to pass that along. 🤷🏼‍♂️

Bizordec:
I think i've fixed that SSL problem by adding some pypi packages.
https://stackoverflow.com/a/46467942
https://github.com/Bizordec/sphinx-immaterial/actions/runs/6137800799

2bndy5:
I was starting to suspect 
1. the google server was changed
2. the Google API key we use to fetch fonts maybe outdated (🤷🏼‍♂️)

Bizordec:
Okay, it failed again... I don't get it... Is it possible to somehow reset this runner?

2bndy5:
Actually, I think the answer should be a short-lived cache of the fonts per CI run (specific to each OS). Currently, each unit test and docs build we do re-downloads the same sets of font families.

jbms:
We could cache if necessary but it does add some complications.  The failing requests seem to be the plain requests to retrieve individual fonts.  Those requests do not use an API key and occur anytime someone loads a page that uses Google fonts --- i.e. expected to get a lot of traffic.

I see that the CI is still failing, despite what was intended to fix it?

jbms:
Does it always fail with Python 3.11 on Windows, or is it random?

2bndy5:
It seems to be specific to windows runner, but I've seen it happen on macos too. Caching the fonts only ensures the tests can exec without re-downloading the fonts (saves a few seconds per job) -- see #292 CI runs.

Bizordec:
And now windows build failed even before creating a font cache 😕

2bndy5:
I've been starting to see the google fonts problem downstream in my sphinx-social-cards ext CI (specific to windows and mostly python v3.11).

In researching the possible problems, I noticed google fonts has a newer API to support variable fonts: https://developers.google.com/fonts/docs/css2. Could be worth a try (for fixed fonts), but I don't think variable fonts are supported in this theme.

2bndy5:
Maybe [this change in py3.11.5](https://docs.python.org/3/whatsnew/3.11.html#notable-changes-in-3-11-5) might have something to do with the `SSLEOFError` we keep seeing on the windows runners.

2bndy5:
I'm unable to reproduce this `SSLError` locally with Windows 11 + Python v3.11.5.

Bizordec:
Python v3.11.6 just got released, maybe it will fix the issue
https://docs.python.org/release/3.11.6/whatsnew/changelog.html#python-3-11-6

> Tools/Demos
[gh-109991](https://github.com/python/cpython/issues/109991): Update GitHub CI workflows to use OpenSSL 3.0.11 and multissltests to use 1.1.1w, 3.0.11, and 3.1.3.

2bndy5:
Unfortunately, the `actions/setup-python` still uses 3.11.5 as the latest.
image
I am eager to see if this was identified and fixed though.

sphinx_immaterial/__init__.py
```diff
@@ -235,13 +246,7 @@ def html_page_context(
...
                 "repo_url": theme_options.get("repo_url"),
                 "repo_name": theme_options.get("repo_name", None),
-                "extra": {
-                    "version": version_config,
-                    "social": theme_options.get("social"),
-                    "disqus": theme_options.get("disqus"),
-                    "manifest": theme_options.get("pwa_manifest"),
-                    "analytics": analytics,
-                },
+                "extra": extra,
```
2bndy5, line 249: Heads up: I just merged #298 which directly conflicts this diff:
https://github.com/jbms/sphinx-immaterial/blob/a805d83cb974a1d7bb9aeb68297c76eda34da223/sphinx_immaterial/__init__.py#L244

Bizordec: Fixed

2bndy5 (approved):
@jbms Any objections to merging this? I'd like to get this merged so I can push a tag and release the [changes labeled "pending release"](https://github.com/jbms/sphinx-immaterial/issues?q=is%3Aissue+label%3A%22pending+release%22+)

@Bizordec Thank you so much for your patience and work (& rebasing) here!

Bizordec:
Oh wow, build have finally passed! Although v3.11.5 was used 🤷‍♂️ 
Glad to help! 😄

2bndy5:
It would seem that the [github actions runners on windows have a cached version of python](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#python). Among the caches is python 3.11.5 (no later), so we either wait until the windows runner image is updated (takes a long time on purpose) or we explicitly tell [`actions/setup-python` to download the latest version](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#check-latest-version) from [github releases of `actions/python-versions`](https://github.com/actions/python-versions/releases).

2bndy5:
I still hit the SSLError in [CI using python 3.11.6 on Windows](https://github.com/2bndy5/sphinx-social-cards/actions/runs/6461205628/job/17540527002?pr=32#step:7:1081) (on a different repo using this theme in unit tests and docs builds). I think this might be more specific to google fonts API (probably a combination of py 3.11+ and google fonts API).

2bndy5:
`actions/setup-python` now uses v3.12.0 as the latest python version:
image

2bndy5:
released with [v0.11.8](https://pypi.org/project/sphinx-immaterial/0.11.8/)

PR #290, opened 2023-09-07
force all objinfo icons to visually account for text's descender

### LINKED ISSUE #289: the "p" of parameters in the API left sidebar is off
with respect to the other item types, the "p" is vertically off.

image

I thinks it's because you use both "height" and "line-height" parameters for the "obj-info" icons. as the small "p" goes under the line, it's off in the 16px square. 

I'll open a PR to simply change small "p" into upper "P" which will be centered as the others !

12rambau:
Fix #289 

| before                                                                                              	| after                                                                                               	|
|-----------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------------	|
|  	|  	|

### DIFF
--- a/sphinx_immaterial/nav_adapt.py
+++ b/sphinx_immaterial/nav_adapt.py
@@ -461,7 +461,7 @@ def _add_domain_info_to_toc(
             title_prefix = (
                 f'<span aria-label="{label}" '
                 f'class="objinfo-icon objinfo-icon__{toc_icon_class}" '
-                f'title="{label}">{toc_icon_text}</span>'
+                f'title="{label}">{toc_icon_text|upper}</span>'
             )
         span_prefix = "<span "
         assert entry.title.startswith(span_prefix)
--- a/sphinx_immaterial/nav_adapt.py
+++ b/sphinx_immaterial/nav_adapt.py
@@ -461,7 +461,7 @@ def _add_domain_info_to_toc(
             title_prefix = (
                 f'<span aria-label="{label}" '
                 f'class="objinfo-icon objinfo-icon__{toc_icon_class}" '
-                f'title="{label}">{toc_icon_text.upper()}</span>'
+                f'title="{label}">{toc_icon_text}</span>'
             )
         span_prefix = "<span "
         assert entry.title.startswith(span_prefix)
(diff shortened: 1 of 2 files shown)


sphinx_immaterial/nav_adapt.py
```diff
@@ -461,7 +461,7 @@ def _add_domain_info_to_toc(
             title_prefix = (
                 f'<span aria-label="{label}" '
                 f'class="objinfo-icon objinfo-icon__{toc_icon_class}" '
-                f'title="{label}">{toc_icon_text}</span>'
+                f'title="{label}">{toc_icon_text|upper}</span>'
```
12rambau, line 464:
```suggestion
                f'title="{label}">{toc_icon_text.upper()}</span>'
```

2bndy5:
Since the text is customizable (even the font being used is customizable), I think the better solution is to put the colored bg color in a pseudo element that won't affect the line height and encompass the text from descendor to ascendor (arbitrary of the font family used). This way people who enter a lower-case letter in their config don't get an unexpected capitol letter.
Probably somewhere in here:
https://github.com/jbms/sphinx-immaterial/blob/5756b7fabfde981e27a0cc1a3e9459bc897e9d7b/src/assets/stylesheets/main/_api.scss#L257-L263

2bndy5:
This solution assumes that all uppercase letters (in English) have no descendors, but `Q` actually does. Thus this won't fully resolve the issue:
image
using 
```python
    ("py:parameter", dict(toc_icon_class="data", toc_icon_text="Q")),
]
```

----
I have found that by changing the span element's display type to `inline-table`, then we get something more universally desirable.
### using config from above:
image
### using default config:
image
which was done with the following patch:
```diff
--- a/src/assets/stylesheets/main/_api.scss
+++ b/src/assets/stylesheets/main/_api.scss
@@ -239,7 +239,7 @@ $objinfo-icon-size: 16px;
 }

 .objinfo-icon {
-  display: inline-block;
+  display: inline-table;
   flex-shrink: 0;
   width: $objinfo-icon-size;
   height: $objinfo-icon-size;
```

12rambau:
I rollback my changes and opt for your solution instead, it works like a charm !

2bndy5 (approved):

PR #292, opened 2023-09-10
Contributing and CI Improvements

2bndy5:
Based on some discussion in #284


jbms:
The contributing guidelines look good.  As for the CI changes --- caching, via the github actions cache, is a bit tricky, because we need a key that will change if the set of fonts to be downloaded will change.  It is not clear what we would hash to obtain such a key.

2bndy5:
I just create a cache that's specific to the workflow run (and OS); it can't be re-used on a subsequent run but can be reused on re-run of failed jobs. If different fonts are specified, then I'd expect the google_fonts module to pick up on what is missing and download what it needs (even using the font cache).

jbms (approved):

PR #296, opened 2023-10-01
fix for pydantic-core for v2.10

2bndy5:
This is problem that mypy picked up since the release of `pydantic-core` v2.10

I also pinned `pydantic-core` to v2.10 due the breaking change introduced there: https://github.com/pydantic/pydantic-core/pull/980

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 sphinx>=4.5
 markupsafe
 pydantic>=2.0
+pydantic-core>=2.10
 pydantic-extra-types
 typing-extensions
 appdirs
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 sphinx>=4.5
 markupsafe
 pydantic>=2.0
+pydantic-core>=2.10
```
jbms, line 4: I think pydantic-core always needs to be exactly in sync with pydantic, so it might make more sense to just specify a constraint on pydantic.

2bndy5: I could pin to pydantic v2.4 instead (accomplishes the same version lock)

jbms (approved):

2bndy5:
I'm pretty sure this SSL problem is specific to python 3.11.5 on Windows. I can't reproduce it locally though.

PR #307, opened 2023-12-12
Fix the JavaScript search term splitting for compound words

### LINKED ISSUE #262: Search missing results
I recently noticed in [my docs](https://mhostetter.github.io/galois/latest/) that a search for a version specifier (eg "v0.3.5") yields no results, despite there being a page section with that exact title. I then noticed that typing "v0" yields many valid results, but typing `v0." yields nothing. I think there may a bug in the search string processing. Perhaps with handling periods?

## Type "v0"

image

## Type "v0."

image

jbms:
Previously, the JavaScript search code split the search query only on spaces, which meant that with a search query of a compound word like "mkdocs-material", for example, would be treated as a single term "mkdocs-material".  In contrast, when building the search index, the `/\w+/` regexp is used to split compound words, such that the index would only contain "mkdoc" and "material" as separate terms. Consequently, "mkdocs-material" as a single term would never be found.

With this change, the JavaScript search query splitting has been changed to match what is done when building the index.

Fixes #262.

### DIFF
--- a/src/assets/javascripts/sphinx_search.ts
+++ b/src/assets/javascripts/sphinx_search.ts
@@ -709,35 +709,48 @@ export async function getResults(query: string): Promise<SearchResultStream> {
   // Object search terms.
   const objectterms = []
 
-  for (const origTerm of splitQuery(query)) {
-    const lowerTerm = origTerm.toLowerCase()
+  for (let origTerm of splitQuery(query)) {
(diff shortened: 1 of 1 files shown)


2bndy5 (approved):

2bndy5:
I'll push a patch bump later when the CI is all caught up.

PR #310, opened 2024-01-11
Fix merge_from_mkdocs_material.py to handle renames

jbms:
merge_from_mkdocs_material.py now uses git-rebase to the do the merge, which works better than `git apply --3way`.


2bndy5 (approved):
It works now! Thanks!

PR #320, opened 2024-02-02
bump actions/checkout from 3 to 4; fix pypa/gh-action-pypi-publish warnings

2bndy5:
- v4 of the actions/checkout uses node.js v20; v3 uses node.js v16. GH deprecated actions that use node.js v16.
- Includes minor fixes to avoid warnings from pypa/gh-action-pypi-publish (follow up to #316).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
           - nox-sessions: 'mypy pylint'
             python-version: '3.12'
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup Node.js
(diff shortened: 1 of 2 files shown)


2bndy5:
I don't know why dependabot missed actions/checkout and actions/download-artifact as CI dependencies. I'm hoping it was just an init thing; that sort of thing doesn't usually happen on my other repos that use dependabot.

Lastly, since this PR finishes out the crusade-to-update-CI, I see we're still getting a warning about the setup-graphviz action. I have opened ts-graphviz/setup-graphviz#559, so we might have to revisit that problem in September when GH starts disallowing actions that still use node.js v16.

PR #325, opened 2024-02-08
Add note about ':kbd:' role

ruslo:


docs/keys.rst
```diff
@@ -73,6 +73,11 @@ Keys extension role
 
         :keys:`caps-lock`, :keys:`left-cmd+shift+a`, :keys:`backspace`
 
+.. note::
+
+  ``:keys:`` role can be used instead of
+  `:kbd: <https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-kbd>`_ role
```
2bndy5, line 79: We can simply use intersphinx and cross-ref the roles directly:
```suggestion
  The :rst:role:`keys` role can be used instead of :rst:role:`kbd` role
```

I also think this could be mentioned in the page's first paragraph instead of an admonished note after the introduction. For example, this line
https://github.com/jbms/sphinx-immaterial/blob/09d131cd3b6ca8f559e3672e05b6e1125b9e8092/docs/keys.rst#L6
could be augmented to
```rst
This extension is meant to mimic the python markdown extension, `pymdownx.keys`_.
It provides an alternative implementation to Sphinx' :rst:role:`kbd` role.
```

2bndy5 (approved):
Thanks!

PR #326, opened 2024-02-08
Fix few title underlines

ruslo:
Noticed them while checking #325. Also, I realized that Sphinx can warn about underlines that are too short but ignore underlines that are too long.

### DIFF
--- a/docs/keys.rst
+++ b/docs/keys.rst
@@ -61,7 +61,7 @@ And be sure to include the extension in your :file:`conf.py` file.
 .. _keys_extension_role:
 
 Keys extension role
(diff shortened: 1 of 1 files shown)


2bndy5:
> I realized that Sphinx can warn about underlines that are too short but ignore underlines that are too long.

Yep. This is why we don't really care about underlines that are too long. Underlines that are too short will break our docs build in CI.

2bndy5 (approved):
I'm OK with these changes, but it is really an aesthetic preference of doc authors. These changes pose no difference in how the docs are parsed or rendered.

PR #343, opened 2024-04-13
doc theme_result extension

### LINKED ISSUE #323: document the `theme_result` extension
I should document the `rst-example` directive (& its `myst-example` alias for .md doc sources).

I keep seeing possible use cases for this directive in user projects/examples. But there is no documentation to make users aware that the `sphinx_immaterial.theme_result` extension exists and what it provides.

For those passing by, the `rst-example` directive works by taking RST code as content and rendering an RST code block followed by a rendition of the RST code using the given sphinx builder (HTML-based is preferred/tested). This is used liberally throughout our docs and avoids the common copy-and-paste that doc authors would normally have to do to achieve a similar result.

2bndy5:
resolves #323

### DIFF
--- a/docs/_static/extra_css.css
+++ b/docs/_static/extra_css.css
@@ -21,6 +21,11 @@
   font-family: var(--md-text-font-family);
 }
 
+/* style for recursive rst-example rendering */
+div.results.recursive-rst-example > div.result {
+  padding-bottom: 1em;
+}
+
 /* ************************* animated-admonition-border style ******************************** */
 @keyframes flash_border {
   0% {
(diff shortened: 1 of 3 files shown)


jbms:
Looks good -- I'm okay with this as is, but I haven't see the term "showcase" used for this before.  Not sure I have a better term though.

2bndy5:
I couldn't think of a better term either. The extension isn't really specific to the theme, so "Theme Result" (as the module is named) or "Theme Demonstration" (despite vagueness and overuse) isn't really adequate.

"Showcasing" might be a cultural term (eg. a collector's usage as a verb), but I _have_ seen it used like this before. I just can't think of a specific example from past experiences.

jbms (approved):

PR #347, opened 2024-04-27
Fix search term match and highlight for Unicode chars

Bizordec:
A regular expression `/\w+/g` was added in PR #307, which broke the search for non-english words.

I changed it to `/\p{L}+/gu` to add support for all Unicode characters and removed redundant word boundary (`\b`) in highlight regex.


2bndy5:
> [!NOTE]
> The search implementation will be changing completely upon merging #338; it will more closely follow upstream's behavior known as "[the builtin search plugin](https://squidfunk.github.io/mkdocs-material/plugins/search)". I doubt this hack would be needed after that. Furthermore, this hack might become more of a [user customization](https://squidfunk.github.io/mkdocs-material/plugins/search/#search).

Bizordec:
Ok then, for now I'll rollback to v0.11.9, since this version has working search (but not highlighting) for Unicode.

PR #357, opened 2024-06-30
Support methods/properties/functions that are aliased

jbms:
Previously, if the same function, method or property was assigned to an additional name in the same scope, it was documented as an independent entity.

With this change, it is documented together with the entity that it aliases as an additional signature.


2bndy5 (approved):
This seems to satisfy a specific use case, but this is generic enough to be useful to all. 👍🏼

PR #358, opened 2024-07-02
python/apigen: Fix inheritance of sibling/alias members

jbms:
Previously, if class ``A`` defined a member ``p`` with an alias ``q``, and then ``B`` inherits from ``A``, the ``q`` alias would be listed twice, since it would be added when processing the members of ``A`` and then again when processing the members of ``B``..

### DIFF
--- a/sphinx_immaterial/apidoc/python/apigen.py
+++ b/sphinx_immaterial/apidoc/python/apigen.py
@@ -310,7 +310,8 @@ class _ApiEntityMemberReference(NamedTuple):
     name: str
     canonical_object_name: str
     parent_canonical_object_name: str
-    inherited: bool = False
+    inherited: bool
+    siblings: List["_ApiEntityMemberReference"]
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


sphinx_immaterial/apidoc/python/apigen.py
```diff
@@ -593,18 +596,18 @@ def object_description_transform(
...
-    ]
+    if member is not None:
+        all_members = [member] + member.siblings
+        all_entities = [api_data.entities[m.canonical_object_name] for m in all_members]
+    else:
+        all_entities = [
+            entity,
+            *(api_data.entities[s] for s in (entity.siblings or {})),
+        ]
+        all_members = [None for _ in all_entities]
```
2bndy5, line 607:
```suggestion
        all_members = [None] * len(all_entities)
```
No need to traverse the list if you just want a parallel list of `None` objects

jbms: Done.

2bndy5 (approved):

PR #363, opened 2024-07-05
add config value to control font fetching max_workers

### LINKED ISSUE #351: sphinx_immaterial.google_fonts._MAX_CONCURRENT_FETCHES variable is unused
The `_MAX_CONCURRENT_FETCHES` module variable in `sphinx_immaterial.google_fonts` is currently unused:

```console
$ git grep _MAX_CONCURRENT_FETCHES
sphinx_immaterial/google_fonts.py:_MAX_CONCURRENT_FETCHES = 128
<end>
```

I presume this is supposed to be a way for users to monkeypatch a lower (or higher) concurrency limit to help with network issues, but it isn't used in the call to `ThreadPoolExecutor`.

Presuming my understanding is correct, is this the right fix?

```patch
diff --git a/sphinx_immaterial/google_fonts.py b/sphinx_immaterial/google_fonts.py
index 8f355063..80bef6d3 100644
--- a/sphinx_immaterial/google_fonts.py
+++ b/sphinx_immaterial/google_fonts.py
@@ -64,7 +64,7 @@ def add_google_fonts(app: sphinx.application.Sphinx, fonts: List[str]):
     font_dir = os.path.join(static_dir, "fonts")
     os.makedirs(font_dir, exist_ok=True)

-    with concurrent.futures.ThreadPoolExecutor(max_workers=33) as executor:
+    with concurrent.futures.ThreadPoolExecutor(max_workers=_MAX_CONCURRENT_FETCHES) as executor:

         def to_thread(fn, *args, **kwargs) -> asyncio.Future:
             return asyncio.wrap_future(executor.submit(fn, *args, **kwargs))
```

2bndy5:
resolves #351

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -742,6 +742,19 @@ Configuration Options
    bundles to the output directory.  Source maps are helpful when developing the
    theme.
 
+.. confval:: sphinx_immaterial_font_fetch_max_workers
+
+    The number of workers executed in parallel when fetching a cache of the specified
+    :themeconf:`font`. If not specified, this defaults to using 33 maximum *possible* threads.
+    If set less than or equal to 0, then this will be determined by the
+    :py:class:`~concurrent.futures.ThreadPoolExecutor` default.
+
+.. envvar:: SPHINX_IMMATERIAL_FONT_FETCH_MAX_WORKERS
+
+    An environment variable that can be used to override
+    :confval:`sphinx_immaterial_font_fetch_max_workers`.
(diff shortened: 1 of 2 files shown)


jbms:
Since it is network latency-limited rather than cpu-limited I think a fixed default rather than the number of CPU cores is better.

2bndy5:
I also considered supporting an env var in addition to a confval. Maybe it should only be an env var?

jbms (approved):

PR #366, opened 2024-07-10
Add support for Python type parameter lists

jbms:
Sphinx has partial support for type parameter lists: they are supported by the Python domain in signatures, but are not supported by autodoc.

This adds the following support:

- Sphinx Python domain for type parameter fields in docstrings, with sphinx.ext.napoleon support as well.

- Support for type parameters as Sphinx objects, with cross-linking, like the existing support for function parameters as Sphinx objects.

- Support in apigen for PEP 695 type parameters, and for displaying pre-PEP 695 separately-defined TypeVar types as PEP 695 type parameters.

### DIFF
--- a/docs/tensorstore_demo/__init__.py
+++ b/docs/tensorstore_demo/__init__.py
@@ -735,8 +735,8 @@ class DimExpression():
             })
 
         More generally, specifying an ``n``-dimensional `bool` array is equivalent to
-        specifying ``n`` 1-dimensional index arrays, where the ``i``\ th index array specifies
-        the ``i``\ th coordinate of the `True` values:
+        specifying ``n`` 1-dimensional index arrays, where the ``i``\\ th index array specifies
+        the ``i``\\ th coordinate of the `True` values:
 
             >>> x = ts.array([[[1, 2, 3], [4, 5, 6]], [[7, 8, 9], [10, 11, 12]]],
             ...              dtype=ts.int32)
@@ -1049,8 +1049,8 @@ three cases:
           :python:`dims[i] = self.labels.index(other.labels[i])`.  It is an
           error if no such dimension exists.
 
-       2. Otherwise, ``i`` is the ``j``\ th unlabeled dimension of :python:`other`
-          (left to right), and :python:`dims[i] = k`, where ``k`` is the ``j``\ th
+       2. Otherwise, ``i`` is the ``j``\\ th unlabeled dimension of :python:`other`
+          (left to right), and :python:`dims[i] = k`, where ``k`` is the ``j``\\ th
           unlabeled dimension of :python:`self` (left to right).  It is an error
           if no such dimension exists.
 
--- a/sphinx_immaterial/apidoc/python/apigen.py
+++ b/sphinx_immaterial/apidoc/python/apigen.py
@@ -20,6 +20,7 @@ import importlib
 import inspect
 import json
 import re
+import typing
 from typing import (
     List,
     Tuple,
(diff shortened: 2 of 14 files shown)


jbms:
This is basically feature complete but still needs some work on documentation and testing.

jbms:
The build failure due to undefined reference `T` is odd --- the build succeeds on readthedocs and I also can't reproduce the issue locally.  I will have to investigate more.

jbms:
CI now passes.

docs/tensorstore_demo/__init__.py
```diff
@@ -735,8 +735,8 @@ def __getitem__(self, *args, **kwargs) -> None:
             })
 
         More generally, specifying an ``n``-dimensional `bool` array is equivalent to
-        specifying ``n`` 1-dimensional index arrays, where the ``i``\ th index array specifies
-        the ``i``\ th coordinate of the `True` values:
+        specifying ``n`` 1-dimensional index arrays, where the ``i``\\ th index array specifies
+        the ``i``\\ th coordinate of the `True` values:
```
2bndy5, line 739: I'm guessing this is done to resolve warnings about invalid escape sequences. I hadn't thought that we'd need to escape the escaping character in a literal string used as a doc string. It makes sense though.

I'll have to remember this as it still seems to [render as expected in the docs](https://sphinx-immaterial--366.org.readthedocs.build/en/366/python_apigen_generated/DimExpression.__getitem__.html#boolean-array-indexing).

sphinx_immaterial/apidoc/python/apigen.py
```diff
@@ -266,6 +268,10 @@ def _must_shorten():
         if not added_ellipsis:
             added_ellipsis = True
             ellipsis_node = sphinx.addnodes.desc_sig_punctuation("", "...")
+            # When using the `sphinx_immaterial.apidoc.format_signatures`
+            # extension, replace the text of this node to make it valid Python
+            # syntax.
+            ellipsis_node["munged_text_for_formatting"] = "___"
```
2bndy5, line 274: Should probably merge #367 before reviewing this yes? It might be easier for me to differentiate what changes are actually specific to this PR's goal.

docs/tensorstore_demo/__init__.py
jbms: Python ignores invalid escape sequences, so they still work correctly, but I wanted to avoid the warnings.

sphinx_immaterial/apidoc/python/apigen.py
jbms: This hunk is also in #367 --- I accidentally included it here originally, but I have now removed it.

sphinx_immaterial/apidoc/python/parameter_objects.py
```diff
@@ -1,18 +1,42 @@
...
 from .. import object_description_options
 from ... import sphinx_utils
 
-
 logger = sphinx.util.logging.getLogger(__name__)
 
+if sphinx.version_info >= (7, 1):
+    from sphinx.addnodes import desc_type_parameter
+else:
+    # Fake class definition that will always fail instanceof checks.
```
2bndy5, line 36:
```suggestion
    # Fake class definition that will always fail isinstance() checks.
```
this is in python after all 😜

2bndy5 (changes requested):

sphinx_immaterial/apidoc/python/apigen.py
```diff
@@ -519,6 +529,80 @@ def _clean_class_getitem_signature(signode: sphinx.addnodes.desc_signature) -> N
...
+    parent_type_params: tuple[type_param_utils.TypeParam, ...],
+    is_constructor: bool,
+) -> bool:
+    def _make_type_list_node() -> list[docutils.nodes.Node]:
+        tp_list = type_param_utils.stringify_type_params(parent_type_params)
+        tp_list_node = sphinx.domains.python._annotations._parse_type_list(tp_list, env)
+        for desc_param_node in tp_list_node.findall(
+            condition=sphinx.addnodes.desc_type_parameter
+        ):
+            desc_param_node["sphinx_immaterial_type_param_symbol_prefix"] = (
```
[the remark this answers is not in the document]
jbms: Done.

2bndy5 (approved):

PR #367, opened 2024-07-11
Add support for formatting Python signatures using black

jbms:


2bndy5:
What about [`ruff format`](https://docs.astral.sh/ruff/formatter/)? I switched to `ruff` a while back and haven't looked back since. 😃

jbms:
Potentially ruff could be supported in addition or instead.  Most of the work in this PR is related to the Python-specific munging of signatures that would be the same for ruff as well.

The advantage of `black` is that it has a Python API and therefore we can just format each signature individually using `black.format_str`.  For ruff, since it has to be invoked as a subprocess, we need some mechanism of batching to avoid the overhead of starting a new process for each signature.

For clang-format we already do that, by collecting all of the signatures, separating them with special identifying comments, and formatting all of them with a single run of `clang-format`.

However, there are a few caveats with applying the same approach to Python:

- black (and probably ruff also) is much less forgiving than clang-format about syntax errors.  A problem with any signature would mean none get formatted.  We would have to deal with this somehow.
- With the type parameters feature, apigen can produce signatures with two type parameter lists: https://sphinx-immaterial--366.org.readthedocs.build/en/366/python_apigen_generated/Map.get.html   I already have to munge the signatures to make them valid Python syntax so that black can handle them, but I couldn't think of a way to make this work in a single pass.  Instead, signatures with an extra type parameter list on the parent class are handled by first formatting up to the end of the first type parameter list, then formatting the remaining part with a suitable prefix added to make the line length limit work.  This could still be handled via a batch but just with two separate passes possibly being required, but it would add an extra complication.

Potentially the ruff LSP accessible via `ruff serve` could be used to efficiently issue a bunch of independent format requests.  I'm not sure how difficult that would be.

Overall, though, I expect `black` will be fast enough, and other than speed I don't know that there is any real advantage to using ruff, and therefore the added hassle of integrating ruff may not be worth it.

I'd be happy to switch this project itself to using ruff instead of black/pylint, though.

2bndy5:
> I'd be happy to switch this project itself to using ruff instead of black/pylint, though.

Yes, please! That should cut down the longest steps in our CI's check-convention job.

---
You make a good point about the lack of an API for ruff. One of the biggest unspoken hurtles with PyO3 is the compatibility with python ABI vs rust edition.

Now that you mention that this is only for signatures, I agree it may not be worth pursuing. My question was kind of a knee-jerk reaction; I recently fell in love with rust and am looking for ways to use it more and more.

2bndy5:
> image

🤣 That's just damn ironic for this PR.

jbms:
Most of the pylint failures are related to my using `list[T]` in place of `typing.List[T]` and similar.  It would certainly be possible to retain support for Python 3.8, but in the type parameter PR I already included a commit to drop Python 3.8 support, since at this point I don't think it is useful to keep support for it.  Or do you have a strong desire to keep Python 3.8 support still?

2bndy5:
Personally, I'm ok with dropping it. However, EoL for python 3.8 is still 3 months from now, so its just a little early. IMHO, it should probably be done in a separate PR because there is likely more than just type hints in terms of code specific to python 3.8 (or earlier).

jbms:
This now passes CI, and I added some documentation.

jbms:
This kind of works with the latex builder but the spacing doesn't line up correctly because the latex builder doesn't use a monospace font for the signatures.  I'm not sure how easily we could force a monospace font.

More generally, though, while building the latex docs (i.e. actually invoke `make` after running sphinx) I noticed that the latex build is not in a great shape --- the table of contents seems to be broken, and we get a lot of undefined references.

sphinx_immaterial/apidoc/format_signatures.py
```diff
@@ -275,9 +641,29 @@ def merge_info(
...
+@pydantic.dataclasses.dataclass
+class BlackFormatStyle:
+    line_length: Optional[int] = None
+    """Line length limit.
+
+    Defaults to the value of :objconf:`wrap_signatures_column_limit`.
+    """
+
+    string_normalization: bool = True
+    """Indicates whether to normalize quotes."""
```
2bndy5, line 653:
```suggestion
    """Indicates whether to normalize quotes. This corresponds to the inverse of Black
    CLI's `--skip-string-normalization <https://black.readthedocs.io/en/stable/
    usage_and_configuration/the_basics.html#s-skip-string-normalization>`_ option."""
```

Unfortunately, they use a `<h3>` element in black docs, so we can't use intersphinx + `:std:option:'skip-string-normalization'`.

2bndy5 (changes requested):

2bndy5:
> I noticed that the latex build is not in a great shape --- the table of contents seems to be broken, and we get a lot of undefined references.

Yeah, its been like that for a long time. I never devoted enough time to figure out and fix the problems. I'm happy enough just keeping the builder from breaking (for now).

sphinx_immaterial/apidoc/format_signatures.py
```diff
@@ -79,167 +118,489 @@ def apply(self, **kwargs: Any) -> None:
...
+
+    node: Optional[docutils.nodes.Text]
+    """Corresponding text node.
+
+    May be `None` if this component was added only to make the syntax valid for
+    the formatter and should not be preserved.
+    """
+
+
+class FormatApplyer:
```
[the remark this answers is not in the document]
jbms: Done.

sphinx_immaterial/apidoc/format_signatures.py
```diff
@@ -79,167 +118,489 @@ def apply(self, **kwargs: Any) -> None:
...
+        if "sig-wrap" in signature_to_modify["classes"]:
+            signature_to_modify["classes"].remove("sig-wrap")
+
+        signature_to_modify["is_multiline"] = False
+
+
+def _format_signature_with_black(
+    domain: str, objtype: str, node: sphinx.addnodes.desc_signature, options
+):
+    import black  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
jbms: Done.

sphinx_immaterial/apidoc/format_signatures.py
```diff
@@ -79,167 +118,489 @@ def apply(self, **kwargs: Any) -> None:
...
+        munged_prefix = (
+            ensure_prefix + munged_prefix[len(ensure_prefix) - extra_prefix_len :]
+        )
+        extra_prefix = ensure_prefix[:extra_prefix_len]
+
+        signature = munged_prefix + signature[len(initial_prefix) :]
+
+        try:
+            formatted = black.format_str(signature, mode=mode)
+        except Exception as e:  # pylint: disable=broad-exception-caught
```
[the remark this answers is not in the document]
jbms: Done.

sphinx_immaterial/apidoc/format_signatures.py
```diff
@@ -79,167 +118,489 @@ def apply(self, **kwargs: Any) -> None:
...
+    black_format_style = options.get("black_format_style")
+    assert black_format_style is not None
+
+    mode = black.FileMode(
+        line_length=black_format_style.line_length
+        or options["wrap_signatures_column_limit"],
+        string_normalization=black_format_style.string_normalization,
+    )
+
+    applyer = FormatApplier()
```
2bndy5, line 315: I still get spelling errors for this variable.
```suggestion
    applier = FormatApplier()
```

FYI, I'm using a [vscode extension to check spelling](https://github.com/streetsidesoftware/vscode-spell-checker). We could add a [codespell](https://github.com/codespell-project/codespell) check to CI, but it would have a lot of problems with the member names from sphinx/docutils.

sphinx_immaterial/apidoc/format_signatures.py
```diff
@@ -254,7 +615,9 @@ def apply(self, **kwargs: Any) -> None:
             formatted_signature = formatted_signatures.get(signature_id)
             if formatted_signature is None:
                 continue
-            _format_signature(node, formatted_signature)
+            applyer = FormatApplier()
```
2bndy5, line 618: here too

2bndy5 (changes requested):

sphinx_immaterial/apidoc/format_signatures.py
jbms: Done.

sphinx_immaterial/apidoc/format_signatures.py
jbms: Done.

2bndy5 (approved):

PR #368, opened 2024-07-11
Drop Python 3.8 support

jbms:
While Python 3.8 is not officially EOL, NumPy dropped Python 3.8 support with NumPy 1.25 released Jun 2023.

https://numpy.org/news/#numpy-1250-released


2bndy5 (approved):

2bndy5:
should imply a minor version bump

PR #372, opened 2024-07-23
apidoc/python: Fix xrefs to type parameters

jbms:
Previously, if a function introduced a type parameter, any references to that type parameter in the signature itself were correctly resolved. However, when documenting function parameters, the type from the signature is duplicated next to the parameter name, and any references to type parameters in these duplicated types were not correctly resolved.

This commit fixes that issue by applying the same resolution logic to the duplicated types.

### DIFF
--- a/docs/type_param_demo.py
+++ b/docs/type_param_demo.py
@@ -70,7 +70,11 @@ class Map(Generic[K, V]):
     def get(self, key: K, default: T) -> Union[V, T]: ...
 
     def get(self, key: K, default=None):
-        """Return the mapped value, or the specified default."""
+        """Return the mapped value, or the specified default.
+
+        :param key: Key to retrieve.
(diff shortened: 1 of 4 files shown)


2bndy5:
Not sure how you discovered this. It might be worth adding `:param:` fields to the demo API docs in https://github.com/jbms/sphinx-immaterial/blob/f5e3d7c120740ff3fedb6e2391234fcaacc61397/docs/type_param_demo.py#L63-L73

jbms:
> Not sure how you discovered this. It might be worth adding `:param:` fields to the demo API docs in

Done.

Without this fix, the `T` in `default :T` as shown separately for the function parameter leads to an ambiguous reference warning if there is more than `T` anywhere, because it gets resolved as a top-down `T` reference.

2bndy5 (approved):

PR #376, opened 2024-07-28
support for Sphinx v8.0

2bndy5:
using the 8.0.0rc1 from pypi

Thanks to awareness in sphinx-doc/sphinx#12629

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,7 @@ jobs:
       - name: Build docs
         env:
           SPHINX_IMMATERIAL_EXTERNAL_RESOURCE_CACHE_DIR: ${{ github.workspace }}/theme_build_cache
-        run: nox -s docs
+        run: nox -s docs --forcecolor
       - name: Upload doc builds as artifacts
         uses: actions/upload-artifact@v4
         with:
@@ -149,11 +149,6 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
-        sphinx-version:
-          - "sphinx4"
-          - "sphinx5"
-          - "sphinx6"
-          - "sphinx7"
         node-version:
           - "16.x"
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


2bndy5 (commented):
I suppose we could add the git master branch of sphinx to our CI. But that might be over-zealous given our sporadic maintainence.

jbms (approved):

jbms:
> I suppose we could add the git master branch of sphinx to our CI. But that might be over-zealous given our sporadic maintainence.
> 
> Note
> 
> Dependabot does not support notifying us of pre-releases (that I'm aware of).

Could be a good idea.

2bndy5:
I've seen the master branch used in CI on the breathe project. It really just strained the sparse development efforts to maintain support for Sphinx' bleeding edge. We have enough monkey patching to worry about in Sphinx stable channel.

jbms:
Now that sphinx 8 is released it is probably time to submit this.

2bndy5:
First, I should probably add sphinx v8 to CI tests job.

jbms:
Yeah.  The number of combinations is getting quite high.  Potentially we should run more than one sphinx version per job to save on setup cost.

2bndy5:
Are you thinking the CI matrix for tests should only iterate over python version and node version?

2bndy5:
I think I could configure nox to skip certain sphinx versions if testing an unsupported python version.

jbms (approved):

2bndy5:
Ok I reduced the number of CI jobs by running tests per OS, python version and node.js version.

I also incorporated some [GitHub workflow commands to group logs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines) from certain nox sessions.
image
image

This should make it easier to read the CI logs when we need to find what went wrong in a certain nox session.

> [!note]
> There may be some discrepancies about which `log()` outputs (beginning and/or ending statements) are grouped because logger handlers acquire a stdout lock. Using `nox.session.log()` won't group the logs because GitHub expects the grouping commands to have nothing prefixed (the prefix `nox > ` will negate log grouping). I tried to preempt this by using a separate `Logger` instance (that doesn't propagate to root handlers), but it doesn't seem to be consistent, although it is better than using `print(..., flush=True)`.

It would be greatly beneficial (in CI runtime) to be able to cache the graphviz installation, but the amount of dependencies for that app make it near impossible to sustain.

PR #378, opened 2024-08-02
python/apigen: Fix support for pybind11 overloaded functions

jbms:
This was accidentally broken by
https://github.com/jbms/sphinx-immaterial/pull/357.

### DIFF
--- a/sphinx_immaterial/apidoc/python/apigen.py
+++ b/sphinx_immaterial/apidoc/python/apigen.py
@@ -119,10 +119,10 @@ class ParsedOverload(NamedTuple):
     overload_id: Optional[str] = None
     """Overload id specified in the docstring.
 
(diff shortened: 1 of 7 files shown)


2bndy5 (commented):
Its odd that the added test for this uses a pure python. Is it safe to assume pybind11 will generate bindings the same way in the future?

jbms:
Okay, I changed the test to use actual pybind11.

tests/python_apigen_test_modules/pybind11_overloaded.py
2bndy5, line 1: Is this file still needed?

jbms: Removed.

2bndy5 (approved):

PR #390, opened 2024-10-03
fix: support for graphviz v12.1.0

jbms:


2bndy5:
good thing we have that system font test. Looking into that test failure now.

2bndy5:
I also tested this locally with graphviz v11.0.0 and it still worked. Apparently, v12.0.0 through v12.1.0 were yanked from winget's distribution database.

2bndy5:
The coverage report indicates that we are not testing with a font that does not have a weight of 400. 🙄

I'm ready to move on this if you are ok with my added changes.

PR #392, opened 2024-10-11
Sphin 8.1.0 and MathDomain import

Novakov:
Sphinx 8.1.0 removed import of MathDomain into sphinx.ext.mathjax that resulted in sphinx-immaterial crashing with ImportError. Replaced with import from actual package that defines MathDomain.

Sphinx seems to be moving away from using get_domain to get domain instances, however it is available only from versions >=8.1.0, so I left get_domain call as it is currently.

Verified to work with Sphinx 8.0.2 and 8.1.0 but I'm not using mathjax domain.

### DIFF
--- a/sphinx_immaterial/local_mathjax.py
+++ b/sphinx_immaterial/local_mathjax.py
@@ -4,7 +4,8 @@ from typing import Dict, Any, cast
 from sphinx.application import Sphinx
 from sphinx.config import Config
 from sphinx.errors import ExtensionError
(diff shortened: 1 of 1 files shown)


2bndy5:
stems from this commit: https://github.com/sphinx-doc/sphinx/commit/76110c3ea05e4eef02bc1689c4b9253aca743392 which also implies a change in typing

Novakov:
Thanks for taking care of it :) It's been a pleasure to contribute something useful :)

PR #408, opened 2025-02-15
Add support for serializable html builders (e.g. json, pickle)

Bizordec:
There is an exception occurring in `nav_adapt.py` when building docs with JSON builder:

```python
        context.update(nav=_NavContextObject(global_toc))
        context["nav"].homepage = {
>           "url": context["pathto"](context["master_doc"]),
        }
E       KeyError: 'master_doc'
```

I've fixed it and added support for other HTML-like builders.


jbms:
Thanks.

This patch looks good, but can you rebase to deal with the merge conflicts?

sphinx_immaterial/__init__.py
```diff
@@ -16,12 +16,13 @@
 import sphinx.util.docutils
 import sphinx.writers.html5
 from sphinx import version_info
-
+from sphinxcontrib.serializinghtml import JSONHTMLBuilder
 from . import html_translator_mixin
 from .apidoc import apidoc_formatting
 from .apidoc import fix_sphinx_issue_11147  # noqa: F401
 from . import nav_adapt
 from . import sections  # noqa: F401
```
jbms, line 24: Can you rename it from jsonimpl to something more informative, like json_html_builder?

Bizordec: I just used the same name as in sphinxcontrib-serializinghtml
https://github.com/sphinx-doc/sphinxcontrib-serializinghtml/blob/master/sphinxcontrib/serializinghtml/jsonimpl.py

But if you want, i can rename it

jbms: Yes if you can rename it that would be great, since we have a number of other source files related to json in this theme so jsonimpl is a bit confusing.

Bizordec: Is `json_serialising_implementation.py` okay?
Or `json_builder/json_implementation.py`

jbms: json_serializing_html_builder.py would be good I think

Bizordec: The module itself is used as a serializing implementation in `JSONHTMLBuilder`
https://github.com/sphinx-doc/sphinxcontrib-serializinghtml/blob/f0f3b15b2e6bb47570d605ebd5687643d7a85b9d/sphinxcontrib/serializinghtml/__init__.py#L153C7-L166

It would be weird to write something like `implementation = json_serializing_html_builder`

I think `json_builder_serializing_implementation` would be self-explanatory

jbms: Okay sounds good, thanks

PR #410, opened 2025-02-20
Add support for Sphinx 8.2

jbms:
This also changes the CI to use uv.


2bndy5:
Wow, `uv` added a lot more features since I last checked it out.

2bndy5:
I recently also came across [`pixi`](https://pixi.sh/latest/), yet another project management tool written in rust with goals that overlap `uv`. Both are still in beta though. I'm kinda surprised you're migrating to uv, but I see cibuildwheel package is using uv under the hood now...

.github/workflows/build.yml
```diff
@@ -136,7 +144,7 @@ jobs:
           publish_dir: ./docs/_build/html
 
   test:
-    needs: [check_conventions, build]
+    needs: ["mypy", "check_conventions", "build"]
```
2bndy5, line 147: Can probably just depend on `build` because `build` depends on `mypy` and `check_conventions`.
```suggestion
    needs: ["build"]
```

jbms:
> I recently also came across [`pixi`](https://pixi.sh/latest/), yet another project management tool written in rust with goals that overlap `uv`. Both are still in beta though. I'm kinda surprised you're migrating to `uv`, but I see `cibuildwheel` package is using `uv` under the hood now...

I hadn't heard of pixi --- but from a quick look it seems to have a lot of nice features. I see that it is conda-oriented but actually uses uv under the hood, at least for pypi dependencies.

Since we are publishing to pypi rather than conda I think uv is the right choice here, though.

> 
> BTW, I've been using another tool written in rust to manage what node version I'm using for different projects: [`fnm`](https://github.com/Schniz/fnm).

Yes, I've also been using that, much faster than nvm.

jbms:
CI passes now.

2bndy5 (approved):

PR #417, opened 2025-03-10
Fix pending_xref issue with format_signatures

jbms:
Previously, the format_signatures extension could result in pending_xref nodes with more than one child.  Sphinx assumes that pending_xref nodes only contain a single child, and if the reference fails to resolve, will discard all but the first child.

With this change, if there is more than one child to be added to a pending_xref node, they are wrapped in an `inline` node.

### DIFF
--- a/sphinx_immaterial/apidoc/format_signatures.py
+++ b/sphinx_immaterial/apidoc/format_signatures.py
@@ -154,12 +154,29 @@ def _replace_text_nodes(
         # Ensure the new node does not contain any children. `desc_sig_space`,
         # for example, adds a default child text node with content " ".
         del new_node[:]
-        new_node.extend(_transform_node_list(node.children, _transform_node))
-        if len(new_node.children) == 0:
+
+        new_children = _transform_node_list(node.children, _transform_node)
+
+        if isinstance(new_node, sphinx.addnodes.pending_xref):
+            num_new_children = len(new_children)
+            assert num_new_children != 0
             # A pending_xref with no children is an error and indicates a
             # problem with applying the format.
-            assert not isinstance(new_node, sphinx.addnodes.pending_xref)
+
+            if num_new_children != 1:
+                # Wrap children in an `inline` node because Sphinx expects
+                # pending_xref nodes to only have one child.
+                wrapper = docutils.nodes.inline()
+                wrapper.extend(new_children)
+                new_children = [wrapper]
+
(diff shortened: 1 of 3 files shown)


2bndy5:
The RTD build suffered a sporadic error in extracting black wheel... I reran the RTD build and got a different error:
> ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE

I was able to solve this (in a different project) with
```yml
    pre_install:
      - python -m pip install -U pip
```
But I'm not sure that would help with `uv`.

jbms:
> The RTD build suffered a sporadic error in extracting black wheel... I reran the RTD build and got a different error:
> 
> > ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE
> 
> I was able to solve this (in a different project) with
> 
> ```yaml
> build:
>   jobs:
>     pre_install:
>       - python -m pip install -U pip
> ```
> 
> But I'm not sure that would help with `uv`.

Yes I saw that in the github actions build also.  Not clear if it is a uv bug or something else.

2bndy5:
Its definitely a bug in pip (likely related to cache). Still investigating the source though.

jbms:
Pip or pypi?  Because we aren't using pip in our ci except to download uv.

2bndy5:
I think its something to do with pip's local cache. Haven't gotten around to investigating that yet. Latest pip changes don't seem to be a culprit (at a glance).

jbms:
> I think its something to do with pip's local cache. Haven't gotten around to investigating that yet. Latest pip changes don't seem to be a culprit (at a glance).

uv's local cache is independent of pip's local cache though.

2bndy5:
The error happens on RTD before uv is installed. It could be a pypi problem. I recently got the error again (different RTD project) and had to add `--no-cache-dir` which seemed to fix it. Honestly, this is feeling sporadic.

jbms:
> The error happens on RTD before uv is installed. It could be a pypi problem. I recently got the error again (different RTD project) and had to add `--no-cache-dir` which seemed to fix it. Honestly, this is feeling sporadic.

From the rtd build history it seems to occur with both pip and uv.  Probably a pypi issue.  Anyway it is surely unrelated to this pr.

2bndy5 (approved):

Issue #424, opened 2025-03-12
Drop support for sphinx < 6


jbms:
Common themes like sphinx-rtd, pydata, and furo all require sphinx >= 6.

I don't think there is any advantage in supporting Sphinx 4 and 5 anymore.

2bndy5:
No objections from me.

I think there's quite a few spots in the code base that tries to compensate for older versions of sphinx. This won't be a simple patch.

PR #425, opened 2025-03-12
Drop support for Sphinx<6

jbms:
Fixes https://github.com/jbms/sphinx-immaterial/issues/424.


2bndy5 (approved):
somehow this seemed like it was less work than I thought it would be. 👍🏼

PR #430, opened 2025-03-17
Fix navigation.tabs links and support additional file-wide-metadata

jbms:


src/templates/assets/stylesheets/main/components/_status.scss
```diff
@@ -25,18 +25,25 @@
...
+/* stylelint-disable scss/comment-no-loud */
+
+/* sphinx-immaterial: exclude default status icons,
+   they are defined by custom_admonitions.css instead
+
 :root {
   --md-status: svg-load("material/information-outline.svg");
   --md-status--new: svg-load("material/alert-decagram.svg");
   --md-status--deprecated: svg-load("material/trash-can.svg");
   --md-status--encrypted: svg-load("material/shield-lock.svg");
```
2bndy5, line 38: This `encrypted` status doesn't seem to be included in the defaults. Does it serve a special behavior in addition to the page status?

2bndy5 (commented):
This does a lot more than resolve 1 issue. I adjusted the PR description accordingly.

src/templates/assets/stylesheets/main/components/_status.scss
jbms: I think it might be intended for use with https://github.com/unverbuggt/mkdocs-encryptcontent-plugin but I'm not sure.  It isn't documented. 

In any case it doesn't seem likely to be useful for sphinx, and someone can easily add such a status if they want.

Issue #329, opened 2024-02-14
genindex / modindex seems broken


rnestler:
The index pages don't seem to be generated. Also the page https://jbms.github.io/sphinx-immaterial/py-modindex.html (which is linked from https://jbms.github.io/sphinx-immaterial/demo_api.html#generated-index) just shows `None`.

2bndy5:
I've been aware of this, but I couldn't find the cause.

2bndy5:
fix released in [v0.13.4](https://pypi.org/project/sphinx-immaterial/0.13.4)

PR #434, opened 2025-03-19
Fix special pages like genindex

jbms:
Fixes https://github.com/jbms/sphinx-immaterial/issues/329

### DIFF
--- a/docs/apidoc/cpp/apigen.rst
+++ b/docs/apidoc/cpp/apigen.rst
@@ -1,11 +1,6 @@
 C++ API documentation generation
 ================================
 
-.. warning::
-
-   This functionality is experimental, incomplete and subject to
-   backwards-incompatible changes.
-
 This theme includes an optional extension that generates C++ API documentation
 pages.
 
--- a/docs/cpp_apigen_demo.rst
+++ b/docs/cpp_apigen_demo.rst
@@ -1,12 +1,6 @@
 cpp_apigen demo
 ===============
 
-.. warning::
-
-   This functionality is experimental, incomplete and subject to
-   backwards-incompatible changes.
-
-
 Indexing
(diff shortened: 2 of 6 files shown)


docs/apidoc/cpp/apigen.rst
```diff
@@ -1,11 +1,6 @@
 C++ API documentation generation
 ================================
 
-.. warning::
-
-   This functionality is experimental, incomplete and subject to
-   backwards-incompatible changes.
-
```
2bndy5, line 8: So, we're no longer calling this experimental? IIRC, there's unfinished implementations. I created a [label to more easily track `apigen.cpp` issues at a glance](https://github.com/jbms/sphinx-immaterial/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22feat%3A%20apigen.cpp%22)

2bndy5 (approved):
Clever workaround with the jinja context! Thanks! 🎉

jbms:
Well, it isn't perfect but it works for https://google.github.io/tensorstore which has a large C++ API.

mhostetter:
I know this was fixed, but with the fix I only see my top-level Python module/package in the index. I remember a time when each function was listed (however, I can't find a published build with it.)

Is this expected? Or is it expected to enumerate all the functions in the API?

### v0.13.5

https://mhostetter.github.io/galois/v0.4.x/genindex/

image

2bndy5:
That seems to be a behavior (perceivable bug) specific to `apigen.py` ext. In [our docs' `genindex`](https://jbms.github.io/sphinx-immaterial/genindex.html), all the API entities I see are the members documented with `sphinx.ext.autodoc` (for which the namespace is `test_py_module`).

PR #439, opened 2025-04-08
Don't fail on non-existent `docurls`

barseghyanartur:


2bndy5:
Is this related to an issue or bug? What use case is this supposed to solve?

barseghyanartur:
It's a bug I'm experiencing locally.

2bndy5:
I need more info please. In what scenario is `docurls` undefined? What sphinx builder are you using? Are you not building HTML files?

jbms:
This error is known to occur if you do an incremental build (i.e. not -E) after previously building with a different theme. 

Is that what you may have done?

This PR doesn't hurt but I think there may be other problems that may be tricky to fix so it is better to just use a fresh build.

2bndy5:
I had a suspicion that there was a better way to address this.

barseghyanartur:
Sorry. I think all issues are gone when building into a clean output directory. Something that worked with other themes, but somehow doesn't work with this one.

2bndy5:
We're aware that there are some problems with migrating from other themes. That's why there is a note at the bottom of [our docs' landing page](https://jbms.github.io/sphinx-immaterial/).

PR #447, opened 2025-05-10
wrap section titles in span element

### LINKED ISSUE #442: After removing the <span> around the page title the secondary nav has odd formatting for long strings
The following change has an odd impact: 

Image

https://github.com/jbms/sphinx-immaterial/blob/467513b9f7032780fd24ab54ccd5750bb311b46d/sphinx_immaterial/nav_adapt.py#L239

With sphinx-immaterial 0.12.4 it looked like this:
Image

Now with sphinx-immaterial 0.13.5 it is like that:
Image

In the primary navigation (on left side) the page title (navigation item) is still embedded in a `<span class="md-ellipsis"></span>` and there it looks good as before.
Image

As this change came from the upstrem "Material for MkDocs" I wonder if I might have to open a ticket there as well.

Best regards
Tim

2bndy5:
resolves #442

I'm not sure why the span was removed in #338

### DIFF
--- a/sphinx_immaterial/nav_adapt.py
+++ b/sphinx_immaterial/nav_adapt.py
@@ -229,7 +229,7 @@ class MkdocsNavEntry:
 
     def __init__(self, title_text: str, **kwargs):
         self.__dict__.update(kwargs)
-        self.title = _insert_wbr(title_text)
+        self.title = f"<span>{_insert_wbr(title_text)}</span>"
         if not self.aria_label:
             self.aria_label = title_text
         if "meta" not in kwargs:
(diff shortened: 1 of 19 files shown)


jbms:
I don't remember exactly why I removed it --- maybe because I thought the upstream mkdocs-material HTML template already wrapped it in a `<span class="md-ellipsis">` element in all cases (but apparently only some cases), or maybe because of the truncation issue you mentioned.

In any case this fix seems reasonable.

jbms (approved):

2bndy5:
The test snapshots need to be updated. I tried this while in Windows with `pytest --update-snapshots`, but some tests still fail afterwards.

Furthermore, I'm not sure why the CI skipped the commit I pushed to update the the test snapshots.

Issue #270, opened 2023-07-13
Support for `numpy.typing`


mhostetter:
Hey guys. Hope you're well.

I was adding a type annotation to a function that is `npt.DTypeLike`, see [here](https://numpy.org/devdocs/reference/typing.html#numpy.typing.DTypeLike). However, when displayed with `sphinx-immaterial` the underlying `Union` is unpacked. Is there a way to only display `npt.DTypeLike` in our API docs?

```python
from typing import Optional
import numpy as np
import numpy.typing as npt

def pack(x: np.ndarray, bpe: int, dtype: Optional[npt.DTypeLike] = None) -> np.ndarray:
    return x
```

![image](https://github.com/jbms/sphinx-immaterial/assets/12112573/c1b31da0-632b-41fc-8832-d080821750a3)

mhostetter:
I found this https://stackoverflow.com/a/67483317/11694321. It says to add `from __future__ import annotations` to the top of the module to add this to `conf.py`.

```python
autodoc_type_aliases = {
    "npt.DTypeLike": "~numpy.typing.DTypeLike",
}
```

This now give me the following. It is almost perfect. It is succinct, but the pink `DTypeLike` doesn't hyperlink (even though the yellow one below does, using :obj:`numpy.typing.DTypeLike`).

![image](https://github.com/jbms/sphinx-immaterial/assets/12112573/8e535afe-cf5b-401c-80fa-fe010ee59396)

jbms:
Do you get any warnings?

I would expect this to work.  This theme also has its own type alias mechanism that applies to the Python domain generally (not just to autodoc-generated things):

https://jbms.github.io/sphinx-immaterial/apidoc/python/index.html#confval-python_type_aliases

You could try using that instead.  I believe the `autodoc_type_aliases` only apply in certain cases (and not, for example, when using text signatures as obtained from things defined by C extensions), but I would expect them to apply in this case.

If you don't specify the autodoc_type_aliases, do you get the same thing except it shows `npt.DTypeLike`?

mhostetter:
I got no warnings. I'm including a reproducible example.  [foo-270.zip](https://github.com/jbms/sphinx-immaterial/files/12044248/foo-270.zip)

- With no `from __future__ import annotations` and nothing special in `conf.py` I get the expanded Union.
- Adding `from __future__ import annotations` doesn't help.
[...]

PR #467, opened 2025-10-03
Add Python PEP 613 and PEP 695 type alias support

jbms:


pyproject.toml
```diff
@@ -229,3 +229,6 @@ conflicts = [
 
 [tool.ruff]
 lint.extend-select = ["I"]
+
+[too.ruff.per-file-target-version]
```
2bndy5, line 233:
```suggestion
[tool.ruff.per-file-target-version]
```
Should help understand how to format/lint that 1 file better.

2bndy5:
Is this related to (or somehow helpful toward) #270?

jbms:
> Is this related to (or somehow helpful toward) #270?

It didn't directly address that.  However, it does interact with it in the following way:

- The unresolved reference part of that issue is that the Sphinx Python domain attempts to resolve any types that occur in annotations using the `py:class` role, which previously didn't find `py:data` or `py:attr` objects.
- That issue has now been fixed upstream: https://github.com/sphinx-doc/sphinx/issues/10785  The upstream fix modifies lookup so that `py:class` will also find `py:data` and `py:attr` objects if there is no match of a `py:class` or `py:exception` object.  (Not yet in any released sphinx version)
- However, that fix doesn't address `py:type` objects such as are created with the new support in this PR.
- However, I can add a monkey patch to this PR to backport the upstream sphinx fix and also have it properly support `py:type` in the same way.

2bndy5:
> However, I can add a monkey patch to this PR to backport the upstream sphinx fix and also have it properly support `py:type` in the same way

I feel that would only be needed if the implementation relies on the upstream fix. Otherwise, I'm content with waiting for Sphinx to go through their release process.

PR #477, opened 2025-11-30
Compatibility with Sphinx 9

AA-Turner:


docs/conf.py
```diff
@@ -512,8 +512,8 @@ def get_colors(color_t: str):
     ("cpp:identifier", "my_ns3"),
     ("cpp:identifier", "cpp_apigen_demo"),
     # Example JavaScript types
-    ("js:func", "string"),
-    ("js:func", "SomeError"),
+    ("js:class", "string"),
+    ("js:class", "SomeError"),
```
2bndy5: reverting this. Demo docs declare these entities as JS function sugnatures
```suggestion
    ("js:func", "string"),
    ("js:func", "SomeError"),
```

2bndy5:
This essentially prepares for Sphinx v9 support.

I'll be submitting another PR to update lock files, CI, noxfile, tests, and pyproject.toml. However, `uv` cannot currently rectify using sphinx v9 with myst-parser (for the docs group) because they also have not updated their support for sphinx v9.

2bndy5:
> However, uv cannot currently rectify using sphinx v9 with myst-parser (for the docs group) because they also have not updated their support for sphinx v9.

I'm watching https://github.com/executablebooks/MyST-Parser/pull/1076 for updates in regard to MyST support of sphinx v9.

PR #478, opened 2025-12-01
build(deps): bump mermaid to v11.12.0

ruslo:
I tried to add a (fancy) diagram created by https://mermaid.live/, but it didn't render. The reason is an old version of `mermaid`. Updating to version 11.12.0, which is currently used at https://mermaid.live/, seems to fix the issue.

Tested it on Ubuntu 24.04:

```
$ lsb_release -a
Distributor ID:  Ubuntu
Description:     Ubuntu 24.04.3 LTS
Release:         24.04
Codename:        noble
```

Fetch source code:

```
$ git clone git@github.com:jbms/sphinx-immaterial
$ cd sphinx-immaterial/
```

Install `uv` package manager and Graphviz (for `dot`):

```
$ sudo apt update
$ sudo apt install snapd
$ sudo snap install astral-uv --classic
$ sudo apt install graphviz
```

Run build:

```
$ uvx nox -s build
```

Build documentation:

```
$ uvx nox -s docs --forcecolor
```

HTML index file location:

```
$ ls ./docs/_build/html/index.html
```

*If you want, I can add these instructions to https://github.com/jbms/sphinx-immaterial/blob/main/CONTRIBUTING.md for convenience.*

PR #479, opened 2025-12-03
docs: how to build this project's docs

ruslo:
See https://github.com/jbms/sphinx-immaterial/pull/478#issuecomment-3600397378

---

> Although, I would try to keep the instructions agnostic to any platform

I'm on Ubuntu 24.04 and ready to help with testing on it, but I'm afraid I'm not ready to help with other OSes

> BTW, --forcecolor is only used to add color to stdout for non-interactive terminal sessions.

Removed. I copy/pasted it from GitHub Actions

> instruction about necessary dev tools with links to their respective official install docs

There are many ways to install tools. My goal was to create something quick and practical, a fast path from cloning source code to having an index HTML file for testing, so you don't have to spend an hour on web searches for tooling you're not familiar with. As I mentioned before, I just want to update the dependency package https://github.com/jbms/sphinx-immaterial/pull/478#issue-3679464314, check it worked, and nothing is broken, no more, no less

> Also, please add info about node.js needed, but instead of hard-coding the node.js version in CONTRIBUTING.md, I'd recommend installing fnm or some tool that respects a .node-version file in project root (which would need to be added)

This is black magic for me. I can test it if you provide a baby steps guide :upside_down_face:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,37 @@ Anything contained within the _src_ folder is meant to be merged from the upstre
 
 There is no intention to support any mkdocs extensions in this theme because this theme is designed for the Sphinx documentation generator. You will often find that most of the extensions available for mkdocs are natively implemented in Sphinx or available in the form of a Sphinx extension.
 
+## Building this project's docs
+
+Install the following development tools:
+
+- [`uv` package manager](https://docs.astral.sh/uv/getting-started/installation/)
+- [Graphviz](https://graphviz.org/download/) (for generating graphs at docs' build-time)
+- [node.js](https://nodejs.org/en/download) (using [fnm](https://github.com/Schniz/fnm) is highly recommended)
+
+Setup the Python venv:
+
+```text
+uv sync
+```
+
+Assembling/copying over CSS/JS bundles and icons:
+
+```text
+npm install
+npm run build
+```
+
+Build documentation:
+
+```text
+uv run nox -s "docs(html)"
+```
+
+This sets up a Python venv (managed by `nox`), ensures necessary docs' (Python) dependencies are installed, and runs `sphinx-build`.
+
+If the last command completed successfully, the HTML docs can be browsed locally from the generated `docs/_build/html/index.html` file.
+
 ### Merging in changes from upstream
 
 There is a script titled "merge_from_mkdocs_material.py" that is designed to do most of the heavy lifting when pulling in changes from the mkdocs-material theme. It can only be executed in a Linux shell. More usage information is provided by the help menu:


CONTRIBUTING.md
```diff
@@ -39,6 +39,53 @@ Anything contained within the _src_ folder is meant to be merged from the upstre
...
+
+An example of instructions for Ubuntu 24.04:
+
+```
+$ lsb_release -a
+Distributor ID:  Ubuntu
+Description:     Ubuntu 24.04.3 LTS
+Release:         24.04
+Codename:        noble
+```
```
2bndy5: A contribution guide is not a tutorial nor an example. This is not desired because the steps/instructions should cover all supported platforms.

CONTRIBUTING.md
```diff
@@ -39,6 +39,53 @@ Anything contained within the _src_ folder is meant to be merged from the upstre
...
+
+Fetch source code:
+
+```
+$ git clone git@github.com:jbms/sphinx-immaterial
+$ cd sphinx-immaterial/
+```
+
```
2bndy5: I think this would be implied for contributors. In fact, it may be misleading if working from a fork toward a contribution. This can also be removed.

CONTRIBUTING.md
```diff
@@ -39,6 +39,53 @@ Anything contained within the _src_ folder is meant to be merged from the upstre
...
+$ sudo apt install snapd
+$ sudo snap install astral-uv --classic
+$ sudo apt install graphviz
+```
+
+Run build:
+
+```
+$ uv sync
+$ uvx nox -s build
```
2bndy5: 1. `uv sync` is the only required command here.
2. This command should be described as "Setup Python virtual environment"
3. Please remove the leading `$` from every command in these code blocks.
   - These commands are not specific to using a bash shell
   - It is very annoying to click the "copy this code" button and then paste the leading `$` into my terminal. Even if I was using bash, the leading `$` makes the copied command improper bash script syntax (not executable).
   - Seriously, this is a _very_ annoying practice, and I keep seeing it more and more in technical docs nowadays.

CONTRIBUTING.md
```diff
@@ -39,6 +39,53 @@ Anything contained within the _src_ folder is meant to be merged from the upstre
...
+```
```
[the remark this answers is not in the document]
ruslo: > npm run build

Not working for me, error message:

```
> sphinx-immaterial@0.0.0 clean
> rimraf --glob sphinx_immaterial/{*.html,partials,.icons,LICENSE,bundles}

sh: 1: rimraf: not found
```

`uvx nox -s build` works fine

ruslo:
> I think the [section titled "Code style"](https://github.com/jbms/sphinx-immaterial/blob/main/CONTRIBUTING.md#code-style) should be updated. We have migrated to using uv since that section was authored. And this new section's list of needed dev tools is a common pre-requisite for both building docs and static analysis

I can help with the testing, as we did with the npm/uv

> the "code style" section can be updated in a separate PR

Agree, I think it's better to open new PR

